PR #31, opened 2015-08-04
Fix issue with array prototype sanitization

mockdeep:
When sanitizing arrays on older browsers with polyfills we see output
like this:

```
"plugins" => {
  "forEach" => {"bind" => {"bind" => "[CIRCULAR DATA STRUCTURE]"}},
    "map" => {"bind" => "[CIRCULAR DATA STRUCTURE]"},
    "filter" => {"bind" => "[CIRCULAR DATA STRUCTURE]"},
    "every" => {"bind" => "[CIRCULAR DATA STRUCTURE]"},
    "some" => {"bind" => "[CIRCULAR DATA STRUCTURE]"},
    "reduce" => {"bind" => "[CIRCULAR DATA STRUCTURE]"},
    "reduceRight" => {"bind" => "[CIRCULAR DATA STRUCTURE]"},
    "indexOf" => {"bind" => "[CIRCULAR DATA STRUCTURE]"},
    "lastIndexOf" => {"bind" => "[CIRCULAR DATA STRUCTURE]"}
}
```

`Notice#_santize` is serializing arrays like objects and seeing the
functions defined on the prototype. (Side note, `#_sanitize` might also
want to just pass on functions.)

### DIFF
--- a/spec/notice.spec.coffee
+++ b/spec/notice.spec.coffee
@@ -59,6 +59,14 @@ describe 'Notice', ->
           json = JSON.stringify(notice.payload()) # shouldn't cause error.
           expect(json.match(/CIRCULAR DATA STRUCTURE/g).length).toEqual(1)
 
+        it 'converts arrays in context', ->
+          Array.prototype.blah = -> 'whatevs'
+          e = mockError()
+          c = { "foo": ['boo'], 'bar': ['baz'] }
+          notice = new Notice({ stack: e.stack, message: e.message, name: e.name, context: c })
+          json = JSON.stringify(notice.payload()) # shouldn't cause error.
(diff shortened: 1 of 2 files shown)


mockdeep:
@joshuap Feel free to take what you want from this, just wanted to have a stab at fixing it.

joshuap:
@mockdeep thanks for the example; I'm going to play with it and see what approach works best.

joshuap:
@mockdeep I took a slightly different approach with `_sanitize` but I used your test in 2d3e608081d8770b97555dbb7c93863e026a9ebd. What do you think? I'm also going to work on ignoring functions in objects (might just replace them with "[FUNC]").

mockdeep:
Great! lgtm. This will help a lot.

PR #45, opened 2016-06-01
Use typeof instead of instanceof

benjaminwood:
See here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_context_(e.g._frames_or_windows)

> Different scope have different execution environments. This means that they have different built-ins (different global object, different constructors, etc.). This may result in unexpected results. For instance, [] instanceof window.frames[0].Array will return false, because Array.prototype !== window.frames[0].Array and arrays inherit from the former. This may not make sense at first but when you start dealing with multiple frames or windows in your script and pass objects from one context to another via functions, this will be a valid and strong issue. For instance, you can securely check if a given object is in fact an Array using Array.isArray(myObj)

I ran into problems when using honeybadger.js with slimerjs which is a phantomjs alternative. Using typeof solves it for me, but I'm not sure if there are other implications of these changes. However, the tests _do_ pass!

Let me know what you think.

### DIFF
--- a/honeybadger.js
+++ b/honeybadger.js
@@ -147,7 +147,7 @@
       context: {},
       beforeNotifyHandlers: []
     }
-    if (opts instanceof Object) {
+    if (typeof opts == 'object') {
       for (k in opts) { self[k] = opts[k]; }
     }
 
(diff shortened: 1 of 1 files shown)


benjaminwood:
@joshuap did you end up resolving this some other way?

benjaminwood:
🎉

PR #47, opened 2016-08-26
Add type declarations for TypeScript

ianks:
This will be very handy for your typescript users (myself included). It allows a user to have access to all of the types in Honeybadger with no configuration at all.

Let me know what you think!


joshuap:
Love it, thanks. And sorry for the delay!

PR #50, opened 2016-10-27
Add 'async' configuration

wbowling:
We had a situation where we needed to send a Honeybadger notification right before a hard redirect, but as soon as the redirect occurred any XHR requests in progress were cancelled and the notification would never be sent. 

This option allows us to force the request to be synchronous, ensuring that it is send before the redirect happens.


joshuap:
Thanks, and sorry for the delay in merging this. Looks good.

PR #52, opened 2016-12-27
Strict mode

miry:
We have an  issue when use ruby assets and compile the library with other libraries with `'use strict';`. It would great to have proper JS here also to not have exception like `i is not declared`


joshuap:
Looks good. Thanks!

PR #55, opened 2017-01-25
Documents the debug configuration option

jbottigliero:
This doesn't seem to be mentioned on the [documentation site](http://docs.honeybadger.io/lib/javascript.html) or in the `README.md`... not sure if that's intentional.


joshuap:
Thanks!

PR #58, opened 2017-02-24
Update HB typescript definitions for config opts

damien:
Updates "honeybadger-js".Config interface to use camelCase key names


joshuap:
Thanks!

PR #60, opened 2017-04-04
Wrap function: don't try to extend non-extensible object

artygus:
Some third party ad libraries pass [frozen](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) functions to setTimeout which results in errors in console and broken functionality.

If function passed to `wrap` is sealed or frozen, fn.__hb would still be undefined after [assignment](https://github.com/honeybadger-io/honeybadger-js/blob/master/honeybadger.js#L357).

### DIFF
--- a/honeybadger.js
+++ b/honeybadger.js
@@ -350,7 +350,7 @@
     // removeEventListener.
     function wrap(fn, force) {
       try {
-        if (typeof fn !== 'function') {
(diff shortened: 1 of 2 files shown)


joshuap:
:+1: thanks!

PR #63, opened 2017-04-19
Fix README - correct data attribute for API key

kasima:
This tripped me up for a bit. Changes the README to match http://docs.honeybadger.io/lib/javascript.html


joshuap:
@kasima Thanks for looking into this, and sorry for the confusion! We recently switched to support/prefer "apiKey" instead of "api_key" (in keeping with JS naming conventions). You uncovered a bug in reading the "apiKey" value from a data attribute. I've fixed this on the master branch and will get a new CDN/npm/bower release out tomorrow.

kasima:
Great, thanks!

PR #66, opened 2017-05-08
Allow error messages to be ignored

mhuerster:
Allow the user to provide `ignorePatterns` as an array of regexes. Any
error whose message matches one of these patterns will not trigger a
notification.

Re: issue #6

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,10 @@ Honeybadger.configure({
   // Action (optional)
   action: '',
 
+  // Ignore Patterns (optional)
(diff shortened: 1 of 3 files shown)


joshuap:
Looks good, thanks @mhuerster! I'll merge now; I may adjust the location of where the ignore check happens in `notify()` before releasing (the `currentErrIs`/`currentPayload` code is kind of a hack to make multiple notify operations atomic and I think I prefer to keep those clauses separate for clarity).

PR #69, opened 2017-06-01
Add variable declarations

HolixSF:
Why:

* receiving undeclared variable errors upon
  instantiating as a module in a react app using webpack 2

Changes:

* added var declarations in various places in source code
  where needed.


joshuap:
Thanks!

PR #74, opened 2017-08-05
Don't send request if disabled is true

novito:
# Problem

When disabled is set to true (we don't want to send the request to notify), we are not outputting the error in development mode even if debug option was set to true.  The reason being is that we are halting too early.

# Solution

Halt later in the process. Specifically, any time we try to do a request, if disabled is true, don't do it.

### DIFF
--- a/honeybadger.js
+++ b/honeybadger.js
@@ -223,6 +223,8 @@
     }
 
     function request(url) {
+      if (config('disabled', false)) { return false; }
+
       // Use XHR when available.
       try {
         // Inspired by https://gist.github.com/Xeoncross/7663273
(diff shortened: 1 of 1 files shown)


novito:
@joshuap Do you mind taking a look at this? I'm thinking if we need to add a test to make sure that we halt later (by making sure that we do any of the actions we were not doing before), or it's cool as it is.

joshuap:
@novito thanks, this is on my todo list. Will review soon!

joshuap:
@novito this looks good to me. I'll merge it and play with it (the build should be passing on master as far as I can tell). Thanks!

PR #76, opened 2017-08-19
Add missing configuration options to TS type definition

shaungrady:
Missing options:

- debug
- revision
- ignorePatterns
- async


joshuap:
Thanks!

PR #87, opened 2017-12-18
Fix for Google Closure advanced compilation

rads:
We've been trying to use the `honeybadger-js` library from ClojureScript. One problem we ran into was that the Google Closure compiler was renaming the properties of the payload object before it was sent to the API.

This PR changes the payload to use quoted object properties so that Google Closure will not try to optimize it. This enables the library to work with ClojureScript without breaking any existing behavior.

### DIFF
--- a/honeybadger.js
+++ b/honeybadger.js
@@ -261,7 +261,7 @@
         return false;
       }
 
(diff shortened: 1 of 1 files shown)


monde:
thanks for the PR @rads

PR #89, opened 2018-01-05
Ensure main fn is bound to `this`

### LINKED ISSUE #88: Honeybadger + UglifyJS (with webpack) issues
Not entirely sure what's happening, but when I combine Honeybadger with UglifyJS when building with webpack, I get this error:
```
Cannot read property 'console' of undefined
```
It's in reference to `this.console` in `function log()`. I believe it's related to how the Honeybadger script replaces window.setTimeout and uses `func.apply(null, ...` -- `null` being the key issue there.

kellym:
This fixes #88. It seems that UglifyJS is running in an undefined context. This ensures that the function is bound to the current context, i.e. `window`.

### DIFF
--- a/honeybadger.js
+++ b/honeybadger.js
@@ -631,4 +631,4 @@
   });
 
   return factory;
-}));
+}.bind(this)));


kellym:
Also, I attempted to add a spec to ensure console.log was being called when debugging, but the test passed with and without this fix. I'm not entirely sure how to replicate the binding issue in testing without it being forced.

kellym:
Tests pass locally for me. I can't entirely decide if this is an issue to fix on Honeybadger's side (i.e. with this change) or if it's a problem with UglifyJS.

kellym:
Ah, so, I should've noticed this earlier. The core issue is the `use strict` statement in combination with the fact that UglifyJS is nesting the functions inside one another. 
```javascript
// For example, this is undefined here
! function a() {
  'use strict';
  console.log(this, ' is undefined');
}();

// This is essentially what the Honeybadger script is doing:
! function b() {
  'use strict';
  console.log(c(), ' is window');
}();

function c() {
  return this;
};

// and UglifyJS is converting it to this:
! function b() {
  'use strict';
  console.log((function c() {
    return this;
  })(), ' is (no longer) window');
}();
```
This seems more like a fault of UglifyJS than Honeybadger -- you're welcome you either disregard this PR or consider it as a solution to prevent context issues now and in the future.

kellym:
Can't seem to reproduce easily with UglifyJS. /shrug

kellym:
Closing based on discussion in #88.

PR #95, opened 2018-06-13
 fix(ChromeExtension): Fixed issue when this is a undefined

shahen94:
### Current Behavior

We tried to use honeybadger to send analytics for our chrome extension. But after importing this JS file we get an error `Cannot read console of undefined`.

### Expected Result

No error should be thrown.

### What I did ?

Since `this` is undefined in chrome extension's  `content script`  But `window` exists. We should check it and put it in case when this is undefined.

### DIFF
--- a/honeybadger.js
+++ b/honeybadger.js
@@ -48,7 +48,7 @@
     // Browser globals (root is window).
     root.Honeybadger = factory();
   }
-}(this, function (root) {
+}(this || window, function (root) {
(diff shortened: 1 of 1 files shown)


shahen94:
Tests are failing - But I'm not sure how to fix it, because there's no any message that indicates what could be the reason for failing

joshuap:
This should be fixed in #96. Let me know if not. Thanks!

PR #96, opened 2018-06-14
Fix continuing issues with this being undefined/undefined context

### LINKED ISSUE #94: Debug logging broken when loading via Webpack
I just noticed that when `debug: true` is enabled in the config in a Webpack project, it doens't log. It's because `root` is undefined in the `log()` function. I think this PR is probably to blame:

https://github.com/honeybadger-io/honeybadger-js/pull/92

We should make sure this is the only problem that was introduced by that PR when we fix it.

joshuap:
See the commit messages for each of the commits in this PR, which fix specific aspects of this issue. This reverts the behavior added in #88/#92 and takes a different approach.

This PR does not fix the context for `this` being undefined in certain environments inside the factory builder function (see #89); instead, it relies directly on `window` for objects which depend on `window`, rather than assuming that `this` is window inside the function. We're already using this approach elsewhere in the factory (for instance, in `Honeybadger.notify`), so I think it's fairly safe. If we need to revisit this in the future, we can do it as a whole.

This should also fix #94 and #95.

cc @geoffreak @monde @stympy @rathboma @kellym @shahen94 -- lmk what you think.

### DIFF
--- a/honeybadger.js
+++ b/honeybadger.js
@@ -28,7 +28,7 @@
   // Build the singleton factory. The factory can be accessed through
   // singleton.factory() to instantiate a new instance.
   var factory = function(){
-    var f = builder(root);
+    var f = builder();
     var singleton = f(scriptConfig);
     singleton.factory = f;
     return singleton;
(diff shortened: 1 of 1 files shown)


joshuap:
@shahen94 thanks! I am going to go ahead and merge/release this since I haven't heard from others and logging is currently broken in a number of places. I confirmed that this fixes webpack logging. If anyone runs into future issues in other enviropnments such as electron, open a new issue and we'll look into it.

Issue #86, opened 2017-12-14
Add afterNotify, like beforeNotify


stympy:
As [requested by Ryan](https://twitter.com/rkneufeld/status/941395519145852928)

Issue #51, opened 2016-12-15
Add option to send API request with POST


butchler:
Currently, if the context is too large, the request fails with a `414 Request-URI Too Large` status. If the context could be stored in the data of a  `POST` request, maybe the request would not fail.

I know that this might not be possible in browsers that don't support `XMLHttpRequest`, but if your app requires `XMLHttpRequest` anyway, it would be nice to be able to opt-in to using `POST` to avoid `notify`s possibly failing with `414`s.

The obvious solution would be to just not send so much context, but I would ideally like to send the entire Redux state of the app at the time of the error, which could get very large depending on how much content the app is currently showing.

joshuap:
@butchler thanks for the suggestion! This is already on my todo list, actually. I'm planning on looking into adding support for larger payloads specifically for the purpose of including Redux state early next year.

butchler:
Thank you :)

joshuap:
I've reverted the change I made related to this issue for now; it's still on the roadmap, but I need to think about the approach a while longer so that we can maintain cross-browser support.

butchler:
That's fine. Should this issue be re-opened, though?

joshuap:
Yeah, reopening.

butchler:
Hey there, just checking in. Any progress on this?

Right now we're using our own custom Honeybadger client that sends data to the API via POST, but I'd really like to just delete that code and go back to using the official client. However, I feel a little uncomfortable using this knowing that errors might silently fail to send if there's too much data in the context. The ability to detect if an error failed to send (https://github.com/honeybadger-io/honeybadger-js/issues/40) would also be enough to go back to using `honeybadger-js`.

sljavi:
+1
this would be very useful for our team

joshuap:
@butchler yep, this is something I'd like to tackle soon. I think it may require a pretty big overhaul of the current lib and probably deprecation of some of the older browser we support (it's probably time to drop IE6, right? 😀 )

I'm sorry I don't have an ETA yet, but I'll bump this up my priority list. I'm looking for some contract help if you guys know anyone with senior JS+OSS experience.

joshuap:
This is currently on master and will be released in 1.0.0

Issue #40, opened 2016-02-05
How to know Honeybadger.notify is successful


madhurakhal:
While sending notice using Honeybadger.notify, I want to know whether sending notice is success or not. Is there anyway like callback function that I can provide which will be called after successful notification. I feel irritated using because sometimes it works but sometimes not.

joshuap:
@madhurakhal can you give me an example of when the notification does not work? Do you mean it doesn't work due to throttling by the server or something like that?

stympy:
Here's [another request](https://twitter.com/xiongtx/status/941396430639341568) for a post-notify callback.

joshuap:
We now have `afterNotify` handlers to cover this use case: https://docs.honeybadger.io/lib/javascript/reference/configuration.html#afternotify-handlers

PR #100, opened 2019-02-11
Use POST to send error reports

joshuap:
## Status
WIP

## Description
This PR does three things:

- Changes honeybadger.js to use POST for all HTTP requests
- Updates the Jasmine test suite to use the latest version of Jasmine (3.x)
- Drops support for browsers which do not support the CORS/POST method w/ `XMLHttpRequest`&mdash;namely, IE < 10

*Note: Another factor in dropping IE 8/9 support is that it's difficult to test in CI. Jasmine no longer supports < IE 10, which means we can't upgrade if we want to have CI builds for older browsers. The alternatives would be: a) switch to a different test framework that maintains support (if one exists), or b) use a very outdated test framework and deal with other problems resulting from that.*

### Benefits

- With GET, we're currently limited to the maximum data allowed by the query string (exceeding this results in a "Request-URI Too Long" error). That has become a problem, especially in frameworks that provide a lot of context data such as Vue and React. POST allows us to send *more data*.
- POST allows us to JSON encode our payloads similar to our back-end client libraries.
- It should be simpler to address issues like #40 and #86 

### Drawbacks

- In the past 30 days, 0.03% of our client-side error reports were from IE8/9 (lower than the overall usage estimate by [Can I use](https://caniuse.com/usage-table)). Dropping support for those browsers would mean that anyone wanting to maintain support should lock to the legacy version of honeybadger.js (or maybe use a polyfill).

## Related Issues
https://github.com/honeybadger-io/honeybadger-js/issues/51
https://github.com/honeybadger-io/honeybadger-js/issues/40
https://github.com/honeybadger-io/honeybadger-js/issues/86

## Todos
- [ ] Documentation


nshoes:
> That has become a problem, especially in frameworks that provide a lot of context data such as Vue and React. POST allows us to send _more data_.

Bless you 🙌 All of our projects will benefit from this.

> 0.03% of our client-side error reports were from IE8/9

Interesting! Not a lot :P

Kudos for not using a request lib :neckbeard: LGTM 🚢

PR #138, opened 2019-05-16
Add cache-control header

stympy:
## Status
**READY**

## Description
Add cache-control header to objects on s3 so cloudfront will pass it along to browsers.


joshuap:
Thanks!

Issue #133, opened 2019-05-14
Handle and report errors thrown inside of Promises


ryanoglesby-tc:
It seems that errors that are triggered from inside of Promises are not caught by global `window.onerror` handlers, which means Honeybadger does not get notified of these.

I'd like to request that Honeybadger also listen for `unhandledrejection` so that these errors are also reported to Honeybadger. Does that fit into the scope of this library? I'd be willing to put together a PR too.

## What are the steps to reproduce this issue?
1. Throw an error in a Promise
2. Do not catch the error

Example:

```js
/*
  The promise below will throw `ReferenceError: nosuchfunc is not defined`

  The `onerror` listener will NOT be invoked.
  The `unhandledrejection` listener WILL be invoked.
*/
window.addEventListener('error', err => {
  console.log('window onerror', err)
})

window.addEventListener('unhandledrejection', rj => {
  console.log('unhandledrejection', rj)
})

new Promise((resolve, reject) => {
  nosuchfunc()
}).then(data => {
  // never gets called
  console.log('then:', data)
})
```

## What happens?

Errors thrown inside of promises are not caught by `window.onerror` handlers, so Honeybadger does not get notified of these.

## What were you expecting to happen?

Honeybadger is notified of any errors that bubble up from Promises.

## Any logs, error output, etc?

N/A

## Any other comments?

N/A

## What versions are you using?
**Operating System:** …
**Package Version:** …
**Browser Version:** …

joshuap:
Thanks for documenting this, @ryanoglesby-tc! I'm open to adding this, assuming we can add a config flag to disable it, and it can call any existing event bindings after calling ours (like our `window.onerror` handler does). I think some people will want to opt-out so that they can handle promise rejections however their framework recommends.

If we make this the new default, we may need to bump the major version, since this could potentially start duplicating reports for people who have already attached to this event to report errors (or have done so in their framework).

ryanoglesby-tc:
Ok, I'll see what I can do!

joshuap:
Released in v2.0.0

PR #140, opened 2019-05-18
Report unhandled promise rejections #133

ryanoglesby08:
## Status

**READY**

## Description

This PR adds support for reporting unhandled promise rejections, as described in Issue #133.

In short, this PR adds a listener to `window.onunhandledrejection`, which gets invoked when a promise rejection is not handled with either `try...catch` or `Promise.catch()`. This might happen if an un-expected runtime error occurs, such as as an null pointer exception, or not handling a normal Promise rejection. The normal `window.onerror` handler will NOT be invoked in these cases. Here are some examples.

```js
fetch('/api/user/123')
  .then(data => {
    // User 123 does not have any data, but the API call returned a 200. So data = {}
    let name = data.profile.name // 💥TypeError: Cannot read property 'name' of undefined
  })
```

```js
let stuff = undefined
new Promise((resolve, reject), () => {
  if( stuff ) {
    resolve(stuff)
  }
  else {
    reject(new Error("We have no stuff!"))
  }
}).then(stuff => {
    // do stuff
})

// => 💥Uncaught (in promise) Error: We have no stuff!
```

See: https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event

## More info

* This is enabled by default, but you could switch the config so that it is opt in if that's preferred. Or do a major version bump.
* Will want to add the config option of `onunhandledrejection` to the main docs site too.

## Todos
- [x] Tests
- [ ] Documentation

### DIFF
--- a/spec/main.spec.js
+++ b/spec/main.spec.js
@@ -818,9 +818,9 @@ describe('Honeybadger', function() {
 
         afterNotify(done, function() {
           expect(requests.length).toEqual(1);
-          expect(request.payload.error.class).toEqual('window.onunhandledrejection')
+          expect(request.payload.error.class).toEqual('window.onunhandledrejection');
           expect(request.payload.error.message).toEqual('UnhandledPromiseRejectionWarning: Something has gone wrong');
-          expect(request.payload.error.backtrace).toEqual('UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch().')
+          expect(request.payload.error.backtrace).toBeUndefined();
         });
       });
 
@@ -835,9 +835,9 @@ describe('Honeybadger', function() {
 
         afterNotify(done, function() {
           expect(requests.length).toEqual(1);
-          expect(request.payload.error.class).toEqual('Error')
+          expect(request.payload.error.class).toEqual('Error');
           expect(request.payload.error.message).toEqual('UnhandledPromiseRejectionWarning: Error: Something has gone wrong');
-          expect(request.payload.error.backtrace).toEqual('the stack trace\nUnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch().')
+          expect(request.payload.error.backtrace).toEqual('the stack trace');
         });
       });
 
@@ -868,9 +868,9 @@ describe('Honeybadger', function() {
 
         afterNotify(done, function() {
           expect(requests.length).toEqual(1);
-          expect(request.payload.error.class).toEqual('window.onunhandledrejection')
+          expect(request.payload.error.class).toEqual('window.onunhandledrejection');
           expect(request.payload.error.message).toEqual('UnhandledPromiseRejectionWarning: {"errorCode":"123"}');
-          expect(request.payload.error.backtrace).toEqual('UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch().')
+          expect(request.payload.error.backtrace).toBeUndefined();
         });
       });
 
--- a/spec/main.spec.js
+++ b/spec/main.spec.js
@@ -800,6 +800,104 @@ describe('Honeybadger', function() {
     });
   });
 
+  describe('window.onunhandledrejection callback', function() {
+    describe('default behavior', function() {
+      beforeEach(function() {
+        Honeybadger.configure({
+          api_key: 'asdf'
+        });
+      });
+
+      it('reports the promise rejection reason when it is a string', function(done) {
+        let reason = 'Something has gone wrong'
+        let promiseRejection = new PromiseRejectionEvent('unhandledrejection', {
+          promise: Promise.reject(reason),
+          reason
+        })
+        window.onunhandledrejection(promiseRejection);
+
+        afterNotify(done, function() {
+          expect(requests.length).toEqual(1);
+          expect(request.payload.error.class).toEqual('window.onunhandledrejection')
+          expect(request.payload.error.message).toEqual('UnhandledPromiseRejectionWarning: Something has gone wrong');
+          expect(request.payload.error.backtrace).toEqual('UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch().')
+        });
+      });
+
+      it('reports the promise rejection reason when it is an Error object', function(done) {
+        let reason = new Error('Something has gone wrong')
+        reason.stack = "the stack trace"
+        let promiseRejection = new PromiseRejectionEvent('unhandledrejection', {
+          promise: Promise.reject(reason),
+          reason
+        })
+        window.onunhandledrejection(promiseRejection);
+
+        afterNotify(done, function() {
+          expect(requests.length).toEqual(1);
+          expect(request.payload.error.class).toEqual('Error')
+          expect(request.payload.error.message).toEqual('UnhandledPromiseRejectionWarning: Error: Something has gone wrong');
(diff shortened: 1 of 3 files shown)


ryanoglesby-tc:
@joshuap Any feedback or thoughts on this PR?

joshuap:
Hey @ryanoglesby-tc, sorry for the delay on this. I'm going to look at it today. Thanks!

spec/main.spec.js
```diff
@@ -800,6 +800,104 @@ describe('Honeybadger', function() {
...
+          promise: Promise.reject(reason),
+          reason
+        })
+        window.onunhandledrejection(promiseRejection);
+
+        afterNotify(done, function() {
+          expect(requests.length).toEqual(1);
+          expect(request.payload.error.class).toEqual('Error')
+          expect(request.payload.error.message).toEqual('UnhandledPromiseRejectionWarning: Error: Something has gone wrong');
+          expect(request.payload.error.backtrace).toEqual('the stack trace\nUnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch().')
```
joshuap: Question about this stack trace--I don't think I've seen traces with any kind of additional message/warning at the end; usually it's at the beginning:

```
Error message here
  at url:1:1
  at ...
```

Am I missing something there?

src/builder.js
```diff
@@ -553,6 +558,55 @@ export default function builder() {
...
+          stack: PROMISE_REJECTION_WARNING
+        });
+      }
+
+      return function(promiseRejectionEvent) {
+        onunhandledrejection(promiseRejectionEvent);
+        if (typeof original === 'function') {
+          return original.apply(this, arguments);
+        }
+        return false;
```
joshuap: Do you know if the return value here has any significance? I can't remember.

ryanoglesby08: I was following the pattern from the `onerror` callback. Looking back at the docs on MDN, it seems that the return value of `onerror` is significant, but the return value of `` is not. I think I will remove that.

https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onunhandledrejection

spec/main.spec.js
ryanoglesby08: I got this message and pattern directly from node. Node also appends this message at the bottom of the stack trace. I'm fine removing it, but it seems like valuable context for errors of this sort. It doesn't seem like browsers have a message like this built in already.

What do you think?

Here is an example:

```
$ node
> new Promise(resolve => { foo() })
Promise {
  <rejected> ReferenceError: foo is not defined
      at resolve (repl:1:26)
      at new Promise (<anonymous>)
      at repl:1:1
      at Script.runInThisContext (vm.js:124:20)
      at REPLServer.defaultEval (repl.js:325:29)
      at bound (domain.js:425:14)
      at REPLServer.runBound [as eval] (domain.js:438:12)
      at REPLServer.onLine (repl.js:650:10)
      at REPLServer.emit (events.js:198:15)
      at REPLServer.EventEmitter.emit (domain.js:481:20) }
> (node:11338) UnhandledPromiseRejectionWarning: ReferenceError: foo is not defined
    at resolve (repl:1:26)
    at new Promise (<anonymous>)
    at repl:1:1
    at Script.runInThisContext (vm.js:124:20)
    at REPLServer.defaultEval (repl.js:325:29)
    at bound (domain.js:425:14)
    at REPLServer.runBound [as eval] (domain.js:438:12)
    at REPLServer.onLine (repl.js:650:10)
    at REPLServer.emit (events.js:198:15)
    at REPLServer.EventEmitter.emit (domain.js:481:20)
(node:11338) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:11338) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

ryanoglesby08:
Do you want me to write up a changelog entry as part of this PR? Or will that happen as part of the next release?

spec/main.spec.js
joshuap: Gotcha, thanks for the explanation. I'm thinking that we should leave it out of the browser version. I like the extra context, but our backend doesn't support extra logging at the end of the stack trace, unfortunately.

joshuap:
@ryanoglesby08 a changelog entry on this PR would be great. You can add it under the "Unreleased" heading. You'll need to add a subheading too--I'm thinking that this should go under a new "Added" heading. Feel free to include your GitHub handle if you want attribution ([example](https://github.com/honeybadger-io/honeybadger-js/blob/master/CHANGELOG.md#added)).

spec/main.spec.js
ryanoglesby08: Gotcha, will remove.

src/builder.js
ryanoglesby08: Removed the return value.

ryanoglesby08:
@joshuap Ok, ready for another look. I have removed the Node.js promise rejection warning from the stack trace, removed the superfluous return value from the handler, and added the CHANGELOG entry.

ryanoglesby08:
@joshuap Hi Josh. Any updates on this?

spec/main.spec.js
```diff
@@ -941,6 +941,110 @@ describe('Honeybadger', function() {
     });
   });
 
+  describe('window.onunhandledrejection callback', function() {
+    describe('default behavior', function() {
+      beforeEach(function() {
+        if (typeof PromiseRejectionEvent === 'undefined') {
```
joshuap: `PromiseRejectionEvent` is unsupported in Edge and FireFox, apparently:

https://developer.mozilla.org/en-US/docs/Web/API/PromiseRejectionEvent

I added this filter to just skip the specs for these browsers. I didn't add anything to skip the event handler from being attached, but I don't *think* that's a problem, since it just won't be called in those browsers anyway?

src/builder.js
```diff
@@ -569,6 +574,49 @@ export default function builder() {
...
+            name: reason.name,
+            message: `UnhandledPromiseRejectionWarning: ${reason}`,
+            stack
+          });
+
+          return;
+        }
+
+        let message = typeof reason === 'string' ? reason : JSON.stringify(reason)
+        notify({
```
joshuap: Also had to tweak the `notify` call here, to not rely on the `generated` argument (which is reserved for some special cases).

spec/main.spec.js
ryanoglesby08: Yeah that seems to be the case. Tested it in Firefox locally and it Firefox just ignored the `window.onunhandledrejection ` function

ryanoglesby08:
@joshuap Seems okay to me!

spec/main.spec.js
ryanoglesby08: Also good catch! I didn't even realize this was not yet supported in all browsers. A little surprised Firefox is lagging behind on this one!

ryanoglesby08:
Any updates on this PR @joshuap?

joshuap:
@ryanoglesby08 I’ll get this merged and release a beta this week. Thanks for the reminder!

joshuap:
@ryanoglesby08 this has been released as v2.0.0-beta.0: 

https://www.npmjs.com/package/honeybadger-js/v/2.0.0-beta.0

and:

https://js.honeybadger.io/v2.0-beta/honeybadger.min.js

Can you try it out and let me know how it works for you? Thanks!

ryanoglesby08:
Hi @joshuap, we've been using the v2 beta for about a week now and its looking good! We are seeing plenty of unresolved promise errors too. Seems like its working well. How long do you want the beta to run before releasing the real v2?

joshuap:
I think we're probably good to release it. I deployed it to our main Rails app and it seems OK there too. I'll put the release on my todo list for later this week or next.

joshuap:
@ryanoglesby08 v2.0.0 has been released. Thanks for your work on this!!

PR #154, opened 2019-06-11
Fix before notify props

joshuap:
## Status
**READY**

## Description
This PR fixes a bug in the `beforeNotify` logic where many notice properties were not passed in to the handler as documented:

https://docs.honeybadger.io/lib/javascript/guides/filtering-sensitive-data.html

Properties (such as `notice.url` could be assigned, but prior values were not available on the object to inspect, causing (i.e.) this example to fail:

```js
Honeybadger.beforeNotify(function(notice) {
  if (/creditCard/.test(notice.url)) { // <-- notice.url is undefined
    notice.url = '[FILTERED]'; // <-- works, but doesn't get here
  }
});
```

This PR also hides the internal stack trace `generator` properties, which were previously exposed through the `notice` (or would have been ;)).

## Todos
- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce
```bash
git pull --prune
git checkout fix-beforeNotify-props
npm test
```

### DIFF
--- a/src/builder.js
+++ b/src/builder.js
@@ -287,9 +287,9 @@ export default function builder() {
         revision: err.revision || config('revision')
       });
 
-      if (isIgnored(err, config('ignorePatterns'))) { return false; }
-
       if (checkHandlers(self.beforeNotifyHandlers, err)) { return false; }
+
+      if (isIgnored(err, config('ignorePatterns'))) { return false; }
 
       var data = cgiData();
       if (typeof err.cookies === 'string') {
--- a/src/builder.js
+++ b/src/builder.js
@@ -272,8 +272,10 @@ export default function builder() {
         return false;
       }
 
+      let generator;
       if (generated) {
-        err = merge(err, generated);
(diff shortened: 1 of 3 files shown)


src/builder.js
```diff
@@ -124,6 +124,23 @@ export default function builder() {
     return false;
   }
 
+  function objectIsEmpty(obj) {
+    return ((function() {
+      var k, results;
+      results = [];
+      for (k in obj) {
+        if (!Object.prototype.hasOwnProperty.call(obj, k)) continue;
```
rabidpraxis: I know you just refactored this as part of the PR, but this is doing unneeded work. You could just check for hasOwnProperty and return true once it is found, instead of collecting all the keys for a count at the end.

joshuap: @rabidpraxis thanks! I think this does the trick: https://github.com/honeybadger-io/honeybadger-js/pull/154/commits/46d8fab779a13ca052b9d316c7e3903ee3d1664e

PR #210, opened 2019-09-18
Breadcrumbs

joshuap:
## Status
**READY**

## Description
This adds breadcrumbs for JavaScript with a `Honeybadger.addBreadcrumb` API that is similar to Ruby and Elixir.

Instrumentation of common browser events has been completed. I kept the implementation as simple as possible for now (for example, the derived names of HTML elements in click events are rudimentary). There should be areas to improve as we begin to use the instrumentation in practice.

Docs: https://github.com/honeybadger-io/docs/pull/67

## Related PRs
https://github.com/honeybadger-io/honeybadger-ruby/pull/312
https://github.com/honeybadger-io/honeybadger-elixir/pull/235
https://github.com/honeybadger-io/docs/pull/67
https://github.com/honeybadger-io/honeybadger/issues/2040
https://github.com/honeybadger-io/honeybadger/issues/2039
https://github.com/honeybadger-io/honeybadger/pull/2038

## Instrumentation
- [x] Console
- [x] Click events
- [x] Network: XMLHttpRequest
- [x] Network: fetch
- [x] Navigation/history
- [x] Error
- [x] Notice

## Todos
- ~~Refactor browser instrumentation~~ (we'll do this in a future PR/release)
- [x] Instrumentation tests
- [x] Documentation
- [x] Enhance XHR/fetch breadcrumb messages

## Steps to Test
```bash
> git pull --prune
> git checkout breadcrumbs
> npm test
```


src/builder.js
```diff
@@ -313,8 +322,14 @@ export default function builder() {
         data['HTTP_COOKIE'] = encodeCookie(err.cookies);
       }
 
+      const breadcrumbs = {
+        'enabled': breadcrumbsEnabled(),
+        'trail': self.breadcrumbs.slice(0),
```
rabidpraxis: What does 0th slice do?

joshuap: It's the same as calling `array.slice()`. It duplicates the array object so that additional breadcrumbs aren't added to it. I think I'll remove the `0` argument and make it implicit.

rabidpraxis: 😢

src/builder.js
```diff
@@ -516,6 +517,22 @@ export default function builder() {
...
+      try {
+        message = stringNameOfElement(event.target)
+      } catch(e) {
+        message = '[unknown]'
+      }
+
+      if (message.length === 0) return
+
+      self.addBreadcrumb(message, {
+        category: 'ui.click'
```
joshuap: @rabidpraxis is there a specific naming convention you'd like to use for subcategories, or is this OK? We might add additional UI-related events in the future, like key presses, input, focus, etc.

rabidpraxis: Yeah, this makes sense to me. So far I have kept the categories pretty simple, but I like the idea of subcategories. The only current usage of the category is to choose an icon in the interface.

joshuap:
Thanks for the thorough @rabidpraxis!

> In both the Ruby & Elixir implementations we add breadcrumbs for both the Error and the Notify events. I thought it might be weird to add these events in the stack, especially if multiple errors happen in succession (since the breadcrumbs are not purged, like they are in the other libraries). But then I thought it might actually be nice to know if a user was seeing multiple errors by looking into the breadcrumbs.

I added todos ☝️ for this.

> Also, I threw this in a personal project and found I had to click into many of the messages to know what they were related to. This might just be my use case and the project I was in, but I'm wondering if adding some details to the messages might help scanning breadcrumbs when there are many.
> 
> I don't know the right formula, but I played around a little.
> 
> Before:
> 
> [screenshot]
> 
> After:
> 
> [screenshot]
> 
> "Click: " seems redundant, maybe moving the category up into the breadcrumb or perhaps after there is a click icon it will be more apparent what each target dom name means.

Yeah, I was expecting for us to make some changes to the presentation here. I think that the solution probably involves tweaking the UI in the app as well as the messages themselves. For instance, in addition to the icon, I think it would be useful to surface the category in the main list. That alone would add a lot of visual context to the message, and may be enough. From there we could revisit the messages to see what additional info would be useful to have in the list. What do you think about that?

rabidpraxis:
I am with you on the category, we should move that up and I think that will make the UI events 💯. 

It's cool if you want to punt but I would vote to enhance the request breadcrumb message in this iteration. For FE projects that make a ton of async requests per "request" (any RESTful site) I think it would be helpful to have some context in the message. Unfortunately the best context I can think of is the full URI. Putting that in the message will make some of the fetch breadcrumbs really long (especially if there are query params). I wonder if there is something more terse we could add that gives it a little more distinction, like what we do for the SQL breadcrumbs (we concat just the model name).

joshuap:
> For FE projects that make a ton of async requests per "request" (any RESTful site) I think it would be helpful to have some context in the message. Unfortunately the best context I can think of is the full URI. Putting that in the message will make some of the fetch breadcrumbs really long (especially if there are query params). I wonder if there is something more terse we could add that gives it a little more distinction, like what we do for the SQL breadcrumbs (we concat just the model name).

For XHR and fetch, we could do the following:

1. For URLs on the same host, strip the auth/host/port info (i.e. http://localhost:3002/api/users becomes /api/users)
2. For all URLS, strip the query string (i.e. /api/recipes?filter=whatever becomes /api/recipes)
3. Prefix with type of request (xhr or fetch), i.e.: "fetch: /api/recipes"

PR #249, opened 2019-11-21
Deal with nulls in array

rabidpraxis:
Not sure if this is the right solution to nulls being included in sanitized arrays, but nulls are erroring because they are of an "object" type, yet don't implement `hasOwnProperty`


stympy:
👍

Issue #242, opened 2019-11-14
Exponential slowdown when running Jest tests with honeybadger included


aub:
## What are the steps to reproduce this issue?
1. I have a React project that uses Honeybadger for error notification. It has a test suite with a bunch of very simple tests using enzyme's `mount` function to render the components and make sure they work ok.
2. When I run this test suite after adding Honeybadger to my code, the suite slows to a crawl and then locks up completely so that I have to kill it manually.

## What happens?
The tests seem to go into an infinite loop. With some logger debugging, I found that what happens is that Honeybadger is being reinitialized with each test. During its initialization, it installs all of its instrumentation in the form of event handlers. Since it hasn't removed the handlers from the previous test(s) they start piling up, and this makes each test execution slower than the previous one.

Note that there's a little bit of conjecture here... I'm not sure why/if enzyme is holding onto the handlers from the previous test or how one might reset it. I've added console logging and seen it print out tons of logs when the handlers fire though.

## What were you expecting to happen?
Only a single instrumentation of Honeybadger so the tests don't slow down.

## Any logs, error output, etc?
…

## Any other comments?
…

## What versions are you using?
**Operating System:** MacOS 10.15.1
**Package Version:** 1.0.4
**Browser Version:**

joshuap:
Hey @aub, thanks for the report. How are you including honeybadger-js in your code? Our library guards against the instrumentation being installed more than once by the global instance, but doesn't expect to be included multiple times on a page, for instance. I'm not sure what would happen in that case.

aub:
I've got `honeybadger-js` and `@honeybadger-io/react` as dependencies in my package.json file, and then in the component modules where I need access to honeybadger I import it with:

`import Honeybadger from 'honeybadger-js'`

Maybe that's the wrong approach? I can do a quick test to verify, but I _think_ this problem is specific to the tests. We're using Jest with Enzyme and jsdom, and it seems like jsdom is supposed to reset its virtual browser after each test, but that doesn't jive with the pileup of handlers I'm seeing.

joshuap:
I think you should be able to import multiple times and it will return the same singleton object. It would be good to verify that in your project. It does sound like something strange is going on with the test setup, though.

aub:
Yep, verified that it only initializes once in the regular app. I'm not quite sure where to go from here with the tests, but I'll see if I can figure out how/when jsdom reinitializes the virtual browser.

joshuap:
[...]

PR #255, opened 2019-11-26
Unwrap functions before instrumenting

joshuap:
This should fix https://github.com/honeybadger-io/honeybadger-js/issues/242


joshuap:
@rabidpraxis thanks for looking! I think that it probably makes sense to re-replace because if (for some reason) the script was included multiple times, it makes more sense for the most recent version to own the replaced function. Not sure if I explained that very well.

rabidpraxis:
@joshuap Oh okay yeah I understand what you mean. In that case, then yeah that seems perhaps the safest move.

PR #286, opened 2020-01-08
Fix TypeScript type definitions

### LINKED ISSUE #279: Type definition incorrect.
The type definition file dictates that most of the functions provided by the main export return a void value:
https://github.com/honeybadger-io/honeybadger-js/blob/fae5d6178eb08b443dc3a41b1f72f04f050ca58d/honeybadger-js.d.ts#L35
However, it seems this is incorrect upon reviewing the actual source. For instance `configure`:

https://github.com/honeybadger-io/honeybadger-js/blob/fae5d6178eb08b443dc3a41b1f72f04f050ca58d/src/builder.js#L460-L465

joshuap:
Fixes #279

Related: https://github.com/honeybadger-io/honeybadger-react/pull/91

## Status
**READY**

## Description
Fixes a few incorrect return types.

### DIFF
--- a/honeybadger-js.d.ts
+++ b/honeybadger-js.d.ts
@@ -32,15 +32,15 @@ declare module "honeybadger-js" {
 
     class Honeybadger {
         static apiKey: string;
-        static configure(config: Config): void;
-        static context: any;
+        static configure(config: Config): Honeybadger;
+        static context: object;
         static environment: string;
-        static notify(...args: any[]): void;
+        static notify(...args: any[]): object;
         static onerror: boolean;
         static wrap<T extends Function>(func: T): T;
-        static setContext<T extends Object>(context: T): void;
-        static resetContext(): void;
-        static beforeNotify(func: (notice?: Notice) => void): void;
+        static setContext<T extends Object>(context: T): Honeybadger;
+        static resetContext(): Honeybadger;
+        static beforeNotify(func: (notice?: Notice) => void): Honeybadger;
         static beforeNotifyHandlers: ((notice?: Notice) => void)[];
         static factory(config: Config): Honeybadger;
     }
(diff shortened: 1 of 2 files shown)


honeybadger-js.d.ts
```diff
@@ -32,15 +32,15 @@ declare module "honeybadger-js" {
 
     class Honeybadger {
         static apiKey: string;
-        static configure(config: Config): void;
-        static context: any;
+        static configure(config: Config): Honeybadger;
+        static context: object;
```
schontz: `context: any` was right. That allows it to be an object with any properties.

honeybadger-js.d.ts
```diff
@@ -32,15 +32,15 @@ declare module "honeybadger-js" {
...
         static environment: string;
-        static notify(...args: any[]): void;
+        static notify(...args: any[]): object;
```
schontz: Use `any` instead of `object`.

joshuap:
This should take care of that: https://github.com/honeybadger-io/honeybadger-react/pull/91

Assuming that looks right, I'll get both released either later this afternoon or tomorrow morning.

schontz:
Thanks. I’ll install them straightaway so confirm all is well.

joshuap:
@schontz honeybadger-js v2.0.2 and honeybadger-react v0.0.6 have been released with the new type definitions. Let me know how it works for you!

schontz:
@joshuap It works!

I only list `@honeybadger/react` as a dependency. Running `npm i @honeybadger/react@latest` seemed to not pick up the new `honeybadger-js` info, even though it was listed at 2.0.2 in `package-lock.json`. I re-installed the react module and all was well. I think it was just an issue with my TS Server being out-of-date in VS code though. I just mention it in case anyone else has an issue.

Thanks for providing the TypeScript types. It makes integration a lot easier in TS projects.

Issue #266, opened 2019-12-12
Notify throws exceptions with null/undefined error and a custom class name


ryanoglesby-tc:
## What are the steps to reproduce this issue?
1. Call `Honeybadger.notify(undefined, 'DescriptiveClass');` or `Honeybadger.notify(null, 'DescriptiveClass');`

## What happens?

It throws an uncaught TypeError

```
Uncaught TypeError: Cannot read property 'match' of undefined
    at isIgnored (builder.js:48)
    at notify (builder.js:307)
    at Object.self.notify (builder.js:426)
```

## What were you expecting to happen?

I would expect the same thing to happen as if `Honeybadger.notify(undefined)` were called, which is that the error is not reported...

## Any logs, error output, etc?

N/A

## Any other comments?

I can work on a PR to fix, wanted to get an issue report up in the meantime!

## What versions are you using?
**Operating System:**
Mac OS 10.15.1

**Package Version:**
honeybadger-js@2.0.0

**Browser Version:**
Chrome 78.0.3904.108

joshuap:
@ryanoglesby-tc a PR would be great, thanks!

ryanoglesby-tc:
Hi @joshuap, will you be releasing a patch release, v2.0.1 with this fix? We'd like to pull in the bug fix without upgrading to 2.0-beta.

joshuap:
@ryanoglesby-tc [v2.0.1](https://github.com/honeybadger-io/honeybadger-js/releases/tag/v2.0.1) has been released from the [v2.0-stable](https://github.com/honeybadger-io/honeybadger-js/tree/v2.0-stable) branch.

PR #311, opened 2020-02-13
Fix n.match is not a function error w/ ignorePatterns

joshuap:
See https://github.com/honeybadger-io/honeybadger-js/issues/266

This adds an additional check to make sure that the err.message object
can be tested against a regexp.


joshuap:
@stympy thanks!

PR #314, opened 2020-02-23
feat(after-notify): hook to execute handlers after the request was sent

anymaniax:
## Status

**WIP**

## Description
Add a hook to execute handlers after the request was sent.
#86  

I used the onload callback from XMLHttpRequest but I'm not sure If that's what you want?

I also pass the response to the handler but maybe we should also pass the payload sent?

## Todos
- [x] Implementation
- [x] Tests
- [ ] Documentation

### DIFF
--- a/spec/main.spec.js
+++ b/spec/main.spec.js
@@ -795,6 +795,53 @@ describe('Honeybadger', function() {
     });
   });
 
+  describe('afterNotify', function() {
+    beforeEach(function() {
+      Honeybadger.configure({
+        apiKey: 'asdf',
+        environment: 'config environment',
+        component: 'config component',
+        action: 'config action',
+        revision: 'config revision',
+        projectRoot: 'config projectRoot',
+        afterNotifyHandlers: []
+      });
+    });
+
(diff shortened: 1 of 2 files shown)


anymaniax:
@joshuap what do you think about what I did?

anymaniax:
@joshuap sounds great I will do that!

anymaniax:
@joshuap I did the update, can you check again?

For the testing part, I checked and I didn't find a way to test it with what you have right now. I about thought using miragejs to mock the api call. What do you think?

joshuap:
Thanks! I'm on vacation until Tuesday next week; I'll review when I get back. We currently use sinon to mock API calls. We also have an [integration spec](https://github.com/honeybadger-io/honeybadger-js/blob/master/spec/integration.spec.js#L184) which does a couple of real requests (although I try to limit integration specs to testing browser features or other features which are impossible to unit test).

anymaniax:
Enjoy your vacation :). Sorry, I had misunderstood how useFakeXMLHttpRequest works.

PR #319, opened 2020-03-03
Add onunhandledrejection to Config interface type definition

tojofo:
Typescript compiler fails with an error if trying to use `onunhandledrejection`

## Status
**READY**

## Description
Allow `onunhandledrejection` to be used in Typescript projects when calling `Honeybadger.configure`


joshuap:
Thanks!

PR #359, opened 2020-05-15
Add definition for Honeybadger.addBreadcrumb

Fustrate:
## Status

**READY**

## Description
Fixes the following error when running type checks on a project that uses Honeybadger:

    TS2339: Property 'addBreadcrumb' does not exist on type 'typeof Honeybadger'.


joshuap:
Thanks!

PR #406, opened 2020-07-28
Instrument on first Honeybadger.configure call

### LINKED ISSUE #397: Console logging references honeybadger source
All instrumented logging operations are pointing to the Honeybadger source.

Screen Shot 2020-07-14 at 10 02 56 AM

This can be confusing as some users rely on that source to know where their logging calls are referenced from.

joshuap:
This makes two important changes:

1. Breadcrumbs auto-instrumentation is now performed on first `Honeybadger.configure` call, instead of on import.

   For now, `window.onerror` and other pre-breadcrumbs instrumentation is still performed on import due to architectural limitations. This is something we may be able to change in the next version of the library which is currently in-progress.

2. The `breadcrumbsEnabled` option has been enhanced to support an object syntax so that individual breadcrumbs instrumentation types (such as "console") may be disabled. `breadcrumbsEnabled: true` and `breadcrumbsEnabled: false` will also continue to work.

Fixes #397

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ CHANGELOG](http://keepachangelog.com/) for how to update this file. This project
 adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- Perform breadcrumbs instrumentation on first `Honeybadger.configure` call (#406)
+- Add config option to disable auto breadcrumbs by type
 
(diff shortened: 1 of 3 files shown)


DryFlyRyan:
Thanks for the quick work on this @joshuap

joshuap:
Thanks! I'm going to merge and release a beta.

PR #418, opened 2020-08-19
Add referrer and userAgent to notice object

joshuap:
This allows the referrer and userAgent strings to be filtered in `beforeNotify` handlers:

https://docs.honeybadger.io/lib/javascript/guides/filtering-sensitive-data.html

### DIFF
--- a/src/builder.js
+++ b/src/builder.js
@@ -262,7 +262,7 @@ export default function builder() {
     function buildPayload(err) {
       const data = {};
       if (err.userAgent) { data['HTTP_USER_AGENT'] = err.userAgent; }
-      if (err.referrer) { data['HTTP_REFERER'] = document.referrer; }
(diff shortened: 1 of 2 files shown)


src/builder.js
```diff
@@ -269,7 +260,9 @@ export default function builder() {
     }
 
     function buildPayload(err) {
-      let data = cgiData();
+      const data = {};
+      if (err.userAgent) { data['HTTP_USER_AGENT'] = err.userAgent; }
+      if (err.referrer) { data['HTTP_REFERER'] = document.referrer; }
```
rabidpraxis, line 265: Should this assignment be from `err.referrer` not `document.referrer`?

joshuap: Ah, good catch. Yes it should. :)

joshuap: @rabidpraxis fixed, thanks!

PR #477, opened 2020-11-09
Disable honeybadger.js in development environments

### LINKED ISSUE #71: Notifications still sent even when environment is development
# Problem

Even when Honeybadger is configured with environment being 'development', the notifications are being sent. This doesn't happen with the Ruby library.

# Expected behavior

By default don't notify if environment is 'development'

joshuap:
Fixes #71 

- [x] Rename the `disabled` config option to `reportData` to match Ruby config (keep legacy support, maybe?)


rabidpraxis (approved):
👍

PR #510, opened 2021-01-14
Filter URL params

### LINKED ISSUE #509: Filter URL params
Similar to #503, we can deconstruct URL params in `payload.request.url`, apply filters, and then reconstruct the URL like we do in the Ruby gem.

joshuap:
Fixes #509


stympy (approved):
This looks good, though you might want to add some tests for the unhappy path... urls like these:

* https://www.example.com/
* https://www.example.com/?secret=&foo=bar
* https://www.example.com/?secret&foo=bar

Issue #539, opened 2021-02-26
No TypeScript types for requestHandler or errorHandler


getabetterpic:
## What are the steps to reproduce this issue?
1. Set up a new Node project using TypeScript (I'm using Nest)
2. Add `@honeybadger-io/js` to the project
3. Attempt to add `app.use(Honeybadger.requestHandler)`

## What happens?
TypeScript complains since there isn't a `requestHandler` or `errorHandler` type on the Honeybager export.

## What were you expecting to happen?
I should be able to use `Honeybadger.requestHandler` and `Honeybadger.errorHandler` without any TypeScript error.

## Any logs, error output, etc?
N/A

## Any other comments?
Here's a new Nest app illustrating this failure https://github.com/getabetterpic/honeybadger-not-working.

```
git clone git@github.com:getabetterpic/honeybadger-not-working.git
npm install
npm run start
```

Error:

```
yarn start
yarn run v1.23.0-20210103.1434
$ nest start
src/main.ts:12:23 - error TS2339: Property 'requestHandler' does not exist on type 'Honeybadger'.

12   app.use(Honeybadger.requestHandler);
                         ~~~~~~~~~~~~~~
src/main.ts:14:23 - error TS2339: Property 'errorHandler' does not exist on type 'Honeybadger'.

14   app.use(Honeybadger.errorHandler);
                         ~~~~~~~~~~~~

Found 2 error(s).

error Command failed with exit code 1.

```

## What versions are you using?
**Operating System:** macOS and Linux
**Package Version:** 3.0.3
**Browser Version:** N/A

joshuap:
Fix released in 3.0.4

PR #540, opened 2021-02-26
Add server middleware signatures to type declarations

getabetterpic:
## Status
**READY**

## Description
Fixes https://github.com/honeybadger-io/honeybadger-js/issues/539.

We're using a TypeScript-based Node framework ([Nest](nestjs.com)) and we need to use Honeybadger as the error reporting lib. However we can't set up the Honeybadger middleware `requestHandler` and `errorHandler` functions due to no types for them.

We don't actually need the `lambdaHandler` type but figured I'd go ahead and add that in here as well while I was at it.

## Todos
- [x] ~~Tests~~ N/A, only touches the type declarations file
- [x] Documentation

## Steps to Test or Reproduce
I tested this out by linking locally to our project and ensuring the app compiled.

### DIFF
--- a/honeybadger.d.ts
+++ b/honeybadger.d.ts
@@ -11,6 +11,11 @@ declare class Honeybadger {
   public resetContext(context?: Record<string, unknown>): Honeybadger
   public addBreadcrumb(message: string, opts?: Partial<Honeybadger.BreadcrumbRecord>): Honeybadger
   public wrap(func: (...args: unknown[]) => unknown): (...args: unknown[]) => unknown
+
+  // Server middleware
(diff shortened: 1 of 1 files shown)


getabetterpic:
Let me know if there's anything I need to do to fix that integration test failure. I'm guessing there are some secrets that aren't exposed to my fork that are causing that to fail but wasn't entirely sure.

joshuap (approved):
This looks good, thanks! 💯

joshuap:
@shipjs prepare

joshuap:
Fix released in 3.0.4

PR #543, opened 2021-03-02
Add clearContext method to clear context and breadcrumbs

joshuap:
This should allow people to clear the context and breadcrumbs in between requests (in Express, for instance).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased][latest]
 ### Added
-- Add `Honeybadger.clearContext()` method to clear context and breadcrumbs
+- Add `Honeybadger.clear()` method to clear context and breadcrumbs
 - Deprecate `Honeybadger.resetContext()`
 
 ## [3.0.4] - 2021-03-01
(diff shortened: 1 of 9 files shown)


joshuap:
I renamed the `resetContext()` method (and removed the optional argument) to be more like the Ruby client. It also felt like it was mixing responsibilities too much with the addition of breadcrumbs.

rabidpraxis (approved):
This all looks good to me. 

I am curious why you chose to go with `clearContext` vs a more generic `clear`? The former looks like it would not be also clearing breadcrumbs.

joshuap:
> I am curious why you chose to go with clearContext vs a more generic clear? The former looks like it would not be also clearing breadcrumbs.

@rabidpraxis I could be talked into this. :) The reason is that in Ruby it's `Honeybadger.context.clear!`, which suffers from the same issue. I don't know... should we move towards a more standardized name for "all the session-based stuff?"

rabidpraxis:
@joshuap My vote would for a more general name, yeah. Either way, it would be good to be consistent between languages. Also, the `clear!` method in the Ruby client is on the Agent, so the call ends up `Honeybadger.clear!`.

joshuap:
Cool, thanks for the suggestion. I'll rename it to `Honeybadger.clear()` in JS.

joshuap:
@shipjs prepare

Issue #530, opened 2021-02-08
[Ship.js] Don't create a release PR if there are no unreleased changes


joshuap:
Links:

- [Our Ship.js config file](https://github.com/honeybadger-io/honeybadger-js/blob/master/ship.config.js)
- [Our Changelog file](https://github.com/honeybadger-io/honeybadger-js/blob/master/CHANGELOG.md)
- [Ship.js on GitHub](https://github.com/algolia/shipjs)
- [Ship.js docs](https://community.algolia.com/shipjs/guide/)

We have the [ship.js trigger](https://github.com/honeybadger-io/honeybadger-js/blob/master/.github/workflows/shipjs-schedule-prepare.yml) run the `npm run release` command once per week. Our [custom Ship.js config](https://github.com/honeybadger-io/honeybadger-js/blob/master/ship.config.js) is used to determine the next release version based on the unreleased section of our [changelog](https://github.com/honeybadger-io/honeybadger-js/blob/master/CHANGELOG.md). (Edit: see additional info about our Ship.js setup in the [readme](https://github.com/honeybadger-io/honeybadger-js#release-automation).)

Currently `npm run release` creates a PR even if there are no unreleased changes in the changelog. I want to skip those PRs—if there are no unreleased changes, then skip creating the PR. I'm not sure if this is [a configurable option](https://community.algolia.com/shipjs/reference/all-config.html#monorepo) w/ Ship.js; you might want to [ask the maintainers](https://github.com/algolia/shipjs/issues).

incredimike:
Update: Looking into using the `shouldPrepare` callback to check if a release should be prepared. 
https://community.algolia.com/shipjs/reference/all-config.html#shouldprepare

Having some trouble testing it with --dry-run, since the dry run flag causes shipjs to bail on the preparation step BEFORE the shouldPrepare function has been called. I've modified the library for my own testing, and I'm reaching out to the ShipJS community to see if this should be adjusted in the library at large.

joshuap:
It turns out the `shouldPrepare` approach didn't work because `commitNumbersPerType` is always 0 (since we're not using keep a changelog instead of conventional commits).

https://github.com/honeybadger-io/honeybadger-js/commit/37c6822b311ef3a6d12c54d97e08fc9b91e41115

joshuap:
@shipjs prepare

(testing that revert.)

shalvah:
[...]

PR #550, opened 2021-03-09
Do not prepare PR if no commits

incredimike:
## Status
**READY?**

## Related issues
https://github.com/honeybadger-io/honeybadger-js/issues/530

## Description
If commits are ready to release, the `commitNumbersPerType` object will have properties like `{ feat: 2, fix: 4, chore: 8 }`. If there are no changes, the object is empty. 

This took a little longer than expected to sort out since ShipJS was not running the `shouldPrepare` callback when `--dry-run` flag was used. I checked with the maintainers and have submitted a PR to change this behavior.

I haven't managed to generate a "positive" test (i.e. have commits to push), but all the tests run fine!

### DIFF
--- a/ship.config.js
+++ b/ship.config.js
@@ -33,6 +33,9 @@ module.exports = {
       return `${yyyy}-${mm}-${dd}`
     }
   },
+  shouldPrepare: ({ commitNumbersPerType }) => {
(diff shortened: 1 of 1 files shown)


joshuap (approved):
Thanks!

PR #557, opened 2021-03-25
feat: add support for tags

ParamagicDev:
## Status

READY

## Description

Adds support for tags. I wasnt sure how to add to the honeybadger docs to reflect these changes.

You can now add tags 3 ways:

```js
Honeybadger.configure({
  tags: "1, 2, 3"
})

const client = Honeybadger.configure()
client.setContext({tags: ["1", "2", "3"]})

client.notify("testing", {tags: "1, 2, 3"})
```

## Todos

- [x] Tests
- [x] Documentation (updated the typings)

## Steps to Test or Reproduce

```bash
> git pull --prune
> git checkout <branch>
> npm run test
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add tagging to errors with 3 possible API's
 
-```bash
-Honeybadger.configure({tags: string | string[] | undefined})
-Honeybadger.configure().setContext({tags: string | string[] | undefined})
-Honeybadger.configure().notify('error', {tags: string | string[] | undefined})
-```
+    ```bash
+    Honeybadger.configure({tags: string | string[] | undefined})
+    Honeybadger.configure().setContext({tags: string | string[] | undefined})
+    Honeybadger.configure().notify('error', {tags: string | string[] | undefined})
+    ```
 
 ## [3.1.0] - 2021-03-04
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased][latest]
 
+### Added
+
+- Add tagging to errors with 3 possible API's
+
+```bash
+Honeybadger.configure({tags: string | string[] | undefined})
+Honeybadger.configure().setContext({tags: string | string[] | undefined})
+Honeybadger.configure().notify('error', {tags: string | string[] | undefined})
+```
+
 ## [3.1.0] - 2021-03-04
 ### Fixed
 - Add default reason for unhandled promises (previously reason was "undefined") (#546)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased][latest]
-
 ### Added
-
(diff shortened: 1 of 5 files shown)


src/core/client.ts
```diff
@@ -131,14 +144,20 @@ export default class Client {
 
     if (objectIsEmpty(notice)) { return false }
 
+    const noticeTags = this.__constructTags(notice.tags)
+    const contextTags = this.__constructTags(this.__context["tags"])
+    const configTags = this.__constructTags(this.config.tags)
+    const tags = noticeTags.concat(contextTags).concat(configTags)
```
joshuap, line 150: I forget, does concat remove duplicates?

src/core/client.ts
```diff
@@ -216,15 +235,19 @@ export default class Client {
...
       },
       request: {
         url: filterUrl(notice.url, this.config.filters),
         component: notice.component,
         action: notice.action,
         context: notice.context,
         cgi_data: cgiData,
-        params: filter(notice.params, this.config.filters) || {},
+        params: {
+          tags: notice.tags,
```
joshuap, line 248: We don't need to send the tags in params—just in `payload.error.tags` like you have above.

joshuap (commented):
Looking good! A few comments.

joshuap:
One more: don't forget the changelog entry.

src/core/client.ts
ParamagicDev: Survey says no!

```js
["1"].concat(["1"])
// => ["1", "1"]
```

ParamagicDev: @joshuap would you like it to remove duplicates? I could easily run a `.filter()`

ParamagicDev: @joshuap I totally forgot about the `Set` data structure and used that rather than a filter to extract unique tags.

ParamagicDev: So i tried using `const uniqueTags = Array.from(new Set(tags))` and it broke CI. Interesting...

src/core/client.ts
```diff
@@ -216,15 +239,16 @@ export default class Client {
...
+        tags: notice.tags
       },
       request: {
         url: filterUrl(notice.url, this.config.filters),
         component: notice.component,
         action: notice.action,
         context: notice.context,
         cgi_data: cgiData,
-        params: filter(notice.params, this.config.filters) || {},
+        params: filter(notice.params, this.config.filters),
```
joshuap, line 251: Does this still need the fallback?

```suggestion
        params: filter(notice.params, this.config.filters) || {},
```

src/core/client.ts
joshuap: It might be that `Set` is not available in IE 11?

joshuap: > @joshuap would you like it to remove duplicates? I could easily run a `.filter()`

Yeah, I would like to remove duplicates. 👍

ParamagicDev: Bingo. Set is supported in IE11, but cannot be converted back to an array with an iterable.

src/core/client.ts
ParamagicDev: Good catch. Forgot to put it back since I was making an object.

joshuap (approved):
Looks good!

CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Added
+
+- Add tagging to errors with 3 possible API's
+
+```bash
+Honeybadger.configure({tags: string | string[] | undefined})
+Honeybadger.configure().setContext({tags: string | string[] | undefined})
+Honeybadger.configure().notify('error', {tags: string | string[] | undefined})
+```
```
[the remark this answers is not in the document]
ParamagicDev: Got it!

PR #606, opened 2021-07-01
Dependency updates

shalvah:
- Updated jest and ts-jest to new releases
- Configuresd dependabot to only reduce noise
- Made some npm scripts cross-platform


shalvah:
@joshuap  Re dependencies: are we using depfu or dependabot? I'd like to exclude the examples/ and test/ directories from dependency upgrades like this https://github.com/honeybadger-io/honeybadger-js/pull/584, and transitive dependencies like this https://github.com/honeybadger-io/honeybadger-js/pull/587. I think this new Dependabot configuration should work, but I don't know about depfu.

Issue #604, opened 2021-06-29
Typescript module error


sadiqawos:
## What are the steps to reproduce this issue?
1. Set up a new Node project using TypeScript (I'm using electron)
2. Add @honeybadger-io/js to the project

## What happens?
Typescript compiler complains about the following
```
node_modules/@honeybadger-io/js/honeybadger.d.ts:3:49 - error TS2307: Cannot find module 'express' or its corresponding type declarations.

3 import { NextFunction, Request, Response } from 'express'
```

## What were you expecting to happen?
No complaints by typescript compiler about `express` since I'm not using it in my project

## Any logs, error output, etc?
```
> tsc && electron dist/app.js --development

node_modules/@honeybadger-io/js/honeybadger.d.ts:3:49 - error TS2307: Cannot find module 'express' or its corresponding type declarations.

3 import { NextFunction, Request, Response } from 'express'
                                                  ~~~~~~~~~
Found 1 error.

```

## What versions are you using?
**Operating System:* Ubuntu 
**Package Version:** 3.2.1
**Browser Version:** N/A

joshuap:
Thanks for the bug report!

The issue may be that express is in our `devDependencies`, which aren't installed. I don't want to add express to our main dependencies just to get the type definitions, so we'll have to come up with some different type definitions (or a workaround).

https://github.com/honeybadger-io/honeybadger-js/blob/master/honeybadger.d.ts#L3

sadiqawos:
Thanks for the prompt response.

If the concern is to keep dependencies to a minimum, why don't you split up the typedef into its own package?

I'm not too familiar with this myself, so I had to read [this](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies) very quickly!

joshuap:
Hi @sadiqawos, this should be fixed in v3.2.2 (just released). Can you try that out and let me know if you have any more problems? Thanks!

PR #611, opened 2021-07-14
Add missing methods to typedefs

### LINKED ISSUE #609: Typescript : Missing type for clear()
1. Use `clear()` in a project with TS

## What happens?
TypeScript return `Property 'clear' does not exist on type 'Honeybadger'`

`clear()` should be in the Honeybadger type

**Package Version:** 3.2.1

shalvah:
Fixes #609 

Addendum: Any reason why we manually write the typedefs? Methinks we could use tsc to generate them. thereby keeping them in sync with any code changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased][latest]
+### Fixed
+- TypeScript: Add missing methods typedefs (#611)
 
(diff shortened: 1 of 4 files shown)


shalvah:
Doing that ☝ may also be a good way to resolve #604—we could import/re-export the Express types in our codebase, without having our users being exposed to the express module types directly.

joshuap (approved):
Looks great, thanks!

> Any reason why we manually write the typedefs? Methinks we could use tsc to generate them. thereby keeping them in sync with any code changes.

I'd love to do this, I just wasn't quite sure how to set it up with npm + the isomorphic package setup (if you look in package.json we have different entry points for node/browser imports). I opted to just maintain an additional universal public type def instead. You're welcome to take a look at it if you want to.

PR #612, opened 2021-07-16
Improve TypeScript usage

shalvah:
(Apologies for the size of the PR; the changes involved turned out to be a little more than anticipated.)

This PR:
1. Adds `@types/express` to our dependencies, fixing #604 
2. Adds some missing null checks. A number of places had a check for `if (typeof thing === 'object')`, with the subtle bug that `null` in JS also returns `object`.
3. Improves typing. A number of types (some of which were user-facing) were incorrect or incomplete. I've fixed them so they're (mostly) correct, providing better IDE suggestions.
4. Switches to (mostly) automated generation of type declarations. This is done by enabling declarations in our `tsconfig.json`, then updating our Rollup config to point to the right paths for each build. However, this generates separate type decl files for the two builds, so there's a manual `honeybadger.d.ts` file that re-exports them as one file as before. Additionally, users can still import the `browser/honeybadger.d.ts` or `server/honeybadger.d.ts` files like before (with the added advantage of accurate types for the build).
5. Updated the README to explain this for first-time contributors.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -7,6 +7,7 @@ import pkg from './package.json'
 // These plugins are used for all builds
 const sharedPlugins = [
   replace({
+    preventAssignment: false,
     exclude: 'node_modules/**',
     __VERSION__: pkg.version
   }),
@@ -16,13 +17,25 @@ const sharedPlugins = [
 // These plugins are used for UMD builds
 const umdPlugins = [
   ...sharedPlugins,
-  typescript()
+  typescript({
+    tsconfig: './tsconfig.json',
+    exclude: [
+      "./src/server.ts",
+      "./src/server/**"
+    ]
+  })
 ]
 
 // These plugins are used for Node builds
 const nodePlugins = [
   ...sharedPlugins,
-  typescript()
+  typescript({
+    tsconfig: './tsconfig.json',
+    exclude: [
+      "./src/browser.ts",
+      "./src/browser/**"
+    ]
+  })
 ]
 
 export default [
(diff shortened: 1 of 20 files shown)


rollup.config.js
```diff
@@ -7,6 +7,7 @@ import pkg from './package.json'
 // These plugins are used for all builds
 const sharedPlugins = [
   replace({
+    preventAssignment: false,
```
shalvah, line 10: Added this because rollup gives a deprecation warning about the default value of this flag changing in the next major, so I pinned this to the current default pre-emptively.

test-d/browser.test-d.tsx
```diff
@@ -12,8 +12,6 @@ Honeybadger.configure({
   revision: 'git SHA/project version',
   component: 'example_comonent',
   action: 'example_action',
-  onerror: true,
-  onunhandledrejection: true,
```
shalvah, line 16: Noticed these aren't used anywhere in the code so I removed them. Relics from a previous version, maybe?

shalvah:
Also: while working on this, I noticed there's nothing in the docs about working with TypeScript or ES Modules. For instance, should users do `import Honeybadger from '...'` or `import * as Honeybadger from '...'`. It's something that would be worth addressing soon.

test-d/browser.test-d.tsx
joshuap: Ah yeah, these got renamed to `enableUncaught` and `enableUnhandledRejection`.

joshuap (approved):
This is beautiful. Thanks!

joshuap:
@shipjs prepare

shalvah:
👍

image

Issue #624, opened 2021-08-12
Include source snippet in node backtraces when available


joshuap:
See how this works in Ruby:

https://github.com/honeybadger-io/honeybadger-ruby/blob/master/lib/honeybadger/backtrace.rb#L83

Here's the schema:

https://github.com/honeybadger-io/honeybadger-schemas/blob/master/exceptions/v1/notices.json#L165

joshuap:
> @joshuap I'm having some trouble reading ruby code 😅
> Is it possible to post here an example JSON of the request that is supposed to be sent?
> The schema also does not help very much, it only states that `source` is an object.

Sure, check out the sample backtrace here:

https://docs.honeybadger.io/api/reporting-exceptions/#sample-payload

It's an object keyed by `line_number => line_contents` from the source file. Each line of the backtrace can have a `source` object if one is available.

Issue #327, opened 2020-03-14
Return a promise from Honeybadger.notify


joshuap:
We now support this:

```javascript
Honeybadger.notify('testing', {
  afterNotify: (err, notice) => console.log(err, notice)
});
```

Could we return a promise, so it could look like this?

```javascript
Honeybadger.notify('testing').then(notice => {
  console.log(notice);
}).catch(error => {
  console.log(error);
});
```

ParamagicDev:
@joshuap I hacked around and have a rough idea of how this would look. I have it on this branch here: https://github.com/honeybadger-io/honeybadger-js/tree/feat/return-a-promise-from-notify

There are a couple problems with trying to overload `notify` to handle both cases, first: this function returns booleans and users could potentially be using those booleans to determine various paths IE: 

```js
const notifyBool = Honeybadger.notify("testing")
if (notifyBool) {
  doX()
} else {
  doY()
}
```

The other issue is that any node based user will now have to catch any errors or they will get an `unhandledPromiseException` for not `catching` errors.

Perhaps instead of trying to make notify work like a promise (its hard because it doesnt use the traditional callback path, it uses a hook based approach) maybe we could split notify between the following:

```js
Honeybadger.notify('testing', {
  afterNotify: (err, notice) => console.log(err, notice)
});

Honeybadger.notifyAsync("testing").then((notice) => { console.log(notice) }).catch((err) => console.log(err))
``` 

This would allow both implementations coexist. Obviously extracting out the behavior of `Honeybadger.notify()` into reusable chunks will be key as well. Let me know how youd like to move forward on this!

joshuap:
@ParamagicDev thanks for the research, this is helpful! I'll think about this one a bit more. In the meantime let's continue on other issues. 👍

subzero10:
Hey @joshuap, I think we have to re-visit the idea of having an async version of `notify`, see #682.

subzero10:
This has been released with v4 (currently in beta).

PR #647, opened 2021-10-15
#624 Include source snippet in node backtraces when available

subzero10:
## Status
**READY**

## Description
Adds method `addSourceToBacktrace` in `client.ts`.
Used only in `server.ts` (nodejs) to send a snippet of the source code where the error occurred to honeybadger.

## Related PRs
n/a

## Todos
- [x] Fix eslint
- [x] Decide on solution: callbacks + nodejs only
- [x] Remove reference to getSourceCode for browser client
- [x] Tests
- [ ] Documentation

## Steps to Test or Reproduce
```bash
> git pull --prune
> git checkout backtrace_source_code#624
> npm i
> cd examples/express
> npm i
> HONEYBADGER_API_KEY=YOUR_API_KEY node index
> # open new terminal
> curl --url http://localhost:3000/fail

```


joshuap (commented):
> @joshuap I made the following assumptions. Please correct me if there is something I should not consider or if I left something out.
> 
> 1. I will always want to read the source code file, regardless if the module is running in browser or server.

I hadn't thought of this, but it could be a nice touch to fetch sources for the browser. The only issue I see is that most of that code is currently minified, and I'm not sure how useful it is. In any case, fetching browser sources isn't a hard requirement for #624, but I'm not against it.

> 2. I will parse the source code even it is minified or ugilfied.

I'm not sure how useful it would be to include minified source content. I'm not totally against it, but what would it look like? Also, how large would the resulting strings be? Most minified files I've seen are only one or two lines, so would including minified sources be basically the same as including the entire file?

> 3. The UI on Honeybadger shows only the source of the first backtrace, so I'm only parsing the source for the first item in the backtrace (after the `shift` is applied, see `makeBacktrace()`). This is simply for performance reasons.

I'd like to include the source snippet for each line of the backtrace, at least for node. Reading lots of files from the filesystem should be OK—I'm less certain about making lots of network requests for the source files in the browser, though.

> For the second point above, I think we could improve our solution by doing some post-processing on our backend after we receive the Honeybadger notice in case we have source maps uploaded from the user. I don't even know if this is possible as I haven't done it before (reading source code using source maps).

Right, we do currently apply source maps when they are available (for both node and browser). This is our recommended solution for browser users. I assume it's more common in node for the application code to be un-minified, and in that case, most users will not be uploading source maps (hence sending the source snippets in the error report itself). Is that the right assumption?

https://docs.honeybadger.io/lib/javascript/guides/using-source-maps/

In my experience, most browser users tend to use source maps, and don't really care if the minified sources are displayed in Honeybadger.

> To cover the above assumptions, I have come up with a few solutions:
> 
> * Attach the source as a first step in `__send()` method, which is implemented separately in browser and server. This seemed to be the simplest solution at first, but the `beforeNotifyHandlers` in server modify the file path (replace root with `[PROJECT_ROOT]`) and I would have to write additional logic just to revert to the original path.
> * Make `beforeNotifyHandlers` async (either by returning a promise or accepting a callback) and attach the source to the backtrace in such a handler (one for browser and one for server). I didn't go for this because I feel that `notice.backtrace.source` should be resolved in the `makeBacktrace()` function. If I went for this approach, I assume I would have to make `afterNotifyHandlers` async too.
> * So, the solution I'm proposing here is changing `makeBacktrace()` to be async and accept a new parameter, which is a function to fetch the source (either from filesystem or remotely from the network). This function (`getSourceFile`) is defined in `browser/util.ts` and `server/util.ts`. I decided to change `notify()` to return `Promise<boolean>` because I had to build the backtrace before calling `__send`.
> 
> Note: I noticed that we are using callbacks for asynchronous code. The code I've written is using promises. I think we could improve code quality and readability if we switch to async/await and promises. If you agree I can let the new code as is and perhaps we can create a new ticket to transform the code throughout the package to use async/await and promises. If not, I will change my code back to callbacks.

I've briefly reviewed the code so far and it's looking OK, but I don't have enough of a handle on it to decide if async/promises is the right way to get (yet). Can we discuss more? Check out this issue for a bit of extra context around switching `Honeybadger.notify` to return a promise: https://github.com/honeybadger-io/honeybadger-js/issues/327

@ParamagicDev was thinking that there may be some gotchas with changing the return of the current notify function that people are already using. If we *do* make that change, it would definitely need to be a major version release. I'm not against moving to promises as long as we can maintain [browser support](https://github.com/honeybadger-io/honeybadger-js/blob/master/test/integration/browsers.js) and we have a good migration path for users. That said, in general, users think of `Honeybadger.notify` as a "fire and forget" type of function, and I don't want them to have to think of extra things like error handling or delivery (at least in the default use case).

subzero10 (commented):
The PR appears to be large, but most of the changes are due to eslint.

joshuap (approved):
This looks good to me, thanks!

> What should we do about documentation? Is there something to update in the docs project?

I don't think we need any docs atm, since this is an automatic feature (there is no configuration necessary).

joshuap:
@subzero10 can you rebase against master?

subzero10:
> @subzero10 can you rebase against master?

I merged master and resolved conflicts.
PR is ready to merge!

Issue #648, opened 2021-10-19
Unable to get an error thrown from inside the handler after wrapping it in lambdaHandler


tweaktastic:
## What are the steps to reproduce this issue?
1. Create an APIGateway handler method for AWS lambda
2. Throw an error inside the handler.
3. Add a test to execute the handler to throw an error. The error never gets thrown.

I have my main handler as defined below:
```
import { APIGatewayProxyHandler } from 'aws-lambda';
import Honeybadger from '@honeybadger-io/js';
const handler: APIGatewayProxyHandler = async(event, context) => {
	//performs some logic or throw an error
	try { // blah } 
	catch(err) { throw new Error('Some Error); }
}

export default Honeybadger.lambdaHandler(handler);
```

In my tests I import this handler and want the test to fail

```
import handler form '../index.ts';
describe('Handler', () => {
  test('Error is handled', async () => {
      try { 
        // valid event and context are manually added fixtures
         await handler(validEvent, context);
         throw new Error('handler should throw an error');
      } catch(err) {
         expect((<Error>e).message).toBe('Some Error');
      }
  });
});
```

## What happens?
There are two things that are happening wrong.
- The handler code doesn't get executed synchronously during the tests. Any `console.log` that I add to the handler show up after the test has ended. I am using `Jest` for testing.
- The handler never throws the error as is being expected due to it being executed in an async manner.

## What were you expecting to happen?
I was expecting that my handlers and tests keep on working as they were before I started wrapping them in `lambdaHandler`.

## Any logs, error output, etc?
As stated above, the code works in an async manner. I see logs after the test has ended. The `await` doesn't seem to work.

## Any other comments?
It looks like the wrapper is not using `Promise`. It expects a callback to be passed to it to which is passes the `error` in case and error is produced. Please correct me if my understanding is wrong.

## What versions are you using?
**Operating System: Mac OS Big Sur 11.4 (20F71)** …
**Package Version: `"@honeybadger-io/js": "^3.2.5"`** …
**Browser Version: Node 14, Typescript, "aws-lambda": "^1.0.6"** …

tweaktastic:
Hi @subzero10 Thanks for the update regarding this. I tried to make this work as per your example that you posted yesterday itself.
```
import handler from './index';
describe('Handler', () => {
    test('Error is handled', (done) => {
        handler({}, {}, (err?: Error, resp?: any) => {
            if (!err) {
                done(new Error('should not reach here'))
                return
            }
[...]

PR #650, opened 2021-10-21
#648 Callback not called on lambda handler if notify returns false

subzero10:
## Status
**READY**

## Description
Our lambda handler implementation relies on `afterNotify` to call `callback(err)`.
`afterNotify` will not be called if the precondition checks fail in `notify()`.
Please see [ticket](https://github.com/honeybadger-io/honeybadger-js/issues/648) for more details.

## Related PRs
n/a

## Todos
- [x] Fix bug
- [x] Tests 

## Steps to Test or Reproduce
Create a lambda handler and test without providing an apiKey.
More info on this [ticket](https://github.com/honeybadger-io/honeybadger-js/issues/648).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased][latest]
+### Fixed
+- Call lambda handler callback if notify preconditions fail (#648)
 
 ## [3.2.5] - 2021-09-22
 ### Fixed
--- a/src/server/middleware.ts
+++ b/src/server/middleware.ts
@@ -49,27 +49,25 @@ function lambdaHandler(handler: LambdaHandler): LambdaHandler {
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     const hb = this
 
-    dom.on('error', function(err) {
-      hb.notify(err, {
+    const hbHandler = function (err?: Error) {
+      const willNotify = hb.notify(err, {
         afterNotify: function() {
           hb.clear()
           callback(err)
         }
       })
-    })
+      if (!willNotify) {
+        callback(err)
+      }
+    }
+
+    dom.on('error', hbHandler)
 
     dom.run(function() {
       process.nextTick(function() {
         Promise.resolve(handler.apply(this, args))
           .then(() => { hb.clear() })
-          .catch(function(err) {
-            hb.notify(err, {
-              afterNotify: function() {
-                hb.clear()
-                callback(err)
-              }
-            })
-          })
+          .catch(hbHandler)
       })
     })
   }.bind(this)
--- a/test/unit/core/client.test.ts
+++ b/test/unit/core/client.test.ts
@@ -194,15 +194,6 @@ describe('client', function () {
       expect(client.notify({})).toEqual(false)
     })
 
-    it('does not send notice when message is ignored', function () {
-      client.configure({
-        api_key: 'testing',
-        ignorePatterns: [/ignore/i]
-      })
-
-      expect(client.notify('you should ignore me')).toEqual(false)
-    })
-
     it('accepts options as first argument', function () {
       client.configure({
         apiKey: 'testing'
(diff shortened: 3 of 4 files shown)


joshuap (approved):
This looks good to me @subzero10, thanks!

One question—do you think that we should be calling the `beforeNotify` and `afterNotify` handlers in the `Honeybadger.notify` function itself if the preconditions fail? i.e., are users expecting the callbacks to run even if an error is not reported?

I think we could ship this fix either way, I'm just curious if this is something we should revisit in the future.

joshuap:
> > This looks good to me @subzero10, thanks!
> > One question—do you think that we should be calling the `beforeNotify` and `afterNotify` handlers in the `Honeybadger.notify` function itself if the preconditions fail? i.e., are users expecting the callbacks to run even if an error is not reported?
> > I think we could ship this fix either way, I'm just curious if this is something we should revisit in the future.
> 
> That's a good question. I think I'm leaning towards not calling them:
> 
> * If the preconditions fail -> don't call `beforeNotify` nor `afterNotify`
> * if `beforeNotify` returns `false` -> don't call `afterNotify`
> 
> I'm not sure I understand this:
> 
> > are users expecting the callbacks to run even if an error is not reported?
> 
> Are you referring to the `beforeNotify` and `afterNotify` callbacks? Or the `lambdaHandler` callback? I think it's OK to assume that users should not expect `beforeNotify` and `afterNotify` to be called if the report is not going to be sent. However, I think they expect the lambda handler's callback to be called in any case (report sent or not), which is what this PR is about.
> 
> What do you think?
> 
> (In any case, I will go ahead and merge this since you approved)

I was referring to the `beforeNotify` and `afterNotify` callbacks—I'm thinking about the case where a user wants to do something similar to what we're doing in `lambdaHandler`—basically "report this error, then do something else, like call a callback." That may not be a super common use case, but it would be more convenient to always call the callbacks in that case, so that the user doesn't have to add an extra conditional (like you did in this PR). 🤷

subzero10:
> > > This looks good to me @subzero10, thanks!
> > > One question—do you think that we should be calling the `beforeNotify` and `afterNotify` handlers in the `Honeybadger.notify` function itself if the preconditions fail? i.e., are users expecting the callbacks to run even if an error is not reported?
> > > I think we could ship this fix either way, I'm just curious if this is something we should revisit in the future.
> > 
> > 
> > That's a good question. I think I'm leaning towards not calling them:
> > 
> > * If the preconditions fail -> don't call `beforeNotify` nor `afterNotify`
> > * if `beforeNotify` returns `false` -> don't call `afterNotify`
> > 
> > I'm not sure I understand this:
> > > are users expecting the callbacks to run even if an error is not reported?
> > 
> > 
> > Are you referring to the `beforeNotify` and `afterNotify` callbacks? Or the `lambdaHandler` callback? I think it's OK to assume that users should not expect `beforeNotify` and `afterNotify` to be called if the report is not going to be sent. However, I think they expect the lambda handler's callback to be called in any case (report sent or not), which is what this PR is about.
> > What do you think?
> > (In any case, I will go ahead and merge this since you approved)
> 
> I was referring to the `beforeNotify` and `afterNotify` callbacks—I'm thinking about the case where a user wants to do something similar to what we're doing in `lambdaHandler`—basically "report this error, then do something else, like call a callback." That may not be a super common use case, but it would be more convenient to always call the callbacks in that case, so that the user doesn't have to add an extra conditional (like you did in this PR). 🤷

Got it. Hmm, indeed it could be a valid use case. Personally, I think it's OK as it is now, because the name `afterNotify` implies that a callback would be called _after_ Honeybadger is _notified_. However, we could provide a `reported` boolean value to the `afterNotify` callbacks that would indicate whether the error was reported or not. I'm not sure though, what do you think?

joshuap:
> > > > This looks good to me @subzero10, thanks!
> > > > One question—do you think that we should be calling the `beforeNotify` and `afterNotify` handlers in the `Honeybadger.notify` function itself if the preconditions fail? i.e., are users expecting the callbacks to run even if an error is not reported?
> > > > I think we could ship this fix either way, I'm just curious if this is something we should revisit in the future.
> > > 
> > > 
> > > That's a good question. I think I'm leaning towards not calling them:
> > > 
> > > * If the preconditions fail -> don't call `beforeNotify` nor `afterNotify`
> > > * if `beforeNotify` returns `false` -> don't call `afterNotify`
> > > 
> > > I'm not sure I understand this:
> > > > are users expecting the callbacks to run even if an error is not reported?
> > > 
> > > 
> > > Are you referring to the `beforeNotify` and `afterNotify` callbacks? Or the `lambdaHandler` callback? I think it's OK to assume that users should not expect `beforeNotify` and `afterNotify` to be called if the report is not going to be sent. However, I think they expect the lambda handler's callback to be called in any case (report sent or not), which is what this PR is about.
> > > What do you think?
> > > (In any case, I will go ahead and merge this since you approved)
> > 
> > 
> > I was referring to the `beforeNotify` and `afterNotify` callbacks—I'm thinking about the case where a user wants to do something similar to what we're doing in `lambdaHandler`—basically "report this error, then do something else, like call a callback." That may not be a super common use case, but it would be more convenient to always call the callbacks in that case, so that the user doesn't have to add an extra conditional (like you did in this PR). 🤷
> 
> Got it. Hmm, indeed it could be a valid use case. Personally, I think it's OK as it is now, because the name `afterNotify` implies that a callback would be called _after_ Honeybadger is _notified_. However, we could provide a `reported` boolean value to the `afterNotify` callbacks that would indicate whether the error was reported or not. I'm not sure though, what do you think?

`afterNotify` already supports an optional error as the first argument—what if we just made it an error, something like `new Error('API key is not configured')`?

subzero10:
Yeah, that could do it. Should I proceed to create a ticket and add to my backlog?

Issue #654, opened 2021-11-02
`afterNotify` handlers should be called even if notice is not sent to honeybadger


subzero10:
`afterNotify` handlers are not called if preconditions fail. We can call them and pass on an `error` as the first parameter.
Example: `new Error('API key is not configured')`

More on this [here](https://github.com/honeybadger-io/honeybadger-js/pull/650#issuecomment-956787935).

Note to self:
~- Should this be a major version release? What if there are users that rely on `afterNotify` as an indication that the error was indeed reported on Honeybadger? 🤔~

joshuap:
> * Should this be a major version release? What if there are users that rely on `afterNotify` as an indication that the error was indeed reported on Honeybadger? 🤔

I kind of doubt this, but even if they are, they *should* be handling the error argument already (since the error report can fail in other ways).

PR #659, opened 2021-11-16
#654 Always call afterNotify

subzero10:
## Status
**READY**

## Description
Issue #654.
Make sure to always call `beforeNotify` and `afterNotify` handlers, even if report is not sent to Honeybadger.

## Todos
- [x] Always call `afterNotify`
- [x] Always call `beforeNotify`
- [x] Lambda functions: should the callback always be called in `afterNotify`? (as it was before #650)
- [x] Tests (`afterNotify` in notice, multiple `afterNotify`, `afterNotify` in notice and in client instance, `afterNotify` in lambda functions)

## Steps to Test or Reproduce
1. Create an HB instance without setting `apiKey`.
2. Set an `afterNotify` handler: client.afterNotify((err) => console.log(err.message))
3. Call client.notify('should not send report')
4. The `afterNotify` handler should be called with an error saying that `apiKey` was not set.

Front logo [Front conversations](https://app.frontapp.com/open/top_3u7vl)


src/core/util.ts
```diff
@@ -93,6 +93,10 @@ export function runBeforeNotifyHandlers(notice, handlers: BeforeNotifyHandler[])
 }
 
 export function runAfterNotifyHandlers(notice, handlers: AfterNotifyHandler[], error = undefined): boolean {
+  if (notice && notice.afterNotify) {
+    handlers.unshift(notice.afterNotify)
```
joshuap, line 97: Do you want to duplicate `handlers` before mutating it? I believe that's why we were doing the `Array.prototype.slice.call` dance. I don't know if it's necessary, though.

joshuap:
Will `beforeNotify` handlers also always be called?

src/core/util.ts
subzero10: I don't see why they were being duplicated though 🤷 .
One theory is that the array could be mutated from the initial call to `notify` until the `runAfterNotifyHandlers` is called, but that scenario is unlikely to happen.
So, I decided to use them in the same way we use the `runBeforeNotifyHandlers()`.
Should I clone them just to be on the safe side?

subzero10:
> Will `beforeNotify` handlers also always be called?

Good question. At the moment, they are not always called.
For example, if a pre-condition fails (i.e. api key not set), they are not called.
`beforeNotify` handlers are used to add/modify data in `notice` and/or decide whether the `notice` will be sent to Honeybadger, right? If the `notice` is not going to be sent to Honeybadger, is there any point at all to call them?

On the other hand, you could argue that we are calling `afterNotify` without actually _notifying_ Honeybadger, so we should do the same for `beforeNotify`, regardless whether we intend to notify Honeybadger or not.

What do you think?

subzero10:
@joshuap This is ready, unless you want to discuss and modify based on the above.

src/core/util.ts
joshuap: Nope, I think it's fine as-is! I don't remember why they were duplicated either.

joshuap:
> > Will `beforeNotify` handlers also always be called?
> 
> Good question. At the moment, they are not always called. For example, if a pre-condition fails (i.e. api key not set), they are not called. `beforeNotify` handlers are used to add/modify data in `notice` and/or decide whether the `notice` will be sent to Honeybadger, right? If the `notice` is not going to be sent to Honeybadger, is there any point at all to call them (and waste resources)?
> 
> On the other hand, you could argue that we are calling `afterNotify` without actually _notifying_ Honeybadger, so we should do the same for `beforeNotify`, regardless whether we intend to notify Honeybadger or not.
> 
> What do you think?

I think we should run both sets of hooks always. That's what we do in Ruby—although, we also [stop running the `before_notify` hooks if one returns `false`/ignores the notice](https://github.com/honeybadger-io/honeybadger-ruby/blob/5650ecbd5d5a0e8db9324aa01f9fda3ee1e20cc3/lib/honeybadger/agent.rb#L147). I think it probably makes more sense to run all of the hooks in JavaScript and skip the notice if one or more returned false.

joshuap (commented):
This is looking good. Let's make the `beforeNotify` handlers change (if you think it makes sense) in this PR, since all of this logic sort of goes together.

src/server/middleware.ts
```diff
@@ -45,28 +45,37 @@ type LambdaHandler = (event: unknown, context: unknown, callback: unknown) => vo
...
+     * we don't have to do that if the handler has third callback function parameter,
+     * because it will be called directly from inside the handler.
+     */
     const shouldInvokeCallbackExplicitly = handler.length < 3
 
+    /**
+     * this flag prevents an infinite loop in case of an error thrown
+     * inside the domain error handler {@link hbHandler}
+     */
+    let domainErrorHandlerAlreadyCalled = false;
```
subzero10, line 60: @joshuap Just making sure you saw this.

subzero10:
> This is looking good. Let's make the `beforeNotify` handlers change (if you think it makes sense) in this PR, since all of this logic sort of goes together.

Done. It makes sense to apply this in Javascript if the behavior already exists in Ruby.

Note: There is a potential issue with current integrations. From now on, `beforeNotify` handlers can be called without a notice object (in case someone calls `honeybadger.notify({})`). I documented this change in Typescript by modifying the `notice` in `BeforeNotifyHandler` to be optional:
```
export interface BeforeNotifyHandler {
    (notice?: Notice): boolean | void;
}
```

joshuap (approved):
Looks good to me. This may be a major release due to the changed behavior—I'll have to think about that. I may also release a beta version first so that we can try it out first.

subzero10:
> Looks good to me. This may be a major release due to the changed behavior—I'll have to think about that. I may also release a beta version first so that we can try it out first.

Any updates on this? Have you decided whether it is going to be a major release?

joshuap:
@subzero10 yeah, I'm leaning towards a minor release (3.3), after I finish up #670. I'll probably release a beta in December and do the actual release in January.

PR #663, opened 2021-11-23
Log errors in development mode

### LINKED ISSUE #661: Logging errors in development mode
Some users want to log errors rather than report them in development mode. Currently [we do `console.debug`](https://github.com/honeybadger-io/honeybadger-js/blob/3d9312daaf7c570888cbf53c8c435e9a42d595b1/src/core/client.ts#L146) when skipping errors in development mode, but we don't include the original error in the output, and the `debug` level is easy to miss because it's usually suppressed by default.

Should we be logging the original error, i.e. with a `console.error` call? Fwiw, I'm not totally sure myself, because uncaught errors do appear in the console as far as I know. 

Another option could be to let users do their own logging with a `beforeNotify`/`afterNotfiy` handler (now that we're changing these callbacks to always run).

OR, we could add the error object to the current logging, and then improve the docs so that people realize they can enable debug logs to see more output.

Front logo [Front conversations](https://app.frontapp.com/open/top_3sie9)

joshuap:
## Status
**WIP**

## Description
This should cause errors (notices) to be logged in development mode. With this approach, the full notice object is logged rather than the original error object (which could also be a string, object, etc.)

Closes #661

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -137,23 +137,23 @@ export default class Client {
   }
 
   notify(noticeable: Noticeable, name: string | Partial<Notice> = undefined, extra: Partial<Notice> = undefined): boolean {
+    const notice = this.makeNotice(noticeable, name, extra)
+    if (!notice) {
+      return false
+    }
+
     if (this.config.disabled) {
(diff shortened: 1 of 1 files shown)


subzero10:
Looks good to me.

PR #665, opened 2021-11-29
docs#102 Chrome Extension Example Project

subzero10:
## Status
**READY**

## Description
Example project for Chrome Extension

## Related PRs
[docs](https://github.com/honeybadger-io/docs/pull/147)

## Todos
- [x] Create project
- [x] Documentation

## Steps to Test or Reproduce
See `examples/chrome-extension/README.md`


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased][latest]
 ### Added
-- Nodejs: Include source snippet in backtraces when available (#624) 
+- Nodejs: Include source snippet in backtraces when available (#624)
```
joshuap, line 9: I removed the changelog line in https://github.com/honeybadger-io/honeybadger-js/pull/665/commits/1a35edb9fd0a5f33b77911a1b8e0729c1d98f413 — we can skip the changleog for this PR since it's not actually changing the lib.

examples/chrome-extension/vendor/honeybadger.min.js
```diff
@@ -0,0 +1,1456 @@
+(function (global, factory) {
```
joshuap, line 1: I'm curious if there is a way to use honeybadger.min.js from our CDN? Either just load it from there directly in the script tag in the extension, OR add a little `wget` script to download it to `vendor/` (and maybe add a `.gitkeep` so that the directory already exists)?

joshuap: Basically, I want to reduce any long-term maintenance overhead related to example projects.

joshuap (commented):
Looking good—a few comments.

joshuap:
I tried out the README instructions and it worked!

image

examples/chrome-extension/vendor/honeybadger.min.js
subzero10: I agree with you. Actually, I tried loading the file directly from our CDN, but ran into security issues with chrome extensions. Plus, it is not the recommended practice, as noted in the [How to comply with CSP] (https://developer.chrome.com/docs/apps/contentSecurityPolicy/#how) section of chrome development guides:

>All JavaScript and all resources should be local (everything gets packaged in your Chrome App).

I like the `wget` script. I can try that! Is there a CDN url that always points to the latest version?

subzero10: Done. I added a script to download (I used `curl`) the browser client and modified the `README.md` file. Let me know what you think.

joshuap (approved):
Looks good, thanks!

christianopaets:
@subzero10 please give an example how to use it inside background script. I'm trying to migrate to Manifest V3 that uses service workers, and there is no window object inside it, so it crashe
Screenshot 2022-01-20 at 17 08 26
s

subzero10:
Hey @christianopaets, thank you for reporting this error.
This looks to be a bug that we will have to fix. I have created an [issue](https://github.com/honeybadger-io/honeybadger-js/issues/686) to track it. I will take a look in the coming days.

PR #670, opened 2021-12-10
Improve dev-mode logging

### LINKED ISSUE #661: Logging errors in development mode

joshuap:
Debug logs are now logged at the default level in Chrome, so that you
don't have to enable verbose logging to see them (you still need to
configure the `debug: true` option.

Also changed a few of the log/error messages and used separate messages
for the development mode notice vs. disabled via `reportData: false`.

Closes #661

Edit: this also includes the notice in log messages, when available:

image

### DIFF
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -145,12 +145,12 @@ export default class Client {
     let preConditionError: Error = null
     const notice = this.makeNotice(noticeable, name, extra)
     if (!notice) {
-      this.logger.debug('could not make notice')
-      preConditionError = new Error('could not make notice')
+      this.logger.debug('failed to build error report')
+      preConditionError = new Error('failed to build error report')
     }
 
     if (!preConditionError && this.config.reportData === false) {
-      this.logger.debug('skipping notice: honeybadger.js is disabled', notice)
+      this.logger.debug('skipping error report: honeybadger.js is disabled', notice)
       preConditionError = new Error('honeybadger.js is disabled')
     }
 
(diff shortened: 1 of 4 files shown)


src/core/client.ts
```diff
@@ -138,38 +142,32 @@ export default class Client {
...
     }
 
-    // we need to have the source file data before the beforeNotifyHandlers,
-    // in case they modify them
-    const sourceCodeData = notice ? notice.backtrace.slice() : null
-
     const beforeNotifyResult = runBeforeNotifyHandlers(notice, this.__beforeNotifyHandlers)
     if (!preConditionError && !beforeNotifyResult) {
-      preConditionError = new Error('Will not send error report, beforeNotify handlers returned false')
+      this.logger.debug('skipping notice: beforeNotify handlers returned false', notice)
```
subzero10, line 169: Should we make sure to always have the same vocabulary used? Here it refers to the honeybadger notice as `notice` but in lines 158 and 163 it says `report`.

Issue #677, opened 2021-12-28
AWS Lambda Integration: async handler


subzero10:
### Discussed in https://github.com/honeybadger-io/honeybadger-js/discussions/674

Improve implementation of callback based and async (promise based) lambda handlers.

- Modify lambdaHandler to return async handler instead of callback based handler.

Question: Should this be a major release?

shalvah:
> Should this be a major release?

By itself, hopefully not. But if we go ahead to replace the callback code with promises, then perhaps.

subzero10:
> > Should this be a major release?
> 
> By itself, hopefully not. But if we go ahead to replace all the callback code in the package with promises, then perhaps.

We are not replacing all code with promises in this PR but we are going to be returning a promise-based lambda handler instead of callback-based.

subzero10:
This has been released with v4 (currently in beta).

PR #680, opened 2021-12-30
Improve `Honeybadger.lambdaHandler()` to return async handler

subzero10:
## Status
**READY**

## Description
This PR changes the result of `Honeybadger.lambdaHandler()` to be a promise based (async) handler instead of the outdated callback based handler.

Discussion: https://github.com/honeybadger-io/honeybadger-js/discussions/674
Issue: https://github.com/honeybadger-io/honeybadger-js/issues/677

## Todos
- [x] Return async handler instead of call back based
- [x] Add proper type definitions to the handler (from `@types/aws-lambda`)
- [x] Add example project for AWS Lambda
- [x] [Honeybadger.notifyAsync()](https://github.com/honeybadger-io/honeybadger-js/pull/684)
- [x] Tests - Make sure new and existing tests pass
- [x] Test manually with example project
- [ ] ~~AsyncLocalStorage (or something equivalent)~~ Separate issue: #688 
- [x] After release: Modify `examples/aws-lambda/package.json` to install `@honeybadger-io/js` latest release (it currently installs from `aws_lambda_async_#677`)
- [ ] After merge: Update AWS projects to point to aws lambda example. Discussion [here](https://github.com/honeybadger-io/honeybadger-js/pull/680/files#r788020748).  

## Steps to Test or Reproduce
See `examples/aws-lambda/README.md`


subzero10:
I know we haven't reached to a conclusion on this, but based on the implementations of [Sentry](https://github.com/getsentry/sentry-javascript/blob/master/packages/serverless/src/awslambda.ts) and [Bugsnag](https://github.com/bugsnag/bugsnag-js/blob/next/packages/plugin-aws-lambda/src/index.js), I decided to drop the usage of `domain` and simply opt for a try/catch block. What do you think? @shalvah @joshuap

test/unit/server/middleware.server.test.ts
```diff
@@ -116,448 +113,4 @@ describe('Express Middleware', function () {
...
-                    setTimeout(() => {
-                        expect(callCount).toEqual(1)
-                        resolve()
-                    }, 500)
-                }
-
-                const handler = client.lambdaHandler(function(_event, _context, _callback) {
-                    setTimeout(function() {
-                        throw new Error("Badgers!")
-                    }, 0)
```
subzero10, line 556: The new _lambdaHandler_ implementation does not cover this case:
- An unhandled error is thrown in `setTimeout` and not passed on to the `callback` function.

However, it should be caught by the `unhandledException` plugin and report to Honeybadger as expected.

src/server/middleware.ts
```diff
@@ -39,64 +39,3 @@ function errorHandler(err: Noticeable, req: Request, _res: Response, next: NextF
   })
   return next(err)
 }
-
-type LambdaHandler = (event: unknown, context: unknown, callback: unknown) => void|Promise<unknown>
```
subzero10, line 43: I moved this block of code to its own file.

shalvah:
Well done @subzero10 . I've got some thoughts, but I'm currently on holiday, so it'll have to be next week

package.json
```diff
@@ -42,13 +42,14 @@
...
   "dependencies": {
-    "@types/express": "^4.17.13",
     "stacktrace-parser": "^0.1.10"
   },
   "devDependencies": {
     "@rollup/plugin-node-resolve": "^13.0.6",
     "@rollup/plugin-replace": "^3.0.0",
     "@rollup/plugin-typescript": "^8.3.0",
+    "@types/aws-lambda": "^8.10.89",
+    "@types/express": "^4.17.13",
```
shalvah, line 52: Good catch, moving these to dev deps.

src/server/aws_lambda.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * async (event) => async handler
+ * async (event, context) => async handler
+ * (event, context, callback) => sync handler
+ * @param handler
+ */
+function getAsyncHandler(handler: APIGatewayProxyHandler): AsyncHandler {
+    if (handler.length > 2) {
+        return ((event, context) => {
+                return new Promise((resolve, reject) => {
```
shalvah, line 24: I like this approach, but I wonder if they're identical? For instance, [`context.callbackWaitsForEmptyEventLoop`](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html)—if a user sets it in a sync handler, will AWS still respect it, considering we aren't using the callback anymore?

shalvah:
I agree with you that a try/catch is enough, but I wonder about context (which I think is what domains were used for). For instance, in this `handler.js`:

```js
const hb = require('honeybadger');

module.exports.handler1 = () => {
  hb.context({a: "1"});
});

module.exports.handler1 = () => {
  hb.context({a: "2"});
});
```

can we be sure these context calls won't clash if both these functions get executed, considering the same Lambda instance might be reused? I think WWE should add a test for that. If so, we could (should) use AsyncLocalStorage and polyfills to fix it.

src/server/aws_lambda.ts
subzero10: > I like this approach, but I wonder if they're identical?

Can you please elaborate? I'm not sure I understand. The goal is to always return an async handler.

> will AWS still respect it, considering we aren't using the callback anymore?

That's a good question. My guess is that AWS will ignore it since that property only has effect in non-async handlers.

subzero10:
> I agree with you that a try/catch is enough, but I wonder about context (which I think is what domains were used for). For instance, in this `handler.js`:
> 
> ```js
> const hb = require('honeybadger');
> 
> module.exports.handler1 = () => {
>   hb.context({a: "1"});
> });
> 
> module.exports.handler1 = () => {
>   hb.context({a: "2"});
> });
> ```
> 
> can we be sure these context calls won't clash if both these functions get executed, considering the same Lambda instance might be reused? I think we should add a test for that. If so, we could (should) use AsyncLocalStorage and polyfills to fix it.

I was worried that there would be a performance penalty with `AsyncLocalStorage`, but it seems this is not the case anymore: 
- https://github.com/PhakornKiong/ALS-CLS-Benchmark
- https://github.com/nodejs/node/issues/34493

Still, I am wondering if we could just use a dictionary and store _honeybadger context_ using key [`context.awsRequestId`](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html). Wouldn't that be enough?

shalvah:
The dictionary approach should work, but wouldn't that possibly lead to a memory leak? We'll keep adding to the dictionary and never remove unless there's a crash. I don't know how long Lambda instances stick around for, though...🤔

subzero10:
> The dictionary approach should work, but wouldn't that possibly lead to a memory leak? We'll keep adding to the dictionary and never remove unless there's a crash. I don't know how long Lambda instances stick around for, though...🤔

Good question. I wonder if there's a way to find that out. I will do some googling.

Also, question that came up now: Even if the lambda instance is re-used, would it be the same nodejs process? If not, we wouldn't need to do anything here right?

shalvah:
Yeah, if it's a different process,, we would be safe. But considering that the reused instance already has the state initialised, I'm leaning towards same process.

shalvah:
Hmm. https://docs.aws.amazon.com/lambda/latest/dg/runtimes-context.html#runtimes-lifecycle-shutdown

Screenshot_20220106-115845~2

examples/aws-lambda/README.md
```diff
@@ -0,0 +1,27 @@
+# Honeybadger AWS Lambda - Serverless Example
```
joshuap, line 1: Here are a few example projects that we may need to update after this is done:

https://github.com/honeybadger-io/serverless-quickstart
https://github.com/honeybadger-io/honeybadger-lambda-node

I think it would make sense to update serverless-quickstart and deprecate/archive the honeybadger-lambda-node repo (which seems similar to this example?)

joshuap: We may also want to link to this example project in the serverless-quickstart readme.

joshuap: I'd also consider archiving both of those repos if they are both redundant. :)

src/server/aws_lambda.ts
```diff
@@ -0,0 +1,58 @@
...
+        }) as AsyncHandler
+    }
+
+    return (handler as AsyncHandler)
+}
+
+export function lambdaHandler(handler: APIGatewayProxyHandler): AsyncHandler {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const hb = this
+    const asyncHandler = getAsyncHandler(handler)
```
joshuap, line 42: The idea here is that everyone still calls `lambdaHandler`, but that function now converts the handler to async if it's the old style? If so, is this totally transparent to end-users?

joshuap (commented):
I read through the code, looking good. I added a few questions. @subzero10 are you still having issues with the `setTimeout` errors? I think that may be related to why we were using domains.

src/server/aws_lambda.ts
subzero10: It should be transparent, but we cannot be 100% sure:
If end-users are using the result of `Honeybadger.lambdaHandler` in their test code, they will be getting unexpected errors since this function does not take a callback as a third parameter anymore and it returns a promise.

examples/aws-lambda/README.md
subzero10: >https://github.com/honeybadger-io/honeybadger-lambda-node

I think this is redundant now that we have the aws-lambda example project.

>https://github.com/honeybadger-io/serverless-quickstart

This appears outdated. If it's still being used from the team as a boilerplate, it makes sense to update (and maybe link to the aws-lambda example project), otherwise it could also be archived. `serverless.com` provides a dashboard and a cli so you don't have to log into AWS Console to configure env variables.
However, I should note that our serverless integration only supports AWS Lambda. In the future, we could extract the aws-lambda implementation to a separate _serverless_ package/repository and add support for more cloud providers (azure, google cloud).

subzero10:
> I read through the code, looking good. I added a few questions. @subzero10 are you still having issues with the `setTimeout` errors? I think that may be related to why we were using domains.

Yes, I am getting this error when throwing an uncaught error in a setTimeout: 
>Unknown application error occurred

For the rest, there are times where I get this error:
>Error report failed: an unknown error occurred. message=Client network socket disconnected before secure TLS connection was established

I think it's because I throw the error before the http request to Honeybadger is completed.

I'm still investigating.

test/unit/server/middleware.server.test.ts
subzero10: Tested with the latest changes and it the setTimeoutError was reported as expected.

subzero10:
>can we be sure these context calls won't clash if both these functions get executed, considering the same Lambda instance might be reused? I think we should add a test for that. If so, we could (should) use AsyncLocalStorage and polyfills to fix it.

@shalvah @joshuap

Should we implement this in this PR? I'm thinking that this potential issue already exists with the current lambdaHandler implementation, so I'm wondering if `AsyncLocalStorage` is out of the scope of this PR: callback based to promise based lambda handler.
I'm only suggesting a separate PR/ticket because I like PRs to be as small as possible.

subzero10:
> > can we be sure these context calls won't clash if both these functions get executed, considering the same Lambda instance might be reused? I think we should add a test for that. If so, we could (should) use AsyncLocalStorage and polyfills to fix it.
> 
> @shalvah @joshuap
> 
> Should we implement this in this PR? I'm thinking that this potential issue already exists with the current lambdaHandler implementation, so I'm wondering if `AsyncLocalStorage` is out of the scope of this PR: callback based to promise based lambda handler. I'm only suggesting a separate PR/ticket because I like PRs to be as small as possible.

I tried to start work on AsyncLocalStorage, but it seems that it won't be that simple to implement. We will have to do some refactoring to externalize the store for context and breadcrumbs.

shalvah:
Will review over the weekend @subzero10 . Sorry for the delay.

src/server/aws_lambda.ts
shalvah: Yup, this is my main worry. Can't we write `lambdaHandler` such that it returns something isomorphic to what was given?

```js
function lambdaHandler(handler) {
  if (handler.length == 3) {
    return (event, context, cb) => {
      try {
        handler(event, context, cb);     
      } catch (e) {
        hb.notify(e);
        cb(e);
      }
    };
  } else {
  return (event, context) => {
    return new Promise(...); 
  }
}
```

Not ideal, but this way we avoid breaking any expectations.

_Aargh, Node.js is such a mess, what with error handling split across callbacks, promises and regular exceptions.😩_

shalvah (commented):
Great work @subzero10, just a few concerns.

examples/aws-lambda/README.md
joshuap: I think I'll archive both of these projects after this PR is finished. I can update their readmes to point to the new example directory in this repo.

joshuap (commented):
I've reviewed everything so far and am following along. The code looks good to me, though I think @shalvah is more in the headspace of the problem. I'll take another look once #684 gets merged in.

src/server/aws_lambda.ts
subzero10: Done. Now I have to think about the type definitions. I want to avoid `any`, but it may not be possible.

shalvah: Hmm, maybe something like this?

```typescript
function lambdaHandler<T extends SyncHandler | AsyncHandler>(handler: T): T {
  ...
}
```

subzero10: Considering this [ticket](https://github.com/honeybadger-io/honeybadger-js/issues/682#issuecomment-1013816752), it might not be enough. I will check if `@types/aws-lambda` has something generic enough to support all their handlers, otherwise I believe we'll have to go with `unknown`.

shalvah: Hmm, why wouldn't it be enough? 🤔 You could define `SuncHandler` and `AsyncHandler` either generically, or with `any`/`unknown`. _That_ could get complex, if you're trying to support all the AWS event types and return values (unless they have a parent type). But for the `lambdaHandler` function itself, it simply needs to return the same type as it gets.

Also, hehe, it's quite timely that we started working on this at this time😀. I suspect that user's timeout problems were because of the Promise-returning handler being converted into a callback handler.

shalvah: Yeah, looking at the code again, I think we just need to redefine the `event` parameter and return types of the handlers to be more generic. Hopefully the AWS event types and return types inherit from a base type, so we can use that.

src/server/aws_lambda.ts
subzero10: This won't be a problem anymore since we are returning the same type of handler we receive.

src/server/aws_lambda.ts
subzero10: > Also, hehe, it's quite timely that we started working on this at this time😀. I suspect that user's timeout problems were because of the Promise-returning handler being converted into a callback handler.

Yes!

> Yeah, looking at the code again, I think we just need to redefine the event parameter and return types of the handlers to be more generic. Hopefully the AWS event types and return types inherit from a base type, so we can use that.

Done! I was looking to find a way so that the return type of `Hb.lambdaHandler` would be narrowed to the sync or async version based on the input handler, but I couldn't make it. I will try this again at some other point.

subzero10:
@shalvah @joshuap I think this is ready. I created [this issue](https://github.com/honeybadger-io/honeybadger-js/issues/688) to tackle context isolation.

examples/aws-lambda/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "aws-lambda",
+  "version": "1.0.0",
+  "description": "",
+  "main": "handler.js",
+  "dependencies": {
+    "@honeybadger-io/js": "^3.2.8",
```
subzero10, line 7: The example project points to a _yet-to-be-released_ version. It should be OK as long as `v3.2.8` includes the latest aws-lambda work.

subzero10:
@joshuap @shalvah I added `notifyAsync` documentation, see PR [here](https://github.com/honeybadger-io/docs/pull/150).

shalvah (approved):
Nice work! I'll try to take it for a spin on a project soon.

subzero10:
@joshuap I think this is safe to merge and release, unless you have any other feedback.

examples/aws-lambda/package.json
joshuap: I have to decide, but I'm thinking this may be a new major 4.0.0 release, since we're changing a few things (or at least 3.3.0).

subzero10: We are not introducing breaking changes so not sure about the `4.0.0` release, but `3.3.0` makes sense.

Issue #682, opened 2022-01-09
AWS Lambda - Serverless Framework - Nestjs -  Getting Timed Out


sedhuait:
## What are the steps to reproduce this issue?

I have hosted a REST API with AWS Components like API GATEWAY & LAMBDA using Nestjs and Serverless Framework. 

```
// HoneyBadger init code 
import Honeybadger from '@honeybadger-io/js';
declare global {
  // eslint-disable-next-line no-var
  var Honeybadger: Honeybadger;
}
if (!global.Honeybadger) {
  Honeybadger.configure({
    reportData: true,
    apiKey: process.env.HONEYBADGER_SECRET,
    environment: process.env.NODE_ENV,
  });
  console.log('Loading Honeybadger', !!process.env.HONEYBADGER_SECRET);
  global.Honeybadger = Honeybadger;
}

``` 
So to integrate with Honeybadger Lambda Handler I have made the changes below. It had typescript error and I refactored it as follows 
![image](https://user-images.githubusercontent.com/1628931/148668863-058ac9a6-be8a-46ae-a6a2-90c41a34e815.png)

After refacorting, 
```
export const handler: Handler = Honeybadger.lambdaHandler(async (event, context) => {
  cachedServer = await bootstrapServer();
  return proxy(cachedServer, event as APIGatewayProxyEvent, context as Context, 'PROMISE').promise;
})
``` 

## What happens?
This setup works properly in serverless offline start "sls offline start" 

But When I deploy this on AWS Lambda, the functions are getting timed out after 30 seconds. Its getting stuck in `Honeybadger.lambdaHandler` as I see the success response in my middleware. 
Kindly check my aws metrics dashboard below. Spikes are during my Honeybadger integration test. After that, I removed it, and it came back to normal.

![image](https://user-images.githubusercontent.com/1628931/148668881-4fd4e769-9fd6-4cf2-893d-db4270c81125.png)

## What were you expecting to happen?
It should have got the proper response. 

## Any logs, error output, etc?

![image](https://user-images.githubusercontent.com/1628931/148668891-6ae6e3a5-9889-4905-b357-9d00f8d85197.png)

![image](https://user-images.githubusercontent.com/1628931/148668894-7eee1ebd-1d32-4a07-bdb7-b99786b34ec3.png)
![image](https://user-images.githubusercontent.com/1628931/148668895-58770a51-c5d0-46ef-9d2b-cac8b8b19c4b.png)

## What versions are you using?
[...]

PR #684, opened 2022-01-15
#327 notifyAsync

subzero10:
## Status
**READY**

## Description
Related Issue: #327 

I came up with a simple implementation of notify that returns a promise (notifyAsync).
It all became simple because now we always call the `afterNotify` hook, so I used it to know when to resolve/reject the promise.
I decided to go for this because of our recent work on AWS Lambda (see [here](https://github.com/honeybadger-io/honeybadger-js/issues/682#issuecomment-1013654404) and [here](https://github.com/honeybadger-io/honeybadger-js/pull/680#issuecomment-1008439277)). The _fire-and-forget_ approach does not work well with Lambda functions.

Let me know what you think.

Note: The target branch is [aws_lambda_async_#677](https://github.com/honeybadger-io/honeybadger-js/tree/aws_lambda_async_%23677)

## Todos
- [x] Tests
- [ ] Documentation

### DIFF
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -206,39 +206,12 @@ export default class Client {
    */
   notifyAsync(noticeable: Noticeable, name: string | Partial<Notice> = undefined, extra: Partial<Notice> = undefined): Promise<void> {
     return new Promise((resolve, reject) => {
-      const applyAfterNotify = (partialNotice: Partial<Notice>) => {
-        const originalAfterNotify = partialNotice.afterNotify
-        partialNotice.afterNotify = (err?: Error) => {
-          originalAfterNotify?.call(this, err)
-          if (err) {
-            return reject(err)
-          }
-          resolve()
+      this.afterNotify(err => {
+        if (err) {
+          return reject(err)
         }
-      }
-
-      // We have to respect any afterNotify hooks that come from the arguments
-      let objectToOverride: Partial<Notice>
-      if ((noticeable as Partial<Notice>).afterNotify) {
-        objectToOverride = noticeable as Partial<Notice>
-      }
-      else if (name && (name as Partial<Notice>).afterNotify) {
-        objectToOverride = name as Partial<Notice>
-      }
-      else if (extra && extra.afterNotify) {
-        objectToOverride = extra
-      }
-      else if (name && typeof name === 'object') {
-        objectToOverride = name
-      }
-      else if (extra) {
-        objectToOverride = extra
-      }
-      else {
-        objectToOverride = name = {}
-      }
-
-      applyAfterNotify(objectToOverride)
+        resolve()
+      })
       this.notify(noticeable, name, extra)
     })
   }
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -198,6 +198,51 @@ export default class Client {
     return true
   }
 
+  /**
+   * An async version of {@link notify} that resolves only after the notice has been reported to Honeybadger.
+   * Implemented using the {@link afterNotify} hook.
+   * Rejects if for any reason the report failed to be reported.
+   * Useful in serverless environments (AWS Lambda).
+   */
+  notifyAsync(noticeable: Noticeable, name: string | Partial<Notice> = undefined, extra: Partial<Notice> = undefined): Promise<void> {
+    return new Promise((resolve, reject) => {
+      const applyAfterNotify = (partialNotice: Partial<Notice>) => {
+        const originalAfterNotify = partialNotice.afterNotify
+        partialNotice.afterNotify = (err?: Error) => {
+          originalAfterNotify?.call(this, err)
+          if (err) {
+            return reject(err)
+          }
+          resolve()
+        }
+      }
+
+      // We have to respect any afterNotify hooks that come from the arguments
+      let objectToOverride: Partial<Notice>
+      if ((noticeable as Partial<Notice>).afterNotify) {
+        objectToOverride = noticeable as Partial<Notice>
+      }
+      else if (name && (name as Partial<Notice>).afterNotify) {
+        objectToOverride = name as Partial<Notice>
+      }
+      else if (extra && extra.afterNotify) {
+        objectToOverride = extra
+      }
+      else if (name && typeof name === 'object') {
+        objectToOverride = name
+      }
+      else if (extra) {
+        objectToOverride = extra
+      }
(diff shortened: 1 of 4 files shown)


shalvah (approved):
Great! Love the test coverage. I do wonder, though—isn't there a global version of `afterNotify`? From [the docs](https://docs.honeybadger.io/lib/javascript/reference/configuration/#afternotify-handlers):

```js
Honeybadger.afterNotify((err, notice) => {
  if (err) { return console.log(`Honeybadger notification failed: ${err}`); }
  console.log(`Honeybadger notice: https://app.honeybadger.io/notice/${notice.id}`);
});
```

I didn't see that covered here.

subzero10:
> Great! Love the test coverage. I do wonder, though—isn't there a global version of `afterNotify`? From [the docs](https://docs.honeybadger.io/lib/javascript/reference/configuration/#afternotify-handlers):
> 
> ```js
> Honeybadger.afterNotify((err, notice) => {
>   if (err) { return console.log(`Honeybadger notification failed: ${err}`); }
>   console.log(`Honeybadger notice: https://app.honeybadger.io/notice/${notice.id}`);
> });
> ```
> 
> I didn't see that covered here.

Thanks!
I wanted this `afterNotify` hook to be the last to be called. 

And here's what I just realized:
- The `afterNotify` hook that you specify with `Honeybadger.notify()` is actually called first.
```
// util.ts
  if (notice && notice.afterNotify) {
    handlers.unshift(notice.afterNotify)
  }

  for (let i = 0, len = handlers.length; i < len; i++) {
    handlers[i](error, notice)
  }
  return true
}
```

I think I can just change the code to:
```
return new Promise((resolve, reject) => {
  this.afterNotify((err) => {
          if (err) {
            return reject(err)
          }
          resolve()
  })
  this.notify(noticeable, name, extra)
})

```
and skip all the block of code of _objectToOverride_. Then modify the tests to check that this hook is indeed called last.

shalvah:
Okay, that sounds cool.

Glad we're adding this. Hopefully we can even get rid of `afterNotify` in the future. It's a bad pattern, basically reimplementing continuations, which Node.js has supported since forever (callbacks, promises, events).

subzero10:
@joshuap Waiting for your feedback on this. We should also mention `notifyAsync` in the documentation, but I'm not sure how to. `notifyAsync` is a must for lambda functions, but optional server or browser.

test/unit/core/client.test.ts
```diff
@@ -331,6 +331,70 @@ describe('client', function () {
...
+      client.configure({
+        apiKey: 'testing'
+      })
+    })
+
+    it('resolves when configured', async () => {
+      await client.notifyAsync(new Error('test'))
+    })
+
+    it('calls afterNotify in noticeable', async () => {
```
shalvah, line 345: @subzero10 just to be covered,could we add a test for the global `afterNotify()`?

subzero10: Ah yes, forgot that one. Done!

subzero10:
> Glad we're adding this. Hopefully we can even get rid of `afterNotify` in the future. It's a bad pattern, basically reimplementing continuations, which Node.js has supported since forever (callbacks, promises, events).

Sounds like another topic for discussion!

shalvah (approved):

shalvah:
On second thought, since `notify` might be called by code the end user, I guess it's not a bad pattern at all.

src/core/client.ts
```diff
@@ -198,6 +198,24 @@ export default class Client {
...
 
+  /**
+   * An async version of {@link notify} that resolves only after the notice has been reported to Honeybadger.
+   * Implemented using the {@link afterNotify} hook.
+   * Rejects if for any reason the report failed to be reported.
+   * Useful in serverless environments (AWS Lambda).
+   */
+  notifyAsync(noticeable: Noticeable, name: string | Partial<Notice> = undefined, extra: Partial<Notice> = undefined): Promise<void> {
+    return new Promise((resolve, reject) => {
+      this.afterNotify(err => {
```
joshuap, line 209: Doesn't using `this.afterNotify` mean that the handler is added to the client's permanent `afterNotify` handlers? What would happen if `notifyAsync` is called twice? Wouldn't the previous handler still fire, in addition to the new one?

joshuap (commented):
Nice, I like the approach!

src/core/client.ts
shalvah: That's true. @subzero10 if the afterNotify handlers are exposed publicly (like process.listeners(event)`), we could ensure that the afterNotify function removes itself from the list when it's done.

subzero10: Correct. Good catch on noticing that @joshuap !
It now makes sense why I originally chose to use the `afterNotify` hook from the `notify` method.
@shalvah I'm not sure if it's a good idea to add the hook and then remove after when done. Consider this:
- `await Honeybadger.notifyAsync('message 1')`
- `await Honeybadger.notifyAsync('message 2')`
- Two `afterNotify` hooks are going to be registered on global `afterNotify` array and both will be called from the one that receives the response first, right? That would produce unexpected results.

It seems that the solution is:
- revert to the original `notify('message', { afterNotify })` implementation, but we have to make sure that this is handler is called last in the from the list of handlers.
@joshuap Is there a reason why we chose to run the handlers that come from the `notify()` method first, and not last?
```
  if (notice && notice.afterNotify) {
    // why is this handler not pushed at the end of the array?
    handlers.unshift(notice.afterNotify)
  }

  for (let i = 0, len = handlers.length; i < len; i++) {
    handlers[i](error, notice)
  }
  return true
}
```

subzero10: Actually, I just wrote a test and it seems that we don't have to put the `notice.afterNotify` handler at the end of the array:
The current execution context will go all the way through the `afterNotifyHandler`s array, even if the promise is resolved with the first `afterNotify` handler. Resolving the promise simply puts the `then()` handler in a queue that will be invoked after the event loop takes back control. This will happen only after the current execution context is complete, which is when it goes through the array. Similar question [here](https://stackoverflow.com/questions/28896280/why-does-javascript-es6-promises-continue-execution-after-a-resolve).

With that said, I will go ahead and merge this PR (it's not in `master` branch, so we are pretty safe).

subzero10: I will let as is for now and will reconsider changing if I come across this again.

PR #685, opened 2022-01-17
#678 Callback to Promises

### LINKED ISSUE #678: Codebase: Callback to Promise based
### Discussed in https://github.com/honeybadger-io/honeybadger-js/discussions/674

Replace callbacks with promises in the package.
Consider using a [polyfill](https://github.com/rollup/plugins/tree/master/packages/inject) to support Promises for browsers that do not support it.

subzero10:
## Status
**READY**

## Description
Discussed in #674 
Fixes #678 
Replace callbacks with promises in codebase.

## Todos
- [x] Tests
- [x] ~~[Polyfill](https://github.com/rollup/plugins/tree/master/packages/inject) for browser~~
- [ ] ~~Changelog Entry (unreleased)~~ (is a Changelog entry needed?)


subzero10:
I ran the `npm run lint -- --fix` in this branch so it may be better if you hide whitespaces while reviewing.

joshuap (commented):
This is looking good. Is supporting IE 11 the only reason we'd need to include the polyfill? If so, I'm in favor of dropping support for IE 11 in latest versions now that's it's deprecated. We could also recommend that users include their own Promise polyfill if they want to support it.

Relevant: https://caniuse.com/promises

Edit: I guess another question I have is how much will including the polyfill in the browser bundle increase the file size?

subzero10:
> Is supporting IE 11 the only reason we'd need to include the polyfill? If so, I'm in favor of dropping support for IE 11 in latest versions now that's it's deprecated. We could also recommend that users include their own Promise polyfill if they want to support it.

Yes, correct. I also noticed that in our integration tests we already have a promises polyfill:
```
// test/integration/karma.conf.js
...
files: [
      // Polyfills for IE 11
      'test/integration/node_modules/promise-polyfill/dist/polyfill.js',
      'test/integration/node_modules/whatwg-fetch/dist/fetch.umd.js',
...
```

subzero10:
@joshuap The integration tests for IE11 are failing, even though a promise polyfill is available.
I confirmed that the issue is promises:
- The tests on IE11 pass if I comment out the block of code that uses promises:
```
/* client.ts | line 200
getSourceForBacktrace(sourceCodeData, this.__getSourceFileHandler)
      .then(sourcePerTrace => {
        sourcePerTrace.forEach((source, index) => {
          notice.backtrace[index].source = source
        })

        this.__send(notice)
      })
*/

// it works if only let this:
this.__send(notice)
```

- To make sure it's a promises issue, I tried replacing the above block of code with a simple promise resolution:
```
const prom = () => new Promise<void>(resolve => resolve())
prom()
  .then(() => {
    this.__send(notice)
  })
```
**No luck**.

- In any case, I have confirmed that the promise polyfill is indeed available:
```
// sandbox.html
var prom = () => new Promise(resolve => resolve());
prom().then(() => results.promiseResolved = true)

// test.js
expect(results.promiseResolved).toEqual(true);
```

- Finally, I added 50ms delay to the setTimeout of the sandbox, just in case the assertions are executed before the promises are resolved. **No luck**.

Conclusion:
I don't know why IE11 fails, given that promises are available. The failure might be due to the way the integration tests are setup. I decided that it's not worth it to spend time resetting up the integration tests.

Suggestions:
- If we are indeed going to stop supporting IE11, we can remove IE11 from the integration tests.
- We can close this PR for now and let the code with callbacks.

joshuap (dismissed):
LGTM, glad you got it working!

subzero10:
Just merged with master. Will have to review again.

PR #690, opened 2022-01-26
Apply enableUncaught setting to timers and event listeners

### LINKED ISSUE #635: Should the `enableUncaught` config option apply to all instrumentation?
Currently, the `enableUncaught` option disables reporting errors caught by `window.onerror`,  but there are a few other places we automatically report errors:

1. [timers plugin](https://github.com/honeybadger-io/honeybadger-js/blob/master/src/browser/integrations/timers.ts)
2. [event_listeners plugin](https://github.com/honeybadger-io/honeybadger-js/blob/master/src/browser/integrations/event_listeners.ts)

Both of those plugins use the [`__wrap` function](https://github.com/honeybadger-io/honeybadger-js/blob/4e00f2b158182929c7aac745dedbbcccaa73fb60/src/browser.ts#L134) to automatically report errors in callbacks. Currently when `window.onerror` is disabled, `try/catch` is used instead (warning: the logic in `__wrap` is fairly complex).

Recently a customer reported an issue where they are using honeybadger.js in a third-party chat widget that will be installed in many different host applications, so they disabled `enableUncaught` so that they could report errors in their widget by intentionally calling `Honeybadger.notify`. They were confused when an unrelated error was reported in an event listener.

Does it make sense to expand the meaning of `enableUncaught` to apply to *all* automatically reported errors?

joshuap:
This should be enough, since both the timer and event listener plugins use `__wrap()`. This will still add the breadcrumbs, but skip reporting the errors if `enableUncaught === false`.

Fixes #635

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -137,14 +137,13 @@ class Honeybadger extends Client {
         // eslint-disable-next-line @typescript-eslint/no-this-alias
         const client = this
         func.___hb = <WrappedFunc>function () {
-          const onErrorEnabled = client.config.enableUncaught
           // Catch if:
           //   1. We explicitly want to catch (i.e. if the error could be
           //      caught before reaching window.onerror)
           //   2. The browser provides less information via the window.onerror
           //      handler
           //   3. The window.onerror handler is unavailable
-          if (opts.catch || preferCatch || !onErrorEnabled) {
+          if (opts.catch || preferCatch || !client.config.enableUncaught) {
             try {
               // eslint-disable-next-line prefer-rest-params
               return func.apply(this, arguments)
@@ -163,7 +162,9 @@ class Honeybadger extends Client {
                   }
                 }
               )
-              client.notify(err)
+              if (client.config.enableUncaught) {
+                client.notify(err)
+              }
               throw (err)
             }
           } else {
--- a/src/browser/integrations/onerror.ts
+++ b/src/browser/integrations/onerror.ts
@@ -28,8 +28,6 @@ export function onError(_window: any = window): Plugin {
             return
           }
 
-          if (!client.config.enableUncaught) { return }
-
           if (line === 0 && /Script error\.?/.test(msg)) {
             // See https://developer.mozilla.org/en/docs/Web/API/GlobalEventHandlers/onerror#Notes
             client.logger.info('Ignoring cross-domain script error: enable CORS to track these types of errors', arguments)
@@ -56,7 +54,9 @@ export function onError(_window: any = window): Plugin {
             }
           )
 
-          client.notify(notice)
+          if (client.config.enableUncaught) {
+            client.notify(notice)
+          }
         }
 
         return function (msg, url, line, col, err) {


joshuap:
No rush @subzero10—requesting review for when you feel better!

src/browser.ts
```diff
@@ -137,14 +137,13 @@ class Honeybadger extends Client {
...
         func.___hb = <WrappedFunc>function () {
-          const onErrorEnabled = client.config.enableUncaught
           // Catch if:
           //   1. We explicitly want to catch (i.e. if the error could be
           //      caught before reaching window.onerror)
           //   2. The browser provides less information via the window.onerror
           //      handler
           //   3. The window.onerror handler is unavailable
-          if (opts.catch || preferCatch || !onErrorEnabled) {
+          if (opts.catch || preferCatch || !client.config.enableUncaught) {
```
subzero10, line 146: This is confusing, but I think I understand it.
Basically, add the breadcrumbs if `opts.catch` or `preferCatch` but only call notify if `enableUncaught` is set.
What is the point of `opts.catch` if it cannot take precedence over the global config though (`enableUncaught`) ?

joshuap: That's a good question. The answer is that in the past we supported a `Honeybadger.wrap` method that shared this underlying implementation, and it used `opts.catch` to signal that it wants to explicitly report the error. I think I can actually remove `opts.catch` since `Honeybadger.wrap` was removed (it was a pain to support).

src/browser.ts
```diff
@@ -137,14 +137,7 @@ class Honeybadger extends Client {
...
         func.___hb = <WrappedFunc>function () {
-          const onErrorEnabled = client.config.enableUncaught
-          // Catch if:
-          //   1. We explicitly want to catch (i.e. if the error could be
-          //      caught before reaching window.onerror)
-          //   2. The browser provides less information via the window.onerror
-          //      handler
-          //   3. The window.onerror handler is unavailable
-          if (opts.catch || preferCatch || !onErrorEnabled) {
+          if (preferCatch) {
```
joshuap, line 140: @subzero10 take a look at this. I think the `opts.catch` and `enableUncaught` checks were both unnecessary because the `window.onerror` is always available (even when `enableUncaught === false`). What we really want is just to choose which method to use to add the breadcrumb (should happens always) and report the error (should only happen when `enableUncaught === true`).

src/browser/integrations/onerror.ts
```diff
@@ -28,8 +28,6 @@ export function onError(_window: any = window): Plugin {
             return
           }
 
-          if (!client.config.enableUncaught) { return }
-
           if (line === 0 && /Script error\.?/.test(msg)) {
             // See https://developer.mozilla.org/en/docs/Web/API/GlobalEventHandlers/onerror#Notes
             client.logger.info('Ignoring cross-domain script error: enable CORS to track these types of errors', arguments)
```
subzero10, line 33: I think we shouldn't log this message if `enableUncaught` is `false`, or we could change it to a _debug_ log.

src/browser.ts
subzero10: The way I understand this is:
- sometimes `window.onerror` has more context
- with this block of code we were deciding whether we want to let a potential error be caught with the `window.onerror` listener or catch it here using try/catch

Based on the above:
- `opts.catch` was here to allow the user to take this decision, though I feel it may be unnecessary
- the `enableUncaught` looks redundant to me. The importance of this check is when calling `client.notify` which you added below (line 159)

In conclusion, I think your latest change makes the code more readable and covers the purpose of this issue/PR.

joshuap: Thanks @subzero10, I think you got it. I'm pretty sure the `enableUncaught` check also made more sense in the past because we didn't install the `window.onerror` handler when it was `false` (and it used to be called `onerror`), so in that case there would be no onerror handler and we'd have to always use `catch`.

src/browser/integrations/onerror.ts
joshuap: Good idea, I'm going to add an `enableUncaught` check here. I think users should be aware of these errors since they can improve their error monitoring by adding the CORS config.

PR #700, opened 2022-02-13
Name wrapped Lambda handlers for better stacktraces

shalvah:
Switched from anonymous functions to named functions so that any stacktraces show `...at <wrappedLambdaHandler(..)` rather than `...at <unknown>` or `...at <anonymous>`.

### DIFF
--- a/src/server/aws_lambda.ts
+++ b/src/server/aws_lambda.ts
@@ -19,7 +19,7 @@ function isHandlerSync(handler: Handler): handler is SyncHandler {
 }
 
 function asyncHandler<TEvent = any, TResult = any>(handler: AsyncHandler<TEvent, TResult>, hb: Honeybadger): AsyncHandler<TEvent, TResult> {
-    return async (event, context) => {
+    return async function wrappedLambdaHandler(event, context) {
         try {
             return await handler(event, context)
         }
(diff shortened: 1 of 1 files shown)


subzero10 (approved):

subzero10:
@joshuap Should we consider hiding these from the stack traces entirely?

cc @joshuap

joshuap:
> @shalvah Should we consider hiding these from the stack traces entirely?
> 
> cc @joshuap

Is what you're suggesting similar to how we hide the first frame(s) of the stack when we [generate a stack trace in `Honeybadger.notify()`](https://github.com/honeybadger-io/honeybadger-js/blob/341ccbbfacfa8764ea3a3f8e75822ebc18b5d291/src/core/client.ts#L284)? I think that could make sense; it's a tradeoff—hiding lines from the stack trace risks making the problem harder to debug, but in some cases it may be worth it (like if it causes regular confusion). It would help if I can see a full example of a stack trace.

shalvah:
I don't think we need to worry about it much; in a typical scenario:

```js
exports.doStuff = Honeybadger.lambdaHandler(function doStuff() {...});
```
the top line of the trace will show the user's function name (doStuff), since an error will likely be thrown in there. I think giving our wrapped function a name just helps with completeness and ease of debugging.

subzero10:
> > @shalvah Should we consider hiding these from the stack traces entirely?
> > cc @joshuap
> 
> Is what you're suggesting similar to how we hide the first frame(s) of the stack when we [generate a stack trace in `Honeybadger.notify()`](https://github.com/honeybadger-io/honeybadger-js/blob/341ccbbfacfa8764ea3a3f8e75822ebc18b5d291/src/core/client.ts#L284)? I think that could make sense; it's a tradeoff—hiding frames from the stack trace risks making the problem harder to debug, but in some cases it may be worth it (like if it causes regular confusion). It would help if I can see a full example of a stack trace.

Yeah, exactly. I did something similar for the .Net Honeybadger Client. But as you said, there are tradeoffs. I'm OK without it for now.

Issue #688, opened 2022-01-23
AWS Lambda - Isolated context per handler


subzero10:
Issue discussed [here](https://github.com/honeybadger-io/honeybadger-js/pull/680#issuecomment-1003383252).

Suggestion from @shalvah is to use `AsyncLocalStorage` to isolate `context` (and `breadcrumbs`) to the executing lambda.

subzero10:
This has been released with v4 (currently in beta).

PR #702, opened 2022-02-13
#688 AWS Lambda Handler - AsyncLocalStorage

subzero10:
## Status
**READY**

## Description
`HoneybadgerStore` interface with `getStore()` method:
- `DefaultStore`: store context and breadcrumbs in a store variable
- `AsyncStore`: an `AsyncLocalStorage` singleton to isolate context and breadcrumbs in lambda functions (only used in `Honeybadger.lambdaHandler`)

## Todos
- [x] Async Handler implementation + tests
- [x] Callback-based Handler implementation + tests
- [x] Use `DefaultStore` if `AsyncLocalStorage` is not available
- [x] More Test Cases:
    - 2 handlers running at the same time => only one throws => the notice should have context and breadcrumbs only from the failed one.
- [x] Test on real AWS lambda function (try to force re-use nodejs process between function invocations)
- [x] Move async_store tests to separate file


src/core/client.ts
```diff
@@ -37,19 +38,11 @@ const STRING_EMPTY = ''
 const NOT_BLANK = /\S/
 
 export default class Client {
-  /** @internal */
```
subzero10, line 40: I think this is redundant now that we use Typescript.

src/core/types.ts
```diff
@@ -94,3 +95,7 @@ export interface CGIData {
   HTTP_COOKIE: string | undefined,
   [x: string]: unknown
 }
+
+export interface HoneybadgerStore {
```
subzero10, line 99: This was created to match the usage of `AsyncLocalStorage`.

src/server/aws_lambda.ts
```diff
@@ -18,46 +19,48 @@ function isHandlerSync(handler: Handler): handler is SyncHandler {
...
-                    }
-                })
+    })
+}
+
+function asyncHandler<TEvent = any, TResult = any>(handler: AsyncHandler<TEvent, TResult>, hb: Honeybadger): AsyncHandler<TEvent, TResult> {
+    return (event, context) => {
+        hb.configure({store: AsyncStore})
+        return new Promise<TResult>((resolve, reject) => {
+            AsyncStore.run({context: {}, breadcrumbs: []}, () => {
```
subzero10, line 35: I know that this could be [awaited](https://nodejs.org/api/async_context.html#usage-with-asyncawait) but I didn't find a way to make it work with Typescript (return type is `void`).

test/unit/core/client.test.ts
```diff
@@ -22,6 +22,7 @@ describe('client', function () {
       logger: nullLogger(),
       environment: null
     })
+    client.configure()
```
subzero10, line 25: We have to make sure to call `configure` any time we create a new instance of `Honeybadger`, otherwise `__store` will not be initialized. I think this is OK since `configure` is necessary to set the apiKey. If a new instance is created with `Honeybadger.factory()`, then it will be OK.

test/unit/server/aws_lambda.server.test.ts
```diff
@@ -286,5 +382,137 @@ describe('Lambda Handler', function () {
                 })
             })
         })
+
+        it('isolates context only to a single lambdaHandler', function () {
```
subzero10, line 386: Tests for callback-based handler.

test/unit/server/aws_lambda.server.test.ts
```diff
@@ -165,9 +163,107 @@ describe('Lambda Handler', function () {
                 }, 50)
             })
         })
+
+        it('isolates context only to a single lambdaHandler', async () => {
```
subzero10, line 167: Tests for async handler.

subzero10:
@shalvah Please take a look at this implementation of `AsyncLocalStorage` and let me know what you think. Can you think of any other tests we can write?

test/unit/core/client.test.ts
joshuap: Couldn't we just set `this.__store = opts.store || new DefaultStore()` when initializing the client, but let `configure` override it?

joshuap (commented):
This is looking good to me. I was not familiar with `AsyncLocalStorage`, but I just read an [article](https://www.freecodecamp.org/news/async-local-storage-nodejs/) about it—cool! It sounds like it could be useful elsewhere, potentially. If I'm understanding it correctly, it helps us gain similar benefits to thread-local storage inside of Node's single-threaded event loop?

src/server/aws_lambda.ts
shalvah: Should this not be `store.run()`? Looks like `AsyncStore` is the type, not the object.

shalvah: Ah, my bad; just realised it is the object. Perhaps we should rename the exported constant to `asyncStore`, so it's clear it's an object, not a type or class.

shalvah:
@joshuap Yeah, ALS is like TLS for Node's single thread. Also helps with propagating contexts across sync-async boundaries, like callbacks, without having to pass the context manually all the way down. When you run `fs.read(file, cb)`, and the Node.js event loop switches to do something else while waiting (like serving another request), ALS makes sure that when it switches back to execute `cb()`, it can recover the specific context as at then, without you having to capture it all in the closure.

@subzero10 how do we deal with older versions of Node.js? I think ALS is v12 upwards.🤔 Polyfill, drop support, use the regular store?

Regarding tests: I think we need real tests on AWS Lambda, with some way to force container reuse, with and without the ALS versions. I think ALS is very useful for local apps like webservers which may toggle between multiple requests on the same Node.js thread (we should look into this, btw!), but I wonder if Lambda's execution environment is susceptible to this. What memory is shared between invocations? If I run `hb.context({a: "2"})` in a handler and don't clear, is this context still there on reuse. I suspect it might, since the Honeybadger object was likely initialized outside the handler, but if we can verify that that happens (and that this fixes it), we should be good.

test/unit/core/client.test.ts
subzero10: Yeah, that's what I had at first, but it seems that `configure()` is a required method call for Honeybadger to work (need to at least set the `apiKey`). It's just that we don't always call `configure()` in our tests. I can add it if you think it may raise some issues.

subzero10:
@shalvah Good point on backwards compatibility. I will make sure to test this. I think I will just use the default store, since AWS supports up to v14.

Yeah, most examples on ALS I see online are about separating context between requests.

>If I run hb.context({a: "2"}) in a handler and don't clear, is this context still there on reuse. I suspect it might, since the Honeybadger object was likely initialized outside the handler, but if we can verify that that happens (and that this fixes it), we should be good.

`hb.context()` will set `{a: "2"}` in `hb.__store`. `hb.__store` is a reference to the ALS instance, which we always create a new _context_ when we call `AsyncStore.run`. If you look at the test, that's what I'm testing. I'm calling the two handlers immediately one after the other and I also delay throwing the error. Finally, I assert that the correct data is sent with each notice.
In any case, I agree with you, we should look for a way to do real tests on AWS and I will do that.

test/unit/server/aws_lambda.server.test.ts
```diff
@@ -76,8 +76,8 @@ describe('Lambda Handler', function () {
                 apiKey: 'testing'
             })
 
-            const handler = client.lambdaHandler(async function(_event, _context) {
-                return mockAwsResult({body:'works!'})
+            const handler = client.lambdaHandler(async function (_event, _context) {
+                return mockAwsResult({body: 'works!'})
```
shalvah, line 80: I know this isn't part of this change, but we don't need to always wrap in `mockAwsResult()`. Lambda functions can return anything, and aren't only triggered from API Gateway.

shalvah:
> If you look at the test, that's what I'm testing. I'm calling the two handlers immediately one after the other and I also delay throwing the error. Finally, I assert that the correct data is sent with each notice.

Ah, nice. The diff was collapsed, so I missed it earlier. We could also have a test where the first handler sets context but does not throw, and but the second adds context and throws, and we assert that only the second's context is used.

> For starters, I could run the same test with the default store and it should fail.

That would be nice, too😄. I think you've answered my questions.

test/unit/server/aws_lambda.server.test.ts
subzero10: You are correct. I think I used this to be consistent in my tests.

subzero10:
> > If you look at the test, that's what I'm testing. I'm calling the two handlers immediately one after the other and I also delay throwing the error. Finally, I assert that the correct data is sent with each notice.
> 
> Ah, nice. The diff was collapsed, so I missed it earlier. We could also have a test where the first handler sets context but does not throw, and but the second adds context and throws, and we assert that only the second's context is used.
> 
> > For starters, I could run the same test with the default store and it should fail.
> 
> That would be nice, too😄. I think you've answered my questions.

Thanks! I have consolidated all your suggestions as checkboxes in the PR description and will start working on them asap.

subzero10:
I took a deeper look at my tests and realized a few things that need improvement:
- In the tests for the async handlers, I call `await` which means I'm not simulating simultaneous execution. I have pushed an improved version of the tests.
- I ran the tests with a _normal variables store_ (`DefaultStore`) and they randomly fail, which validates the correct implementation of `AsyncLocalStorage`. However, I couldn't write deterministic failing tests so I didn't keep them.
- I also discovered a [bug](https://github.com/honeybadger-io/honeybadger-js/pull/702/files#diff-16e04671c5715d58dec62f934b3f9b863853f8c01f73db96fe02eaae82a0540aR98), which I believe I originally introduced 🤦

src/core/util.ts
```diff
@@ -95,7 +95,7 @@ export function runBeforeNotifyHandlers(notice: Notice | null, handlers: BeforeN
 
 export function runAfterNotifyHandlers(notice: Notice | null, handlers: AfterNotifyHandler[], error?: Error): boolean {
   if (notice && notice.afterNotify) {
-    handlers.unshift(notice.afterNotify)
+    notice.afterNotify(error, notice)
```
subzero10, line 98: @joshuap This is a **bug fix** for a bug I introduced. When calling `runAfterNotifyHandlers` we are sending a reference to `this.__afterNotifyHandlers`, and at this line we are adding the notice's `afterNotify` handler. If a notice is to be sent before `hb.clear()` is called, this notice's `afterNotify` handler is still in `this.__afterNotifyHandler`. As a result, that handler is called twice.

I think we should release a patch for v3, but I'm confused where to target the PR. Changelog has `v3.2.8` as latest version, but I can't find a branch for it.

How I introduced this bug? I did it with [this commit](https://github.com/honeybadger-io/honeybadger-js/commit/71b38f23bb86df1a346a2aef7cae7ead70cc39d1#diff-e84e938a5120632cd9bfda48f378416139afdb46dbb35444370bae86a6f1c248L98) a couple of months ago as part of an effort to cleanup the code.

src/server/aws_lambda.ts
subzero10: Poking @joshuap on this also.

src/server/async_store.ts
```diff
@@ -0,0 +1,15 @@
+import { BreadcrumbRecord, HoneybadgerStore } from "../core/types"
+
+let Store: HoneybadgerStore<{ context: Record<string, unknown>; breadcrumbs: BreadcrumbRecord[]; }>
+try {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const { AsyncLocalStorage } = require('async_hooks')
```
subzero10, line 6: We want to fallback to the `DefaultStore` if `AsyncLocalStorage` is not available.
I'm not 100% with my solution, I know this the CommonJS approach (using `require`), but I didn't find the equivalent with ESM (using `import`). Let me know if you have a better solution.

I tested this by enabling tests on Node v10 and the [tests started failing](https://github.com/honeybadger-io/honeybadger-js/runs/5260086192?check_suite_focus=true): `AsyncStore` was an instance `DefaultStore` and as expected, the assertions failed because the lambda handlers did not have isolated context.

shalvah:
> which means I'm not simulating simultaneous execution

To be fair, I don't think the same container would have handlers executing concurrently. I think that's more of a problem for a traditional server juggling multiple requests on the same thread, so maybe instead we move this test to that area?😄 For lambda handlers, the problem is more of "leftover" global state—if X sets context and finishes executing without removing that context, and the same process is used for Y, does Y inherit that context?

Also, do you think we should split the Lambda tests into their own file, similar to how you split out the implementation?

src/server/async_store.ts
subzero10: Yeah true, but we are supposed to have `import` statements at the top of the script, outside of any blocks (try/catch). That's why I used the `require`.

subzero10:
> To be fair, I don't think the same container would have handlers executing concurrently. 

Hmm, I thought of that too, but I wasn't able to rule it out. 

> I think that's more of a problem for a traditional server juggling multiple requests on the same thread, so maybe instead we move this test to that area?😄 

Which area are you referring to?

> For lambda handlers, the problem is more of "leftover" global state—if X sets context and finishes executing without removing that context, and the same process is used for Y, does Y inherit that context?

Actually, this is already handled with this code (that is already on `master`), right?:
```
hb.notify(err, {
  afterNotify: function () {
    hb.clear() // this line clears the global state
    reject(err)
  }
})
```

> Also, do you think we should split the Lambda tests into their own file, similar to how you split out the implementation?

Good idea, will do!

shalvah:
> Actually, this is already handled with this code (that is already on master), right?:

Yes, but only if an error is reported. The important test is what happens if context is set in one handler but nothing is reported, then later on something is reported in another handler.

shalvah:
> Which area are you referring to?

I meant that, after this, we should look at our traditional server context implementation. It probably suffers from this issue, and we need a test that makes concurrent requests to verify.

subzero10:
> > Actually, this is already handled with this code (that is already on master), right?:
> 
> Yes, but only if an error is reported. The important test is what happens if context is set in one handler but nothing is reported, then later on something is reported in another handler.

True, I forgot that. Yes, ALS + the tests cover this case.

shalvah:
Exactly. That's the kind of scenario ALS was introduced to solve.

subzero10:
I think I was able to test this on a real lambda function using the _aws-lambda example project_ (found in `honeybadger-js/examples`).

The next 3 screenshots show same process id (`pid`), i.e. same nodejs process:
1. Lambda handler that throws (which also calls `Honeybadger.clear()`)
```
const randomKey = randomString()
const randomValue = randomString()
Honeybadger.setContext({
  [process.pid]: 'pid as key',
  [randomKey]: randomValue,
  processId: process.pid
})
```
Screenshot 2022-02-22 at 10 21 17 PM

2. Lambda handler that does not throw, but I called `Honeybadger.notifyAsync` to see if the same nodejs process is used. `notifyAsync` does not call `Honeybadger.clear()`, therefore context is not cleared.
```
Honeybadger.setContext({
  [process.pid]: 'pid as key',
  processId: process.pid,
  hello: 'value set from hello handler',
})

await Honeybadger.notifyAsync('hello handler')
```
Screenshot 2022-02-22 at 10 21 41 PM

3. Lambda handler that throws (same as before)
Screenshot 2022-02-22 at 10 21 30 PM

As you can see, the same process is used but the context is different, regardless whether `Honeybadger.clear()` was called or not.

subzero10:
@joshuap @shalvah This is ready. Can you please do another round of review? Thanks!

shalvah (approved):

src/core/util.ts
joshuap: @subzero10 here's the branch! Can you submit a PR there after this is merged? https://github.com/honeybadger-io/honeybadger-js/tree/v3-stable

src/server/aws_lambda.ts
joshuap: @subzero10 I think I've been capitalizing types, classes, interfaces, etc. ([example](https://github.com/honeybadger-io/honeybadger-js/blob/master/src/browser.ts#L1)), and lower casing functions ([example](https://github.com/honeybadger-io/honeybadger-js/blob/d792a67a465089f31d4f62f4b83190dacfe56580/src/browser.ts#L4)). I don't have a strong opinion on the convention as long as we're consistent.

joshuap (approved):

src/server/aws_lambda.ts
subzero10: OK, let me clarify this a little more:
- Class name: PascalCase
- function name: camelCase
- Singleton Instances of Classes: This is the case where a class is instantiated once and used throughout the app, like the Honeybadger instance [here](https://github.com/honeybadger-io/honeybadger-js/blob/master/examples/aws-lambda/honeybadger.js#L1) and [here](https://github.com/honeybadger-io/honeybadger-js/blob/master/src/server.ts#L108). I followed this pattern and did the same for `AsyncStore`.

subzero10: OK, thanks for clarifying. For `AsyncLocalStorage`, I went with a singleton because that's how all the examples I went through do it. Plus, I also tried creating a new ALS instance on for each lambda handler invocation and the tests were failing. That was at the beginning of the implementation though and I didn't dive deep into it.

subzero10: **Update**
I just tried without a singleton:
```
const store = new AsyncStore()
hb.configure({store})
return new Promise<TResult>((resolve, reject) => {
    store.run({context: {}, breadcrumbs: []}, () => {
        ...
    })
})
```
This doesn't work because `hb.configure({ store })` every time sets the `store` in `Honeybadger` to a new ALS object. When running the lambda handlers at the same time, only the handler that set the `store` last will have a context. The other handler(s) returns undefined and fails the test.
Note: I know that lambda handlers running simultaneously is unlikely but I think its better (and future-proof) to cover this scenario.

When `AsyncLocalStorage` is a singleton, `hb.configure({ store })` always sets to the same object and all handlers retrieve the correct context (because they are inside the callback of the `run` method).

So, based on our implementation, I think there is good reason to keep a singleton for ALS.

src/core/util.ts
subzero10: @joshuap False alarm, this bug was introduced with #659 , which is not on `v3`. We should release a new version on `v4` since this PR is now merged and it will be OK.

PR #703, opened 2022-02-15
Catch unknown errors in data sanitizer

joshuap:
This is for a customer issue where we can't reproduce the specific error, but I'm pretty sure something in the data passed to breadcrumbs (via `console.log`) is blowing up the sanitize function.

I created a v3-stable base branch so that I can release immediately. I'll also add it to the main (v4) branch after this is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased][latest]
+### Fixed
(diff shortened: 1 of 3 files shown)


subzero10 (approved):
Looks good to me.

subzero10:
> This is for a customer issue where we can't reproduce the specific error, but I'm pretty sure something in the data passed to breadcrumbs (via `console.log`) is blowing up the sanitize function.
> 
> I created a v3-stable base branch so that I can release immediately. I'll also add it to the main (v4) branch after this is merged.

Even if not the case, I think the _try/catch_ approach on the `sanitize` method is a good move.

joshuap:
Released in 3.2.8

PR #704, opened 2022-02-16
Allow special characters in tags

joshuap:
Also adds support for space-delimited tags; these strings are
equivalent:

- "one, two, three"
- "one two three"
- "one, two three"

This is a change I recently made to honeybadger-ruby (https://github.com/honeybadger-io/honeybadger-ruby/pull/422) and I realized we did the same thing here. I don't know why we were being so strict with the tag sanitizer, but I think this is better.

eslint also made some formatting changes, sorry about the noise.

### DIFF
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -25,10 +25,10 @@ const notifier = {
 }
 
 // Split at commas
-const TAG_SEPARATOR = /,/
+const TAG_SEPARATOR = /,|\s/
 
 // Removes any non-word characters
-const TAG_SANITIZER = /[^\w]/g
+const TAG_SANITIZER = /\s/g
 
 // Checks for blank strings
 const STRING_EMPTY = ''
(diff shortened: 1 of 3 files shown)


src/core/client.ts
```diff
@@ -24,11 +24,11 @@ const notifier = {
...
 
-// Split at commas
-const TAG_SEPARATOR = /,/
+// Split at commas and spaces
+const TAG_SEPARATOR = /,|\s/
 
-// Removes any non-word characters
-const TAG_SANITIZER = /[^\w]/g
+// Removes extra whitespace from individual tags
+const TAG_SANITIZER = /\s/g
```
subzero10, line 31: I don't think this is necessary anymore. If I'm thinking this correctly, by running the `TAG_SEPARATOR` first, there wouldn't be any individual tags with whitespace.

subzero10: Update: To confirm my theory, I just tested removing usages of `TAG_SANITIZER` on this branch and all the tests pass.

subzero10 (changes requested):

joshuap:
Updated

subzero10 (approved):

PR #711, opened 2022-03-05
Implement context isolation for Node.js

### LINKED ISSUE #708: Node.js server integration - isolated context per request
Like #688, but for traditional single-threaded, server-side Node.js apps. The goal is to ensure that context does not "spill over" across _concurrent_ requests being handled by the same process.

## Problem
First, a simple proof of concept (tested with the included express example):

```js
let i = 0;

app.get("/", (req, res) => {
  let reqId = ++i;
  console.log(`Starting request number ${reqId}, current context: ${JSON.stringify(Honeybadger.__context)}`);
  Honeybadger.setContext({ reqId });
  setTimeout(() => {
    console.log(`Ending request number ${reqId}, current context: ${JSON.stringify(Honeybadger.__context)}`);
    res.send("Done")
  }, 1000);
});
```

Then make two concurrent requests (I use [autocannon](https://www.npmjs.com/package/autocannon)):

```bash
autocannon --connections 2 --amount 2 http://localhost:3000
```

Terminal logs:

```
Example app listening at http://localhost:3000
Starting request number 1, current context: {"user_id":"1"}
Starting request number 2, current context: {"user_id":"1","reqId":1}
Ending request number 1, current context: {"user_id":"1","reqId":2}
Ending request number 2, current context: {"user_id":"1","reqId":2}
```

You can see the problem—the context is shared globally. 

## Fix
[AsyncLocalStorage](https://nodejs.org/api/async_context.html#asynchronous-context-tracking) 

Thankfully, @subzero10  has already implemented the stores in his recent PR to start this off.

To Do:
- Add a test that shows this
- Verify Node.js versions supported for this. If we're still tied to older versions, we may be able to use a polyfill like ALS' predecessor, [continuation-local storage](https://www.npmjs.com/package/continuation-local-storage) (although it doesn't seem to still be maintained), or just leave it as-is.

cc  @subzero10  @joshuap

shalvah:
## Status
READY

## Description
This PR fixes #708 by storing the context in [`AsyncLocalStorage`](https://nodejs.org/api/async_context.html) for web requests.

## Implementation
Thanks to the earlier work by @subzero10, the core implementation was relatively straightforward:
- First, when the agent is created, it uses the `GlobalStore` (renamed from `DefaultStore` so its behaviour is more obvious). This allows us to support simple single-request use cases like CLI jobs.
- We're exposing the `hb.run(cb)` method, which sets the store to the `AsyncStore` and runs the given code with it. Each request will have its own context, thanks to ALS. Users can wrap their request handlers in `hb.run()` to properly capture the context for each request. For supported frameworks like Express, they can just add the `Honeybadger.requestHandler` middleware, which is a wrapper over `hb.run`.

However, because of Node.js' oddities, I had to do several more investigations. Here's my tale:

The existing `Honeybadger.requestHandler()` simply wrapped the code in a [_domain_](https://nodejs.org/api/domain.html) and set the `next` handler as the domain's error handler. At first, it seemed strange, but I realized that the advantage of domains is that they can capture asynchronous errors triggered from the same context. In the example below, the error will be caught by `someErrorHandler`.:

```js
functionThatErrors = () => { throw new Error("Oh no") };

handler = () => {
  try {
    setTimeout(functionThatErrors, 100);
  } catch (e) {
    // This won't catch the error!
  }
}

dom = require("domain").create();
dom.on("error", someErrorHandler); // <- This will
dom.run(handler);
```

Without wrapping in a domain, there is **no way** to catch the error (unless you fix `functionThatErrors`). The error would bubble up to `process.on("uncaughtException")` (or crashes if you don't have one registered)`:

```js
try {
  handler ();
} catch (e) {
  // Doesn't work
}

process.on("uncaughtException", (err) => {
  // Catches it
});
```

Your `uncaughtException` handler will catch this, yes, but you've lost the context. In an Express app, for example, you'd no longer have access to the `res` object, so you can't send a response to the user. But with domains, you can create a new one for every request and give it an error handler that captures `res` and anything else you need. So, domains are sorta mini-worlds.

```js
app.get("/", (req, res, next) => {
  dom = domain.create()
  dom.on("error", () => res.send("Oh no"))
  dom.run(() => doStuff())
})
```

As you can see, this is often desirable for a webapp. You don't want your server to crash without sending a friendly error response (and then [you should let it crash](https://nodejs.org/api/domain.html#warning-dont-ignore-errors)).

For this reason, the `hb.run` function runs the user's code in a domain. However, there are problems with domains, The docs have [an entire article](https://nodejs.org/en/docs/guides/domain-postmortem/) on it. Additionally:
- domains are an easy way to swallow an exception and cause a memory leak due to unclaimed resources.
- nested domains can lead to unexpected behaviour. We don't know if the end user or the framework is also wrapping the request handling in a domain.

The biggest mindfuck with domains, though, is that the domain error handler will NOT be called [if there's a try/catch *anywhere* in the stack](https://gist.github.com/owenallenaz/7141699), even if it's way outside the domain.

```js
dom.on("error", errorHandler);
dom.run(() => {
  try {
    throw new Error("ga");
  } catch (e) {
    // The error is caught here, as expected.
    // `errorHandler` is not called.
  }
});

try {
  dom.on("error", errorHandler);
  dom.run(() => {
    throw new Error("ga");
  });
} catch (e) {
    // The error is caught here. WUT?
    // `errorHandler` is still ignored.
}
```
I suspect similar happens with Promises, because the docs say [domains don't catch Promise rejections](https://nodejs.org/api/domain.html#:~:text=Domains%20will%20not%20interfere%20with%20the%20error%20handling%20mechanisms%20for%20promises.%20In%20other%20words%2C%20no%20%27error%27%20event%20will%20be%20emitted%20for%20unhandled%20Promise%20rejections.), but they did in my tests.

There's also the fact that you can't resume execution. Once the domain error handler is triggered, nothing else executes after that, so the error handler has to call the next function to be executed. These two behaviours limited the tests I could write.

Unforunately, there's no way to catch every possible scenario. Error handling in Node.js is a mess, and we want to limit our changes to the user's execution environment.

To mitigate these, `hb.run` takes an `errorHandler` argument, and will only use a domain if this value is supplied. This way, frameworks which are more promise-friendly can run with minimal interference, and users can opt out of domains.

----

Will send in a docs PR when this is merged.

### DIFF
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -77,6 +77,9 @@ export default class Client {
       ...opts,
     }
 
+    // First, we go with the default (shared) store.
+    // Webserver middleware can then switch to the AsyncStore for async context tracking.
+    this.__store = new DefaultStore({ context: {}, breadcrumbs: [] })
     this.logger = logger(this)
   }
 
@@ -97,9 +100,12 @@ export default class Client {
       this.config.__plugins.forEach((plugin) => plugin.load(this))
     }
 
-    this.__store = opts.store || new DefaultStore({ context: {}, breadcrumbs: [] })
-
     return this
+  }
+
+  /** @internal */
+  __setStore(store: HoneybadgerStore<{ context: Record<string, unknown>; breadcrumbs: BreadcrumbRecord[] }>) {
+    this.__store = store
   }
 
   beforeNotify(handler: BeforeNotifyHandler): Client {
@@ -185,7 +191,7 @@ export default class Client {
       }
     })
 
-    const breadcrumbs = this.__store.getStore().breadcrumbs
+    const breadcrumbs = this.__getStoreOrDefaultValues().breadcrumbs
     notice.__breadcrumbs = this.config.breadcrumbsEnabled ? breadcrumbs.slice() : []
 
     // we need to have the source file data before the beforeNotifyHandlers,
@@ -267,7 +273,7 @@ export default class Client {
       return null
     }
 
-    const context = this.__store.getStore().context
+    const context = this.__getStoreOrDefaultValues().context
     const noticeTags = this.__constructTags(notice.tags)
     const contextTags = this.__constructTags(context["tags"])
     const configTags = this.__constructTags(this.config.tags)
@@ -382,5 +388,17 @@ export default class Client {
     }
 
     return tags.toString().split(TAG_SEPARATOR).filter((tag) => NOT_BLANK.test(tag))
+  }
+
+  /**
+   * For ALS, the store may be uninitialized (if .run()` has not been called).
+   * This provides an easy way to read the existing stored values or fall back to defaults.
+   * Returns *a copy* of the store.
+   * @internal
+   */
+  protected __getStoreOrDefaultValues(): {context: Record<string, unknown>, breadcrumbs: BreadcrumbRecord[]} {
+    const existingStore = this.__store.getStore();
+    const store = existingStore || {};
+    return Object.assign({}, {context: {}, breadcrumbs: []}, store);
   }
 }
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -399,6 +399,10 @@ export default class Client {
   protected __getStoreOrDefaultObject(): {context: Record<string, unknown>, breadcrumbs: BreadcrumbRecord[]} {
     const existingStore = this.__store.getStore();
     const store = existingStore || {};
-    return Object.assign({}, {context: {}, breadcrumbs: []}, store);
+    return {
+      context: {},
+      breadcrumbs: [],
+      ...store
+    };
   }
 }
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -25,7 +25,7 @@ import {
   BacktraceFrame,
   BreadcrumbRecord
 } from './types'
-import { DefaultStore } from "./store";
+import { GlobalStore } from "./store";
 
 const notifier = {
   name: 'honeybadger-js',
@@ -77,9 +77,9 @@ export default class Client {
       ...opts,
     }
 
-    // First, we go with the default (shared) store.
+    // First, we go with the global (shared) store.
     // Webserver middleware can then switch to the AsyncStore for async context tracking.
-    this.__store = new DefaultStore({ context: {}, breadcrumbs: [] })
+    this.__store = new GlobalStore({ context: {}, breadcrumbs: [] })
     this.logger = logger(this)
   }
 
--- a/src/core/client.ts
+++ b/src/core/client.ts
@@ -191,7 +191,7 @@ export default class Client {
       }
     })
 
-    const breadcrumbs = this.__getStoreOrDefaultValues().breadcrumbs
+    const breadcrumbs = this.__getStoreOrDefaultObject().breadcrumbs
     notice.__breadcrumbs = this.config.breadcrumbsEnabled ? breadcrumbs.slice() : []
 
     // we need to have the source file data before the beforeNotifyHandlers,
@@ -273,7 +273,7 @@ export default class Client {
       return null
     }
 
-    const context = this.__getStoreOrDefaultValues().context
+    const context = this.__getStoreOrDefaultObject().context
     const noticeTags = this.__constructTags(notice.tags)
     const contextTags = this.__constructTags(context["tags"])
     const configTags = this.__constructTags(this.config.tags)
@@ -392,11 +392,11 @@ export default class Client {
 
   /**
    * For ALS, the store may be uninitialized (if .run()` has not been called).
-   * This provides an easy way to read the existing stored values or fall back to defaults.
+   * This provides an easy way to read the existing store object or fall back to a default.
    * Returns *a copy* of the store.
    * @internal
    */
-  protected __getStoreOrDefaultValues(): {context: Record<string, unknown>, breadcrumbs: BreadcrumbRecord[]} {
+  protected __getStoreOrDefaultObject(): {context: Record<string, unknown>, breadcrumbs: BreadcrumbRecord[]} {
     const existingStore = this.__store.getStore();
     const store = existingStore || {};
     return Object.assign({}, {context: {}, breadcrumbs: []}, store);
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -11,8 +11,9 @@ export class DefaultStore<T> implements HoneybadgerStore<T> {
         return this.store
     }
 
-    run(store: T, callback: (...args: never[]) => void, args: never): void {
-        this.store = store
-        callback(args)
+    // eslint-disable @typescript-eslint/no-explicit-any
+    run<R, TArgs extends any[]>(store: T, callback: (...args: TArgs) => R, ...args: TArgs): R {
+        this.store = store;
+        return callback(...args);
     }
 }
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -11,8 +11,7 @@ export class GlobalStore<T> implements HoneybadgerStore<T> {
         return this.store
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    run<R, TArgs extends any[]>(store: T, callback: (...args: TArgs) => R, ...args: TArgs): R {
+    run<R, TArgs extends never[]>(store: T, callback: (...args: TArgs) => R, ...args: TArgs): R {
         this.store = store;
         return callback(...args);
     }
--- a/src/core/store.ts
+++ b/src/core/store.ts
@@ -11,7 +11,7 @@ export class GlobalStore<T> implements HoneybadgerStore<T> {
         return this.store
     }
 
-    // eslint-disable @typescript-eslint/no-explicit-any
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     run<R, TArgs extends any[]>(store: T, callback: (...args: TArgs) => R, ...args: TArgs): R {
         this.store = store;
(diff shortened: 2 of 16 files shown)


shalvah:
If anyone would like to test this out, I've updated the Express example to easily test this. You can switch between the old and new versions of this package by commenting one out and leaving the other:

```js
// const Honeybadger = require('@honeybadger-io/js'); // Old version
const Honeybadger = require('../../dist/server/honeybadger.js'); // Current
```

Start the server with node index.js, then try making 3 simultaneous requests to `/test-context` in a different tab. With autocannon:

```
autocannon --connections 3 --amount 3 http://localhost:3000/test-context
```

Look at the console logs from the server, and the error that is logged.

When we're done, I can change the example back to something more user-friendly.

joshuap (commented):
This is looking good to me. Thanks for the writeup!

It looks like there's an issue with the integration tests:

```
TypeError: Unable to get property '__breadcrumbs' of undefined or null reference
```

@subzero10, can you review and approve if you agree with everything?

Edit: Oh, this will also need a changelog entry! 👍

test/unit/server/middleware.server.test.ts
```diff
@@ -94,24 +94,31 @@ describe('Express Middleware', function () {
...
+    it('does not leak context between requests', function() {
         const app = express()
 
         app.use(client.requestHandler)
 
-        app.get('/', function(_req, res) {
-            res.send('Hello World!')
-        })
+        app.get("/:reqId", (req, res) => {
+            const initialContext = client.__store.getStore().context;
```
subzero10, line 103: To get proper types, you can use the TestClient class which exposes a `getContext()` method. You can see an example usage in `client.test.ts`.

src/server/middleware.ts
```diff
@@ -20,10 +19,7 @@ function fullUrl(req: Request): string {
 }
 
 export function requestHandler(req: Request, res: Response, next: NextFunction): void {
-  this.clear()
-  const dom = domain.create()
-  dom.on('error', next)
-  dom.run(next)
+  this.run(next, next)
```
subzero10, line 22: Nice!

src/core/store.ts
```diff
@@ -11,8 +11,9 @@ export class DefaultStore<T> implements HoneybadgerStore<T> {
         return this.store
     }
 
-    run(store: T, callback: (...args: never[]) => void, args: never): void {
-        this.store = store
-        callback(args)
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    run<R, TArgs extends any[]>(store: T, callback: (...args: TArgs) => R, ...args: TArgs): R {
```
subzero10, line 15: Maybe you could do:
```suggestion
    run<R, TArgs extends never[]>(store: T, callback: (...args: TArgs) => R, ...args: TArgs): R {
```
so that the eslint warning could go away?

src/core/client.ts
```diff
@@ -97,11 +100,14 @@ export default class Client {
       this.config.__plugins.forEach((plugin) => plugin.load(this))
     }
 
-    this.__store = opts.store || new DefaultStore({ context: {}, breadcrumbs: [] })
```
subzero10, line 100: Is there any reason you moved this from here to the constructor? At first, I was conflicted as well, but then I realized that `Honeybadger.configure` is a required method call for it to work (you need to at least call it to set the `apiKey`). And since we have other initializations happening here, I decided to include the store initialization.

We can leave it as you want, it doesn't make much difference, I just want to know how you thought about this.

Finally, if we are keeping this version, we should remove the `store` property from the `Config` interface (`types.ts`).

src/server.ts
```diff
@@ -103,6 +105,27 @@ class Honeybadger extends Client {
...
+
+    if (onError) {
+      // ALS is fine for context-tracking, but `domain` allows us to catch errors
+      // thrown asynchronously (timers, event emitters)
+      // We can't use unhandledRejection/uncaughtException listeners; they're global and shared across all requests
+      // But the `onError` handler might be request-specific.
+      // Note that this doesn't still handle all cases. `domain` has its own problems:
+      // See https://github.com/honeybadger-io/honeybadger-js/pull/711
+      const dom = domain.create();
+      const onErrorWithContext = (err) => this.__store.run(storeObject, () => onError(err));
```
subzero10, line 122: Doesn't this create a new context? I think that every time we call `ALS.run()` it creates a new isolated context. So if something was added to the store (in the `handler`) before throwing the error event is raised, we will not see it here because the `storeObject` is the initial object we retrieved at the beginning (before calling the `handler`).
Do you agree? Can we write a test for this?
Maybe it would be better if we wrapped all of this inside `this.store.run()` ? Like:
```
const storeObject = this.__getStoreOrDefaultObject();
this.__setStore(AsyncStore)

return this.__store.run(storeObject, () => {
  if (onError) {
    // ALS is fine for context-tracking, but `domain` allows us to catch errors
    // thrown asynchronously (timers, event emitters)
    // We can't use unhandledRejection/uncaughtException listeners; they're global and shared across all requests
    // But the `onError` handler might be request-specific.
    // Note that this doesn't still handle all cases. `domain` has its own problems:
    // See https://github.com/honeybadger-io/honeybadger-js/pull/711
    const dom = domain.create();
    dom.on('error', onError);
    handler = dom.bind(handler);
  }

  return handler();
});
```

src/core/store.ts
shalvah: I initially had the args typed as `never[]` (in a different format). It worked until I had to reference the args (the return statement). Apparently, `never` is for when you don't need the args. But I'll try in this form.

src/server.ts
shalvah: Hehe, I went through the same thought patterns and iterations too.😅

Your suggestion doesn't work because it seems the domain error handler is called *outside* the context of the ALS context (even though we ran the domain inside it). A simple script you can test with:

```js
const als = new (require("async_hooks").AsyncLocalStorage);

als.run({a: true}, () => {
    const dom = require("domain").create();
    onError = () => console.log(als.getStore())
    dom.on('error', onError);
    handler = dom.bind(() => { throw new Error });
    handler();
});
```

Logs `undefined`.

So apparently, the ALS context ends before the domain error handler is called.

As for your first question, the context is an object, so it is preserved, since we're passing the same object to both cases of `als.run`. Test:

```js
const als = new (require("async_hooks").AsyncLocalStorage);

ctx = {a: true}
als.run(ctx, () => {
    const dom = require("domain").create();
    onError = () => console.log(als.getStore())
    dom.on('error', (e) => als.run(ctx, onError));
    handler = dom.bind(() => { 
        als.getStore().b = true;
        throw new Error
    });
    handler();
});
```

Logs:

```
{ a: true, b: true }
```

shalvah: Also, re tests: the `captures errors in timers with the right context` below tests for this. The context object is set in the handler, then we grab it in the error handler and assert that it's correct.

src/core/client.ts
shalvah: For one, the `.configure` method is typically for things the end user should set, and for now, the store is not meant to be set by them. It's too critical and tied to implementation details. And they don't need to set it either, as `hb.run` will set it as needed.

Also, the store is too important to not be set, IMO. It's better to construct the object straightaway with as many key details set as possible, avoiding any invalid states.

Finally, in our tests, we don't always call `.configure`, and having to add that to get things working was a pain.

src/core/store.ts
shalvah: Seems like it worked!👍

test/unit/server/middleware.server.test.ts
shalvah: Hm, the TestClient class seems to be a general client instance, while this test needs a server instance. Think we can leave this  as is; the types don't do much.

src/core/client.ts
subzero10: > For one, the .configure method is typically for things the end user should set, and for now, the store is not meant to be set by them. It's too critical and tied to implementation details. And they don't need to set it either, as hb.run will set it as needed.

Good point.

> Also, the store is too important to not be set, IMO. It's better to construct the object straightaway with as many key details set as possible, avoiding any invalid states.

I agree, but the client should not work without a call to configure. Perhaps we should enforce this 🤔 (i.e. throw an error if `configure` is not called)? Or go the other way make the `configure` method completely optional (i.e. read the `apiKey` from .env)?

> Finally, in our tests, we don't always call .configure, and having to add that to get things working was a pain.

Hehe, I faced the same scenario. I went ahead and added `configure` because otherwise we were not initializing the client correctly.

Anyway, I'm convinced 😄 and I think @joshuap is too, because he suggested adding the store initialization in the constructor too. I just didn't go for it to keep initializations in one place.

src/server.ts
subzero10: Thorough explanation, thanks!

subzero10 (approved):
All my questions have been answered/resolved, thanks @shalvah !

src/core/client.ts
shalvah: Since we're an error-reporting library, we can't afford to throw an error😅

At best we can write a log if there's no API key when we try to report, but I think we maybe already do that.

I don't think we need to change anything here at the moment. I think expecting the user to call `configure` is fair. It's always in the usage instructions, right next to the `require()` part.

shalvah:
@joshuap could we tag a new beta release so I can test this out on more examples?

joshuap:
> @joshuap could we tag a new beta release so I can test this out on more examples?

Yeah I'll release a new beta today

shalvah:
@joshuap 👀

PR #713, opened 2022-03-08
Include reported error link in logs

shalvah:
## Status
READY

## Description
Tiny logging improvement: print the full link to the reported error in the logs, like the Ruby library does.


joshuap (approved):
Nice, thanks!

PR #717, opened 2022-03-21
Node.js Context isolation, Part 2

shalvah:
Follow-up to #711, after some extensive testing and thinking.

## Backstory
There are two broad kinds of web frameworks in Node.js:
1. Middleware-based frameworks. Everything is a middleware, and you can write your own custom middleware that calls `next()` to manually advance a request. Two subcategories here: Express-based, and non-Express-based.
2. Non-middleware-based frameworks. Here, you can still write custom hooks/extensions, but control lies with the framework. You don't call `next()` manually to advance requests; just return a value, and the framework will control execution based on what you return.

Now, for both these types, we have two main concerns:
1. Can we reliably catch errors thrown in a request handler?
2. Can we correctly track and isolate context for each request (since a Node.js server may be handling multiple requests at once in the same thread)?

For MW frameworks, this is relatively simple:
1. Error-handling is typically done as a middleware, so we simply add our own.
2. Since _we_ are the ones who call `next()`, we can simply call it inside an AsyncLocalStorage scope, and the context will automatically be tracked and preserved for all operations done in that request. This is what we implemented in the last PR, with the `run` method.

For non-MW frameworks, it's harder:
1. Error handling is often via some specific mechanism, so the user has to manually do a `Honeybadger.notify` in that specific mechanism.
2. Since we don't control execution, we can't force the request handlers to execute in the scope of our async store. This means that `.run()` doesn't work.

## Solution
This PR introduces a new method, `.withRequest()`. Usage:

```js
app.get((req, res) => Honeybadger.withRequest(req, () => {
  // handle the request
}));

// in your error handler
Honeybadger.withRequest(req, () => {
  Honeybadger.notify(err);
});
```
`withRequest` simply saves the store contents on the request object, so, as long as you pass in the same request object, you can retrieve the store. 

See [the docs PR](https://github.com/honeybadger-io/docs/pull/162), where I've included more usage examples for some popular Node.js frameworks.

## Known supported frameworks:

Express-based:
- Restify
- Sails.js
- Nest.js (not tested, but uses Express middleware, so it should work)

Non-Express-based:
- Adonis.js 5
- Adonis.js 4
- Hapi
- Fastify

I made a basic app for most of these frameworks; perhaps I should upload that as some sort of "honeybadger examples" repo?🤔

There are definitely other frameworks out there, but I think `withRequest` covers them.

### Other notes
1. I removed `.run`, as it's now redundant since `withRequest` does the same. Also, the name `.withRequest` makes it a bit clearer that it's meant for web apps.
2. Sure, all that callback-wrapping is a bit aesthetically unappealing, but beyond making more framework-specific integrations, this is the most painless approach.


joshuap (commented):
I like it, thanks for putting this together. I will leave a more substantial technical review to @subzero10.

> I made a basic app for most of these frameworks; perhaps I should upload that as some sort of "honeybadger examples" repo?

Would it make sense to put them here? https://github.com/honeybadger-io/honeybadger-js/tree/master/examples

test/unit/server.test.ts
```diff
@@ -208,8 +208,8 @@ describe('server client', function () {
         context = client.__store.getStore().context;
```
subzero10, line 208: Since this is going to be called only once, you could make the assertions from L218-L219 and call `done()` here, right?

test/unit/server.test.ts
```diff
@@ -220,5 +220,24 @@ describe('server client', function () {
         done();
       }, 20);
     });
+
+    // eslint-disable-next-line
+    it('retrieves context from the same request object', (done) => {
+      let request1 = {};
+      let request2 = {};
```
subzero10, line 227: This is not used. Did you mean to use it?

src/server.ts
```diff
@@ -18,6 +18,7 @@ import { errorHandler, requestHandler } from './server/middleware'
 import { lambdaHandler } from './server/aws_lambda'
 import { AsyncStore } from './server/async_store'
 
+const kHoneybadgerStore = Symbol.for("kHoneybadgerStore");
```
subzero10, line 21: Curious, is there any specific reason you use a `Symbol` for the key to the store contents?

subzero10 (commented):
I love the new name `withRequest`.

test/unit/server.test.ts
shalvah: No, probably forgot. Good catch!

test/unit/server.test.ts
shalvah: True, but I didn't, because I wanted to make sure the error handler was called. Putting the assertions into the system under test means that the code could stop calling the error handler, and we wouldn't notice.

src/server.ts
subzero10: Thanks for elaborating!

> I think Symbols are not enumerable, so they won't show up in a for..in loop or similar

I just confirmed that this is the case. And this is a good reason to use a `Symbol` for this case.

test/unit/server.test.ts
subzero10: This is inside the `errorHandler` right? So the assertions (+ the call to `done()`) would only be called if the `errorHandler` is called.

shalvah: Yes, but if it isn't, then you'd have a test timeout exception rather than a failing assertion.

subzero10: Ah ok, good point.

shalvah:
Added them in the examples/ folder @joshuap

joshuap (commented):
These are awesome, but I didn't realize that some of the examples contained so many files. It looks like most of the files are from the adonis4 and adonis5 examples. I'm fine with having small examples in the repo (single-file, or a handful of files), but I'd prefer if we didn't have to maintain large/boilerplate examples; in my experience, they go stale fast and no one wants to upgrade them. One way I've tried to address that is by creating recipes:

https://honeybadger.notion.site/549be6376a6246e19d6291bf8bdcee69?v=02c2f9dfbe41498ebd8f43880fb4e782

It's easier to maintain a set of instructions for generating an example app than maintaining the example app itself.

I'd be OK with committing all of these examples if we could document a process to recreate them. Alternatively we could probably just create recipes for the adonis examples and leave the others as-is, but it would be nice to have a short list of instructions for each of them. The recipes can also rely on the Honeybadger docs if they already have instructions for adding Honeybadger to a boilerplate project. A recipe should tell me how to generate a boilerplate app that I can use when following the Honeybadger docs for each framework, to test that our instructions work.

What do you think?

joshuap (dismissed):
I'm going to approve this so that we can merge and release a beta before I leave for vacation next week. You could remove the adonis examples if you decide that a recipe would be better!

I'll release a new beta once this is merged.

shalvah:
I get what you mean re large files. I think I'll just remove the Adonis examples. I don't think I want to go with recipes, because I feel like they're more oriented towards internal use and could lead to some gaps in the main documentation. Basically, I think that between the Adonis docs and the Honeybadger docs, one should be able to recreate this setup pretty easily (it's a fairly basic integration). If that's not doable, then the docs would need to be updated.

shalvah:
@subzero10 final review please, so we can merge🙏

examples/fastify/index.js
```diff
@@ -0,0 +1,39 @@
...
+
+fastify.addHook('preHandler', Honeybadger.requestHandler);
+
+fastify.setErrorHandler((err, req, reply) => Honeybadger.withRequest(req, () => {
+    Honeybadger.notify(err)
+    reply.send("NOT OK")
+}));
+
+let reqId = 0;
+fastify.get('/test-context', function (request, reply) {
```
subzero10, line 13: Since this place is for _examples_, I am thinking that the code should be focused on setting up the framework with Honeybadger. This example is the setup + a test for an internal implementation detail (isolated context). I think we could keep it simpler and have 2 routes:
```
fastify.get('/unhandled', function (request, reply) {
  throw new Error('Unhandled error. Should be reported on Honeybadger dashboard');
})

fastify.get('/report', function (request, reply) {
  Honeybadger.notify('Hello World!')
  return 'Message should have been reported to Honeybadger! Please check your Honeybadger dashboard.'
  throw new Error('Unhandled error. Should be reported on Honeybadger dashboard');
})
```

What do you think?

examples/restify/index.js
```diff
@@ -0,0 +1,32 @@
+var restify = require('restify');
+const Honeybadger = require("../../dist/server/honeybadger.js")
+Honeybadger.configure({});
+
+var server = restify.createServer();
+
+server.use(Honeybadger.requestHandler)
+
+let reqId = 0;
+server.get('/test-context', (req, res, next) => {
```
subzero10, line 10: Same as above.

examples/sails/api/controllers/TestController.js
```diff
@@ -0,0 +1,23 @@
+
+const Honeybadger = require("../../../../dist/server/honeybadger.js")
+
+let reqId = 0;
+
+module.exports = {
+    context: (req, res, next) => {
```
subzero10, line 7: Same as above.

subzero10 (commented):
I have a small question about the examples. Other than that, I think we are good 👌

examples/fastify/index.js
shalvah: Done @subzero10

subzero10 (approved):

subzero10:
@shalvah Can you also modify `CHANGELOG.md` (mainly the change from `run` to `withRequest`) so that we can release a beta?

shalvah:
Will do. Do you have the permissions to tag a release?

subzero10:
> Will do. Do you have the permissions to tag a release?

Yes! Though we can use the CI to do it.

shalvah:
PR: #740

PR #724, opened 2022-03-31
#722 Respect toJSON in `breadrumb.metadata`

### LINKED ISSUE #722: `toJSON` not respected in honeybadger sanitize
1. Add breadcrumb with metadata containing an object with a `toJSON`
2. Send a honeybadger notice

## What happens?
The metadata field in the breadcrumb just says `[object Object]`

It would use `toJSON` to expand it and then sanitize

## Any other comments?
The reason we want to do this is we do our own sanitization and eventual stringify of the breadcrumb metadata field but we don't want to do this eagerly since it wastes time and only want to do it at the point of reporting the error.

**Operating System:** Linux
**Package Version:** 3.2.1
**Browser Version:** Firefox 98

subzero10:
## Status
**READY**

## Description
Fixes #722.
Respects `toJSON()` in `util.sanitize()`.

## Todos
- [x] Tests
- [x] getJson() method that respects toJSON of objects  

### DIFF
--- a/src/core/util.ts
+++ b/src/core/util.ts
@@ -134,8 +134,16 @@ export function sanitize(obj, maxDepth = 8) {
   }
 
   function canSerialize(obj) {
-    // Functions are TMI and Symbols can't convert to strings.
-    if (/function|symbol/.test(typeof (obj))) {
+    const typeOfObj = typeof obj
+
+    // Functions are TMI
+    if (/function/.test(typeOfObj)) {
+      // Let special toJSON method pass as it's used by JSON.stringify (#722)
+      return obj.name === 'toJSON'
+    }
+
+    // Symbols can't convert to strings.
+    if (/symbol/.test(typeOfObj)) {
       return false
     }
 
@@ -189,11 +197,6 @@ export function sanitize(obj, maxDepth = 8) {
 
   function safeSerialize(obj: unknown, depth = 0) {
     try {
-      if (Object.prototype.hasOwnProperty.call(obj, 'toJSON')) {
-        // @ts-ignore
-        return obj.toJSON()
-      }
-
       return serialize(obj, depth)
     } catch(e) {
       return `[ERROR] ${e}`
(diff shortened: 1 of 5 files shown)


src/core/util.ts
```diff
@@ -105,7 +105,7 @@ export function runAfterNotifyHandlers(notice: Notice | null, handlers: AfterNot
 }
 
 // Returns a new object with properties from other object.
-export function newObject<T>(obj: T): T | Record<string, unknown> {
+export function shallowClone<T>(obj: T): T | Record<string, unknown> {
```
subzero10, line 108: I think `shallowClone` makes better sense.

src/core/util.ts
```diff
@@ -116,6 +116,23 @@ export function newObject<T>(obj: T): T | Record<string, unknown> {
...
+  if (typeof obj.toJSON === 'function') {
+    return obj.toJSON()
+  }
+
+  const result = {}
+  for (const k in obj) {
+    result[k] = typeof obj[k] === 'object' ? getJson(obj[k]) : obj[k]
+  }
+
+  return result
```
subzero10, line 133: This could be simplified to `JSON.parse(JSON.stringify(obj))`, though I think the current solution is better for performance.

subzero10: I ended up using `JSON.parse(JSON.stringify(obj))` because integration tests where failing and I couldn't figure out why.

subzero10:
Here's a working example:
Screenshot 2022-04-01 at 8 58 09 PM
Screenshot 2022-04-01 at 8 58 14 PM

src/core/util.ts
```diff
@@ -134,8 +134,14 @@ export function sanitize(obj, maxDepth = 8) {
   }
 
   function canSerialize(obj) {
-    // Functions are TMI and Symbols can't convert to strings.
-    if (/function|symbol/.test(typeof (obj))) {
+    // Functions are TMI
+    if (/function/.test(typeof (obj))) {
+      // Let special toJSON method pass as it's used by JSON.stringify (#722)
+      return obj.name === 'toJSON';
```
subzero10, line 140: @joshuap Here's the solution we discussed about.

joshuap (approved):
Looks great, thanks!

subzero10:
@shipjs prepare

PR #745, opened 2022-04-06
Fix shipjs prepare

subzero10:
Fixes shipjs prepare.

It broke because I merged a PR with a package-lock.json generated from npm 6. The workflow runs with npm 8 and this generated a new package-lock.json file which causes the release to fail (working tree is not clean).


subzero10:
@shalvah Can I please get a quick approval on this? Thanks!

shalvah (approved):

subzero10:
@shipjs prepare

PR #754, opened 2022-04-12
#725 Honeybadger.checkIn()

### LINKED ISSUE #725: Add `Honeybadger.checkIn()` in Honeybadger js client

subzero10:
## Status
**READY**

## Description
Fixes #725 
In order to create a single `Honeybadger.checkIn()` method to be shared between browser and server, I decided to extract the transport layer logic into a separate interface. This way, we can abstract the browser vs server differences and implement a single `checkIn()` method to be used by both. I refactored `notify()` to use the same approach.

Notes:
- This is the first version of `Honeybadger.checkIn()` that simply pings Honeybadger from inside the code. There is ongoing discussion to add more features into this implementation (maintain/create check-ins through code).
- I noticed that the check-in API call does not require an API key, is that intended? Aren't there security concerns with this approach? @joshuap 
- The check-in API call will return HTTP OK even if the check-in id does not exist.

## Todos
- [x] Trasport Interface
- [x] Server Transport
- [x] Browser Transport
- [x] Honeybadger.checkIn()
- [x] Tests
- [x] Fix CI Tests
- [x] Documentation

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -1,12 +1,13 @@
 import Client from './core/client'
-import { Config, Notice, BeforeNotifyHandler } from './core/types'
-import { merge, sanitize, filter, runAfterNotifyHandlers, objectIsExtensible, endpoint } from './core/util'
+import { Config, Notice, BeforeNotifyHandler, TransportPayload } from './core/types'
+import { merge, filter, objectIsExtensible } from './core/util'
 import { encodeCookie, decodeCookie, preferCatch } from './browser/util'
 import { onError, ignoreNextOnError } from './browser/integrations/onerror'
 import onUnhandledRejection from './browser/integrations/onunhandledrejection'
 import breadcrumbs from './browser/integrations/breadcrumbs'
 import timers from './browser/integrations/timers'
 import eventListeners from './browser/integrations/event_listeners'
+import { BrowserTransport } from "./browser/transport";
 
 interface BrowserConfig extends Config {
   async: boolean
@@ -36,6 +37,8 @@ class Honeybadger extends Client {
 
       if (!notice.url) { notice.url = document.URL }
 
+      this.__incrementErrorsCount()
+
       return true
     }
   ]
@@ -46,7 +49,7 @@ class Honeybadger extends Client {
       maxErrors: null,
       projectRoot: window.location.protocol + '//' + window.location.host,
       ...opts
-    })
+    }, new BrowserTransport())
   }
 
   configure(opts: Partial<BrowserConfig> = {}) {
@@ -57,12 +60,12 @@ class Honeybadger extends Client {
     return (this.__errorsSent = 0)
   }
 
-  factory(opts?: Partial<BrowserConfig>): Honeybadger {
+  public factory(opts?: Partial<BrowserConfig>): Honeybadger {
     return new Honeybadger(opts)
   }
 
   /** @internal */
-  protected __buildPayload(notice:Notice): Record<string, Record<string, unknown>> {
+  protected __buildPayload(notice:Notice): TransportPayload {
     const cgiData = {
       HTTP_USER_AGENT: undefined,
       HTTP_REFERER: undefined,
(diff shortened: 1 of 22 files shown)


src/core/client.ts
```diff
@@ -40,18 +40,20 @@ const TAG_SEPARATOR = /,|\s+/
 // Checks for non-blank characters
 const NOT_BLANK = /\S/
 
-export default class Client {
+export default abstract class Client {
```
subzero10, line 43: By marking as abstract, we are restricting instances to only derived classes.

src/core/client.ts
```diff
@@ -204,7 +205,35 @@ export default class Client {
...
+            const uuid = JSON.parse(res.body).id
+            runAfterNotifyHandlers(merge(notice, {
+              id: uuid
+            }), this.__afterNotifyHandlers)
+            this.logger.info(`Error report sent ⚡ https://app.honeybadger.io/notice/${uuid}`)
+          })
+          .catch(err => {
+            this.logger.error('Error report failed: an unknown error occurred.', `message=${err.message}`)
+            runAfterNotifyHandlers(notice, this.__afterNotifyHandlers, err)
+          })
```
subzero10, line 236: All this was duplicated because we had to replicate same behavior in both server and browser.

src/core/client.ts
```diff
@@ -255,6 +284,27 @@ export default class Client {
     })
   }
 
+  checkIn(id: string): void {
+    this.checkInAsync(id).then()
```
subzero10, line 288: `.then()` is simply there so that the IDE won't complain that the promise result is unhandled.

subzero10:
@shalvah The code is ready for review, feel free to take a look and share feedback. I did not mark as ready because of the blocking CORS issue (see PR description).

src/core/client.ts
```diff
@@ -255,6 +284,27 @@ export default class Client {
     })
   }
 
+  checkIn(id: string): void {
```
shalvah, line 287: See my comment on the docs PR.

src/core/types.ts
```diff
@@ -96,5 +101,60 @@ export interface CGIData {
...
+ }
+}
+
+export type TransportOptions = {
+  method: 'GET' | 'POST',
+  headers?: Record<string, number | string | string[] | undefined>,
+  endpoint: string,
+  maxObjectDepth?: number,
+  logger: Logger
+  async?: boolean // don't like this here because it's only for browser
```
shalvah, line 125: Lool. I wonder if there are any users that actually want their reporting to block the thread. There probably are.🤔

src/server/transport.ts
```diff
@@ -0,0 +1,71 @@
...
+export class ServerTransport implements Transport {
+    send(options: TransportOptions, payload?: NoticeTransportPayload | undefined): Promise<{ statusCode: number; body: string; }> {
+        const {protocol} = new URL(options.endpoint)
+        const transport = (protocol === "http:" ? http : https)
+
+        return new Promise((resolve, reject) => {
+            let promise: Promise<void>;
+
+            // this should not be here. it should be done before reaching the transport layer
+            // it could be inside a beforeNotifyHandler, but is not possible at the moment because those handlers are synchronous
```
shalvah, line 17: This is something we really should change in future. The fix might actually be as simple as invoking each handler with `await`.🤔

shalvah (changes requested):
Mostly good, but see my comments.

src/core/client.ts
subzero10: Done!

src/server/transport.ts
subzero10: True, I already suggested this but it may be trickier to push to our users. It will be a breaking change.

src/core/types.ts
subzero10: 🤷

subzero10:
> My main reservation is the async thing.

Same, hence the comment. I think we should be able to remove at some point. Could it be with v4? @joshuap

shalvah:
Oh, I actually meant the `checkInAsync` thing @subzero10

src/server/transport.ts
shalvah: Is it really? You can call synchronous functions with `await`.The main difference is that errors will now be promise rejections.

Oh, except you mean that the `notify` will now end up being promise-based if we do that (since you can only use `await` from an async function).

shalvah (approved):

src/server/transport.ts
subzero10: > Oh, except you mean that the notify will now end up being promise-based if we do that (since you can only use await from an async function).

This 👍

subzero10:
Note: We decided **not** to support `Honeybadger.checkIn()` on the browser at this moment because of CORS issues.

src/browser.ts
```diff
@@ -57,12 +55,16 @@ class Honeybadger extends Client {
     return (this.__errorsSent = 0)
   }
 
-  factory(opts?: Partial<BrowserConfig>): Honeybadger {
+  public factory(opts?: Partial<BrowserConfig>): Honeybadger {
     return new Honeybadger(opts)
   }
 
+  public checkIn(_id: string): Promise<void> {
```
subzero10, line 62: This is only here because of Typescript. Our TS definitions will show that `checkIn` is available (because we export `Server & Browser` in a single object. So I added this to throw an more specific error instead of an error like _Uncaught TypeError: Honeybadger.checkIn is not a function_

shalvah (approved):

PR #758, opened 2022-04-16
#679 AWS Timeout Warning

### LINKED ISSUE #679: AWS Lambda integration: Add timeout warning
### Discussed in https://github.com/honeybadger-io/honeybadger-js/discussions/674

Add a timeout warning when the lambda function is within a number of milliseconds from tis execution time (15 minutes).

subzero10:
## Status
**READY**

## Description
Fixes #679
Report to Honeybadger if AWS lambda function may possibly timeout.

## Todos
- [x] Create `ServerlessConfig` for `reportTimeoutWarning` and `timeoutWarningThresholdMs`
- [x] Send timeout warning for callback-based functions
- [x] Send timeout warning for async functions
- [x] Tests
- [x] Manual Tests
- [x] [Documentation](https://github.com/honeybadger-io/docs/pull/171)

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -49,7 +49,7 @@ class Honeybadger extends Client {
 
     // serverless defaults
     const config = this.config as ServerlessConfig
-    config.reportTimeoutWarning = config.reportTimeoutWarning || true
+    config.reportTimeoutWarning = config.reportTimeoutWarning ?? true
     config.timeoutWarningThresholdMs = config.timeoutWarningThresholdMs || 50
 
     this.__getSourceFileHandler = getSourceFile.bind(this)
--- a/src/server.ts
+++ b/src/server.ts
@@ -5,7 +5,7 @@ import os from 'os'
 import domain from 'domain'
 
 import Client from './core/client'
-import { Config, Notice, BeforeNotifyHandler, DefaultStoreContents } from './core/types'
+import { Config, Notice, BeforeNotifyHandler, DefaultStoreContents, ServerlessConfig } from './core/types'
 import { merge, sanitize, runAfterNotifyHandlers, endpoint } from './core/util'
 import {
   fatallyLogAndExit,
@@ -33,25 +33,37 @@ class Honeybadger extends Client {
     }
   ]
 
-  public errorHandler: typeof errorHandler;
-  public requestHandler: typeof requestHandler;
-  public lambdaHandler: typeof lambdaHandler;
+  public errorHandler: typeof errorHandler
+  public requestHandler: typeof requestHandler
+  public lambdaHandler: typeof lambdaHandler
 
-  constructor(opts: Partial<Config> = {}) {
+  config: Config | ServerlessConfig
+
+  constructor(opts: Partial<Config | ServerlessConfig> = {}) {
     super({
       afterUncaught: fatallyLogAndExit,
       projectRoot: process.cwd(),
       hostname: os.hostname(),
       ...opts,
     })
+
+    // serverless defaults
+    const config = this.config as ServerlessConfig
+    config.reportTimeoutWarning = config.reportTimeoutWarning || true
+    config.timeoutWarningThresholdMs = config.timeoutWarningThresholdMs || 50
+
     this.__getSourceFileHandler = getSourceFile.bind(this)
     this.errorHandler = errorHandler.bind(this)
     this.requestHandler = requestHandler.bind(this)
     this.lambdaHandler = lambdaHandler.bind(this)
   }
 
-  factory(opts?: Partial<Config>): Honeybadger {
+  factory(opts?: Partial<Config | ServerlessConfig>): Honeybadger {
     return new Honeybadger(opts)
+  }
+
+  configure(opts: Partial<Config | ServerlessConfig> = {}): Honeybadger {
+    return super.configure(opts) as Honeybadger
   }
 
   /** @internal */
--- a/src/server/aws_lambda.ts
+++ b/src/server/aws_lambda.ts
@@ -82,7 +82,7 @@ function setupTimeoutWarning(hb: Honeybadger, context: Context) {
 
     const delay = context.getRemainingTimeInMillis() - ((hb.config as ServerlessConfig).timeoutWarningThresholdMs)
     return setTimeout(() => {
-        hb.notify(`${context.functionName}[${context.functionVersion}] may timeout`)
+        hb.notify(`${context.functionName}[${context.functionVersion}] may have timed out`)
     }, delay > 0 ? delay : 0)
 
 }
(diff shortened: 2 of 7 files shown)


subzero10:
Tested manually by adding a handler that never resolves in `examples/aws-lambda`:
Screenshot 2022-04-17 at 9 48 37 AM

src/server.ts
```diff
@@ -33,27 +33,39 @@ class Honeybadger extends Client {
...
     super({
       afterUncaught: fatallyLogAndExit,
       projectRoot: process.cwd(),
       hostname: os.hostname(),
       ...opts,
     })
+
+    // serverless defaults
+    const config = this.config as ServerlessConfig
+    config.reportTimeoutWarning = config.reportTimeoutWarning || true
```
shalvah, line 52: Bug: I think this should be `??` (which will probably get compiled down to some older syntax). `||` will override the value if it's false.

src/server.ts
```diff
@@ -33,27 +33,39 @@ class Honeybadger extends Client {
...
+    config.timeoutWarningThresholdMs = config.timeoutWarningThresholdMs || 50
```
shalvah, line 53: This too. Is `0` an acceptable value? If so, this should be `??`.

src/server/aws_lambda.ts
```diff
@@ -49,21 +53,40 @@ function syncHandler<TEvent = any, TResult = any>(handler: SyncHandler<TEvent, T
...
+}
+
+function setupTimeoutWarning(hb: Honeybadger, context: Context) {
+    if (!shouldReportTimeoutWarning(hb, context)) {
+        return
+    }
+
+    const delay = context.getRemainingTimeInMillis() - ((hb.config as ServerlessConfig).timeoutWarningThresholdMs)
+    return setTimeout(() => {
+        hb.notify(`${context.functionName}[${context.functionVersion}] may timeout`)
```
shalvah, line 85: Maybe "may have timed out?" When they see the notification, it'll be in the past.

src/server/aws_lambda.ts
```diff
@@ -49,21 +53,40 @@ function syncHandler<TEvent = any, TResult = any>(handler: SyncHandler<TEvent, T
...
+    }, delay > 0 ? delay : 0)
```
shalvah, line 86: I'm thinking about this. If the `timeoutWarningThresholdMs` is set to 50ms, does that mean we'll be triggering a callback possibly every 50ms? Wouldn't that be too much of an impact on the function's runtime/performance?

Maybe we should decouple the callback interval from the threshold? We could check, say, every second or 3/4 second. What do you think?

src/server/aws_lambda.ts
```diff
@@ -49,21 +53,40 @@ function syncHandler<TEvent = any, TResult = any>(handler: SyncHandler<TEvent, T
...
+function shouldReportTimeoutWarning(hb: Honeybadger, context: Context): boolean {
+    return typeof context.getRemainingTimeInMillis === 'function' && !!((hb.config as ServerlessConfig).reportTimeoutWarning)
+}
+
+function setupTimeoutWarning(hb: Honeybadger, context: Context) {
+    if (!shouldReportTimeoutWarning(hb, context)) {
+        return
+    }
+
+    const delay = context.getRemainingTimeInMillis() - ((hb.config as ServerlessConfig).timeoutWarningThresholdMs)
```
shalvah, line 83: `delay`'s a bit of a confusing name for me here. Maybe `timeLeftToMarkAsTimeout`? A bit wordy, I know, but I wasn't sure what was being delayed at first (which kinda ties into my other point of decoupling the two.)

src/server/aws_lambda.ts
shalvah, line 85: Also, has this been tested in a real function? Asking because I recently encountered sth that worked on a local emulation but failed on AWS. For instance, in a local emulation, the process might hang around for a while, while Lambda might kill it at a point. If you can't test it live, I think their [runtime client](https://github.com/aws/aws-lambda-nodejs-runtime-interface-client) are pretty much how Lambda invokes functions.

shalvah (changes requested):

src/server.ts
subzero10: Ah yes, correct. Good catch!

src/server.ts
subzero10: `0` should not be acceptable here, as it would probably fail to report to Honeybadger. I think `||` is OK here.

src/server/aws_lambda.ts
subzero10: I'm not sure I'm following.
The logic is:
- Check the remaining time left for the lambda function and remove `timeoutWarningThresholdMs` (default 50ms) = `delay`
- Schedule a callback with the `delay` to fire and notify Honeybadger

Example:
- Function has `getRemainingTimeInMillis = 3000` (3 seconds)
- Schedule a callback to fire in 2950ms and notify Honeybadger
- When [If] the function resolves, we cancel the callback with `clearTimeout()`

Does this make sense?

src/server/aws_lambda.ts
subzero10: I name it `delay` because that's how much we _delay_ the execution of the callback that notifies Honeybadger. I have no objection with longer variable names 😄  ; if you still think it's confusing, we can change it.

src/server/aws_lambda.ts
subzero10: Good point!

src/server/aws_lambda.ts
shalvah: Oh, my bad. I read that as setInterval() for some reason. Cool.

src/server/aws_lambda.ts
subzero10: Yes I [have tested](https://github.com/honeybadger-io/honeybadger-js/pull/758#issuecomment-1100817169)! I also pushed the example lambda function, but if you want to try it you will need to install the build from this branch (until this is merged):
```
cd examples/aws-lambda
npm i honeybadger-io/honeybadger-js#aws_timeout_warning_#679
```

shalvah: Ooh, I missed that comment. All good then.

shalvah (approved):

joshuap (commented):
So basically—when enabled—this uses `setTimeout` to set a timeout for the max execution time of the function - 50ms, and notifies Honeybadger if the function hasn't completed by then?

I will let @shalvah do the final approval.

subzero10:
> So basically—when enabled—this uses `setTimeout` to set a timeout for the max execution time of the function - 50ms, and notifies Honeybadger if the function hasn't completed by then?
> 
> I will let @shalvah do the final approval.

Correct. FYI, it will be enabled by default.
@shalvah already approved but I had to pull/merge master so I had to request a new PR review.

joshuap (approved):
Cool, I can approve then, if that's all you need.

PR #776, opened 2022-05-06
Improve typings for optional deps

subzero10:
## Status
**READY**

## Description
Our d.ts files import types for `node v14` and `aws-lambda`. These deps are optional and this PR makes sure they don't cause type errors when Honeybadger is used with Typescript.

More on `ts-ignore` [here](https://github.com/microsoft/TypeScript/issues/29866#issuecomment-715140696).
More on `import type` [here](https://stackoverflow.com/a/63905376).


joshuap (approved):

subzero10:
@shipjs prepare

PR #787, opened 2022-05-11
fix: make sure shipjs workflows do not modify package-lock.json

subzero10:
[Last shipjs prepare](https://github.com/honeybadger-io/honeybadger-js/runs/6394578790?check_suite_focus=true) failed because working tree was not empty. I suspect it's because `npm install` modified the `package-lock.json` file. We won't have this problem with `npm ci`.


shalvah (approved):

subzero10:
@shipjs prepare

PR #788, opened 2022-05-11
fix: improve typings

subzero10:
## Status
**READY**

## Description
An attempt to fix this [honeybadger-react issue](https://github.com/honeybadger-io/honeybadger-react/runs/6336668195?check_suite_focus=true):
- Another Server vs Browser Typescript issue.

I tested manually with AWS Lambda (in typescript) & React because we had this same error reported again in the past.

Can't wait to get this inside a monorepo and release separate packages 😄 

## Todos
- [x] Tests
- [x] Test with honeybadger-react
- [x] Test with honeybadger-vue - not written in TS, so no issues
- [x] Test with aws lambda handler

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -87,7 +87,10 @@ jobs:
           cache: 'npm'
 
       - name: Build
-        run: npm ci
+        run: |
+          npm ci
+          cd examples/aws-lambda-typescript
+          npm ci
 
       - name: Run lint checks
         run: npm run lint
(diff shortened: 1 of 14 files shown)


.github/workflows/node.js.yml
```diff
@@ -87,7 +87,10 @@ jobs:
           cache: 'npm'
 
       - name: Build
-        run: npm ci
+        run: |
+          npm ci
+          cd examples/aws-lambda-typescript
+          npm ci
```
shalvah, line 93: Why is this in the CI? Are we running tests in the example folder?

shalvah (approved):

.github/workflows/node.js.yml
subzero10: No. But I enabled eslint in examples folder to make sure all codebase conforms to it. Since this is a typescript project, eslint was trying to check types and it kept failing because it couldn't find any modules installed.

PR #793, opened 2022-05-16
fix: make sure file is bundled with rollup

subzero10:
## Status
**READY**

## Description
I was trying to fix honeybadger-react issues with v4.0.1 and v4.0.2. The latest issue is seen [here](https://github.com/honeybadger-io/honeybadger-react/runs/6458315730?check_suite_focus=true):
- The rollup bundle does not include the `core/store.js` because it didn't see the require statement hidden inside the catch block.


joshuap (approved):
👍

subzero10:
@shipjs prepare

PR #828, opened 2022-07-09
chore: added ESM support

### LINKED ISSUE #823: ESM Support: https://honeybadger.slack.com/archives/C7ET3UGD6/p1652717240087049

subzero10:
## Status
**READY**

## Description

Fixes #823.


subzero10:
Integration tests are failing. Looking into it.

package.json
```diff
@@ -21,6 +21,7 @@
   ],
   "main": "dist/server/honeybadger.js",
   "browser": "dist/browser/honeybadger.js",
+  "esm": "dist/esm/honeybadger.js",
```
KonnorRogers, line 24: Is ESM the standard key? I usually use `"module"`, knowing npm, they're probably both right...

joshuap: > Is ESM the standard key? I usually use `"module"`, knowing npm, they're probably both right...

I'm not sure, tbh.

subzero10: Yeah probably it [should](https://esbuild.github.io/api/#main-fields) be `module`. I'm not sure why it's `esm` 🤔.

joshuap: > Yeah probably it [should](https://esbuild.github.io/api/#main-fields) be `module`. I'm not sure why it's `esm` 🤔.

Probably because I copied it from an example somewhere. 🤠

subzero10: Actually, I should take the blame on this one 😅 .
I cherry picked these changes from a fork of our repo (not available any more).

BethanyBerkowitz:
@subzero10 what will happen if someone is using this package in an esm module in Node? Will they potentially end up with the browser version instead of the server version?

subzero10:
> @subzero10 what will happen if someone is using this package in an esm module in Node? Will they potentially end up with the browser version instead of the server version?

I had the same question my self. It's very confusing to say the least:
- esbuild says [here](https://esbuild.github.io/api/#main-fields) that nodejs does not support the `"module"` field. nodejs does not mention the `"module"` in their [docs](https://nodejs.org/api/packages.html#package-entry-points) either.
- esbuild has a recommended section for [package authors](https://esbuild.github.io/api/#main-fields-for-package-authors). So I went ahead and applied this. I tested with webpack. Apparently, webpack will always import the esm build, even if require is used (as in our example project). This broke the bundle, though I was able to fix it by changing the require to an import statement, or forcing webpack to [resolve the cjs build](https://stackoverflow.com/questions/69333759/force-webpack-4-to-import-cjs-only-and-never-esm-with-typescript).

That's why I opted for the simplest non-breaking change, even though it's recommended to avoid 🤷 .
Once again, having the package isomorphic makes things more complicated than they should. I am hoping that soon we'll be able to break the js package to separate packages for node vs browser.

joshuap:
@subzero10 can we close this issue since it’s stale or do you want to keep it open?

subzero10:
I'm closing and we can revisit if we receive new requests for ESM support.

PR #834, opened 2022-07-15
fix: check for null

### LINKED ISSUE #829: TypeError: Cannot read properties of null (reading 'url') when reporting is disabled

1. Initialize Honeybadger with `reportData: false` on a browser.
2. Call `Honeybadger.notify(new Error())`.

## What happens?

`TypeError: Cannot read properties of null (reading 'url')` is thrown.

No error is thrown.

…

## Any other comments?

This is caused by https://github.com/honeybadger-io/honeybadger-js/blob/v4.0.3/src/browser.ts#L33.

**Operating System:** …
**Package Version:** v4
**Browser Version:** …

subzero10:
## Status
**READY**

## Description
Fixes #829.

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -24,13 +24,13 @@ class Honeybadger extends Client {
 
   /** @internal */
   protected __beforeNotifyHandlers: BeforeNotifyHandler[] = [
-    (notice: Notice) => {
+    (notice?: Notice) => {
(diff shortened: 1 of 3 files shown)


joshuap (approved):
👍

subzero10:
@shipjs prepare

subzero10:
@shipjs prepare

Issue #825, opened 2022-07-05
no notifications with sailsJs + docker


subzero10:
### Discussed in https://github.com/honeybadger-io/honeybadger-js/discussions/824

<div type='discussions-op-text'>

<sup>Originally posted by **atex21** July  5, 2022</sup>
Hey!
I'm trying to test honeybadger but it's being impossible.. I'm with sailsJS and docker, configurations as described in documentation:
```
const honeybadgerConfig = {
  apiKey: process.env.HONEYBADGER_API_KEY,
  environment: environment
}
```
variables arrive well to the container in docker.

In a first hand I realized that that configuration about middleware doesn't work.. If I just throw an exception there is no try to send honeybadger notification, the errorHandler is not called =/

In the other hand, if I forgot the last point and use an explicit HoneyBadger.notify, then there is a try to send notification, however I'm stuck with errors coming for no reason =/, namely

Conditions of running | honeybadger-io/js version | Result of a Honeybadger.notify
-- | -- | --
Throw in class + notify in servererror | v3.2.9 | (1)TypeError [ERR_INVALID_ARG_TYPE]: The "listener" argument must be of type function. Received an instance of Object
Throw in class | v3.2.9 | Don’t notify / don’t arrive to errorHandler
Throw in class + notify in servererror | V4.0.3 | (2)TypeError: Cannot read property 'breadcrumbs' of undefined
Throw in class | V4.0.3 | Don’t notify / don’t arrive to errorHandler

a bit more detail:

```
(1)
[API] TypeError [ERR_INVALID_ARG_TYPE]: The "listener" argument must be of type function. Received an instance of Object
[API]     at checkListener (events.js:135:11)
[API]     at ClientRequest.once (events.js:539:3)
[API]     at new ClientRequest (_http_client.js:215:10)
[API]     at Object.request (https.js:374:10)
[API]     at Object.wrappedRequest (/usr/src/app/node_modules/newrelic/lib/instrumentation/core/http.js:412:22)
[API]     at Object.<anonymous> (/usr/src/app/node_modules/agent-base/patch-core.js:25:22)
[API]     at Object.<anonymous> (/usr/src/app/node_modules/socks-proxy-agent/node_modules/agent-base/patch-core.js:23:20)
[API]     at Object.<anonymous> (/usr/src/app/node_modules/agent-base/patch-core.js:25:22)
[...]

PR #843, opened 2022-07-23
Workaround for Node.js-patching libraries

shalvah:
Workaround for [#825](https://github.com/honeybadger-io/honeybadger-js/issues/825#issuecomment-1193113433)

### DIFF
--- a/src/server/transport.ts
+++ b/src/server/transport.ts
@@ -7,7 +7,7 @@ import { sanitize } from '../core/util';
 
 export class ServerTransport implements Transport {
   send(options: TransportOptions, payload?: NoticeTransportPayload | undefined): Promise<{ statusCode: number; body: string; }> {
-    const { protocol } = new URL(options.endpoint)
+    const { protocol, hostname, pathname } = new URL(options.endpoint)
     const transport = (protocol === 'http:' ? http : https)
 
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


subzero10 (approved):
This looks harmless 😅

subzero10:
@shipjs prepare

subzero10:
@shalvah I cannot release a new version with this PR because it does not follow conventional commits syntax and therefore it picked up nothing to release:
```
› Calculating the next version.
The following commit messages out of convention are ignored:
  Merge pull request #843 from honeybadger-io/nodejs-patching-fix
  Workaround for Node.js-patching libraries
› Confirming the next version.
The current version: 4.0.4
The next version: 4.0.5
› Validating preparation conditions.
`shouldPrepare` returned false. So the preparation is skipped.
```
Did you not receive an error when trying to commit from your local environment? Or did you disable git hooks?
(note for me: update [this section](https://github.com/honeybadger-io/honeybadger-js#contributing) to inform about commit syntax)

I will publish a new version when [this](https://github.com/honeybadger-io/honeybadger-js/pull/845) is merged.

shalvah:
> I cannot release a new version with this PR because it does not follow conventional commits syntax and therefore it picked up nothing to release

Isn't there a way to override it? Seems like a flaw if that can't be overridden.

> Did you not receive an error when trying to commit from your local environment

The other one did; this didn't.

subzero10:
> > I cannot release a new version with this PR because it does not follow conventional commits syntax and therefore it picked up nothing to release
> 
> Isn't there a way to override it? Seems like a flaw if that can't be overridden.

Not from what I know. And it makes sense, because CHANGELOG is generated from the commits so they must follow the syntax. The workaround would be to push another dummy commit (new PR) with the right syntax.
> 
> > Did you not receive an error when trying to commit from your local environment
> 
> The other one did; this didn't.

This is strange. Did you pull latest before creating this branch?

subzero10:
> The workaround would be to push another dummy commit (new PR) with the right syntax.

I will do this now so we can proceed with the release.

subzero10:
@shipjs prepare

subzero10:
Done! Released `v4.0.5`.

PR #849, opened 2022-07-28
Support nested exceptions (error.cause)

### LINKED ISSUE #847: Support `Error.cause`
[`error.cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) was [recently standardized in JavaScript](https://2ality.com/2021/06/error-cause.html) (but it's been around on some platforms before that).

`error.cause` is useful in nested error situations, such as when we encounter an error while trying to report an error (as recently in #825 ). In such scenarios, `.cause` would be set to the original error, so we have enough information to investigate the issue. Several other languages, like PHP and Ruby support this or something similar.

I'm not sure if we already support it, but we should. "Support" would mean being able to display a "Caused by ..." pointing to the original error + stacktrace. This might be more of a backend issue than a client issue, but will investigate.

shalvah:
## Status
**READY**

Closes #847

image

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -75,6 +75,8 @@ export interface Notice {
   details: Record<string, Record<string, unknown>>,
   __breadcrumbs: BreadcrumbRecord[],
   afterNotify?: AfterNotifyHandler,
+  cause?: Error|Record<string, unknown>,
+  causes: Array<{class: string, message: string, backtrace: BacktraceFrame[]}>,
   [key: string]: unknown
 }
 
@@ -141,8 +143,8 @@ export type NoticeTransportPayload = {
     enabled: boolean,
     trail: BreadcrumbRecord[]
   },
-  error: Pick<Notice, 'message' | 'backtrace' | 'fingerprint' | 'tags'> & {
-    class: string
+  error: Pick<Notice, 'message' | 'backtrace' | 'fingerprint' | 'tags' | 'causes'> & {
+    class: string,
   },
   request: Pick<Notice, 'url' | 'component' | 'action' | 'context' | 'params' | 'session'> & {
     cgi_data: unknown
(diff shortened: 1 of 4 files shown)


src/core/types.ts
```diff
@@ -75,6 +75,8 @@ export interface Notice {
   details: Record<string, Record<string, unknown>>,
   __breadcrumbs: BreadcrumbRecord[],
   afterNotify?: AfterNotifyHandler,
+  cause?: Error|Record<string, unknown>,
+  causes: Array<{class: string, message: string, backtrace: BacktraceFrame[]}>,
```
subzero10, line 79: What is the need for causes if you always have 1 cause? Or am I missing something here?

shalvah: It's the API spec. But I think you're right—it should probably be on the payload, not the notice object.

shalvah: On second thought, perhaps I'll keep it as is🤔. It's a bit messy because we use the `Notice` type to represent both what we get (possibly an Error object) and what we produce from that. The `cause` property is from the Error object. The `causes` is the resolved value for the API, which we need to generate earlier, because of all the functions and transformations the notice object passes through.

shalvah: But you've reminded me that it might not be a bad idea to expand causes (ie follow the `.cause`'s cause, and so on. I've updated it to do this now. so there may be more than one cause.

src/core/util.ts
```diff
@@ -59,6 +59,23 @@ export function makeBacktrace(stack: string, shift = 0): BacktraceFrame[] {
   }
 }
 
+export function getCauses(notice: Partial<Notice>) {
+  if (notice.cause) {
+    const causes =[]
+    let cause = notice as Error
+    while (causes.length < 3 && (cause = cause.cause) as Error) {
```
subzero10, line 66: Nice, I haven't used `as` in this way. I had to try it out in a playground!
image
But I'm curious to see how `(cause = cause.cause) as Error` compiles into browser parseable expression 😅

subzero10 (approved):

PR #854, opened 2022-08-10
chore: `notice.causes` refactor

subzero10:
## Status
**READY**

## Description
Refactor `notice.causes` outside the `Notice` type, since it's only used when sending the payload to Honeybadger.

## Todos
- [x] Tests


subzero10:
Hey @shalvah, while I was merging master to the monorepo branch, I realized that I could move `causes` from the `Notice` to `NoticeTransportPayload`, then call `getCauses` in `buildPayload`. The tests are passing. Is this OK for you?

shalvah:
Yeah, that would work.

PR #862, opened 2022-08-19
fix: improve type definition for GlobalStore

subzero10:
## Status
**READY**

## Description
Updates the type definition of `GlobalStore` to work with the latest version of node typings.
Discussed on [slack](https://honeybadger.slack.com/archives/C7ET3UGD6/p1660838115322689).


joshuap (approved):
thanks!

Issue #864, opened 2022-08-31
`Configure` not found in '@honeybadger-io/js'


p8:
## What are the steps to reproduce this issue?

After upgrading to version 4.x the following code no longer works with Webpacker.
```
// Copied from https://docs.honeybadger.io/lib/javascript/integration/stimulus/
import * as Honeybadger from '@honeybadger-io/js'

Honeybadger.configure({
  apiKey: 'Your Honeybadger API key',
  environment: 'production',
  revision: 'git SHA/project version'
})
```

## What happens?
I get the following error:
```
Failed to compile.

./app/javascript/packs/application.js 29:0-21
"export 'configure' (imported as 'Honeybadger') was not found in '@honeybadger-io/js'
```

## What versions are you using?
**Webpacker:** 5.4.3
**Package Version:** Tested 4.0.5 and 4.1.1

subzero10:
Hello @p8, can you try importing like below and let me know if it works for you?
```
import Honeybadger from '@honeybadger-io/js'
```

p8:
Hi @subzero10 then I get:
`export 'default' (imported as 'Honeybadger') was not found in '@honeybadger-io/js'`

subzero10:
The difference between v3 and v4 of the Honeybadger client is that on v3 the typescript compiler transpiles to es5 and v4 transpiles to es6. There are some more details in this process, because this is the time we moved to the monorepo but after a few hours of testing, I believe this is what broke compatibility with Webpacker.

I was able to reproduce this issue as @p8 mentioned above with Webpacker. I'm not very familiar with Webpacker but what I know is that it is a webpack wrapper written in Ruby. My guess is that some configuration that comes from webpacker is causing the problem. I say that because I was able to include `@honeybadger-io/js@4.1.2` in a simple webpack project and it worked. There is also an example project [here](https://github.com/honeybadger-io/honeybadger-js/tree/master/packages/js/examples/webpack). I tried `@honeybadger-io/js@4.1.2` with webpack v4 and webpack v5. Everything works OK with the default configuration.

While debugging, I came to realize that we should probably improve the way we export the javascript honeybadger client. For example, we use rollup to provide different bundles of the client. 
For the browser client, it makes sense to have a bundle.
But what about the nodejs client? Should it be bundled?
What about this case, where the client is to be consumed by a bundler? Ideally the bundler should take the source code (or the transpiled ts code) and decide how to bundle it, right?
[...]

PR #866, opened 2022-09-02
fix: revert to es5 transpilation

subzero10:
## Status
**READY**

## Description
Reverts back to es5 transpilation because there were issues with some bundlers.

Fixes: #864

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -1,9 +1,7 @@
 {
   "compilerOptions": {
     "allowJs": false,
-    "target": "es6",
-    "module": "commonjs",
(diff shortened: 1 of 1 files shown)


subzero10:
@shalvah Poking you as well because I know you wanted to move to es6 but it seems that we have to do some more work to do that.

joshuap (approved):

shalvah (approved):

shalvah:
Well, that's a bummer, but, yep, we need to redo publishing/bundling/transpiling at some point. And it's not like we specifically did something wrong—it's just that the ecosystem is a hot mess.🥲

PR #869, opened 2022-09-09
feat: migrate honeybadger-webpack to monorepo

### LINKED ISSUE #868: Move honeybadger-webpack to the monorepo
I will start moving js packages into the monorepo and this is the first one.
With this package, I am composing a [guide](https://www.notion.so/honeybadger/549be6376a6246e19d6291bf8bdcee69?v=02c2f9dfbe41498ebd8f43880fb4e782&p=118e6bb566d34b3e92779038389da79b&pm=c) so that it can be applied on the rest.

subzero10:
## Status
**READY**

## Description
Fixes: #868

## Todos
- [x]  Cleanup - remove any devDeps that are only necessary to the monorepo (i.e. `commitlint`)
- [x]  Make sure that package runs all tests with `npm test`
- [x]  Make sure that package builds with `npm build`
- [x]  Modify release steps to use lerna version and publish
- [x]  Modify `[CHANGELOG.md](http://CHANGELOG.md)` skeleton to match other changelogs in the monorepo **OR** if `CHANGELOG.md` does not exist, commit an empty (which will be auto completed by lerna)
- [x]  Modify CI
- [x]  Add package to the monorepo `README.md`
- [x] Remove unnecessary information from package's `README.md` such as _licence_ and _contributing_ 
- [x]  Add badges for ci, version, download per month and total downloads in package’s `README.md`
- [x]  Create PR with commit type `feat` (should not be breaking - nothing should change for users)

## Todos after approval
- [x] Release new versions after PR is merged
- [x] Merge this [PR](https://github.com/honeybadger-io/honeybadger-webpack/pull/346)
- [x] Archive this [repository](https://github.com/honeybadger-io/honeybadger-webpack)
- [x] Move [open issues](https://github.com/honeybadger-io/honeybadger-webpack/issues) to this repository


packages/webpack/LICENSE
```diff
@@ -0,0 +1,23 @@
+Copyright (c) 2017 Honeybadger Industries LLC
```
subzero10, line 1: Can we delete this? Considering we have the LICENSE in the root folder of the monorepo?

packages/webpack/src/HoneybadgerSourceMapPlugin.js
```diff
@@ -0,0 +1,133 @@
+import async from 'async';
+import request from 'request';
+import VError from 'verror';
+import find from 'lodash.find';
+import reduce from 'lodash.reduce';
+import { handleError, validateOptions } from './helpers';
+import { ENDPOINT, PLUGIN_NAME } from './constants';
+
+export default class HoneybadgerSourceMapPlugin {
```
subzero10, line 9: I added `export default` here and now we have `module.exports` and `export default` in this file.
The reason is that our tests use `import`, but in docs we have instructions with `require()` and eslint complained that a default export is not found.
I tested both import and require and they work, so I left them both to ensure backwards compatibility. What do you think?

packages/webpack/LICENSE
shalvah: I'd guess so, but not sure.

shalvah: I'd guess so, but not sure.

shalvah (dismissed):

packages/webpack/src/HoneybadgerSourceMapPlugin.js
subzero10: The `honeybadger-webpack` repository uses [standard](https://github.com/standard/standard) and not `eslint` (even though there is an `.eslintignore` file 🤷).

Btw, even with eslint complaining, the tests pass and that confuses me 😅 

Here's how the compiled output looks:
```
exports.default = HoneybadgerSourceMapPlugin;
module.exports = HoneybadgerSourceMapPlugin;
```

Previously, it was only:
```
module.exports = HoneybadgerSourceMapPlugin;
```

packages/webpack/LICENSE
subzero10: My guess is that it may be necessary when publishing the package on npm 🤔.

packages/webpack/src/HoneybadgerSourceMapPlugin.js
shalvah: I see. Eh, I guess as can take it and worry about making it "proper" later.

shalvah: But on the flip side, that's a strange rule and we should probably look into it. I know Webpack only uses CommonJS, so `module.exports` is enough for any Webpack plugin. I think the integration between TypeScript and ESLint is off here (maybe that rule needs to be disabled for this file?)

packages/webpack/LICENSE
joshuap: Yeah, we want to make sure the license is included in each package (might want to also double-check the others).

joshuap (dismissed):

joshuap:
This will be a squash and merge, right? :)

packages/webpack/src/HoneybadgerSourceMapPlugin.js
subzero10: OK, I think I know why this is not working: Our eslint is configured globablly for all packages in the monorepo (for better consistency). It is configured with a Typescript parser, but this package does not have a tsconfig, nor typescript files. I'm pretty sure this is related to this problem.
For the moment, I have decided to just ignore the error like you suggested and will fix it by introducing Typescript in the package.

PR #887, opened 2022-09-16
fix: handle sync lambda handlers (non callback-based)

### LINKED ISSUE #860: Serverless.com - throws error for sync lambda
1. Create a serverless project
2. Create a simple synchronous handler that returns 200 (no callback, no async/await)
3. Wrap handler with Honeybadger.lambdaHandler
4. Do `sls deploy`

## What happens?
Honeybadger client throws error.

Should not throw error.

```
TypeError: handler(...).then is not a function
    at [NODE_MODULES]/@honeybadger-io/js/dist/server/honeybadger.js:1153:26
    at async_hooks.js:304:14
    at AsyncResource.runInAsyncScope (async_hooks.js:189:9)
    at AsyncLocalStorage.run (async_hooks.js:302:35)
    at [NODE_MODULES]/@honeybadger-io/js/dist/server/honeybadger.js:1149:40
    at new Promise (<anonymous>)
    at wrappedLambdaHandler ([NODE_MODULES]/@honeybadger-io/js/dist/server/honeybadger.js:1148:16)
    at Runtime.handler ([PROJECT_ROOT]/serverless_sdk/index.js:24:10716)
    at Runtime.handleOnce (/var/runtime/Runtime.js:66:25)
```

## Any other comments?
The issue goes away if I use async/await.

**Operating System:** AWS Lambda - nodejs v12
**Package Version:** v4.1.0
**Browser Version:** n/a

subzero10:
Fixes: #860

## Status
**READY**

## Description
Issue and fix discussed in detail [here](https://github.com/honeybadger-io/honeybadger-js/issues/860#issuecomment-1249508704).

### DIFF
--- a/packages/js/examples/aws-lambda/handler.js
+++ b/packages/js/examples/aws-lambda/handler.js
@@ -17,6 +17,13 @@ module.exports = {
       event,
     });
   },
+  helloWrapped: honeybadgerWrapper((event) => {
+    const hbKey = !!process.env.HONEYBADGER_API_KEY;
+    return formatJSONResponse({
+      message: `Hello, welcome to the exciting Serverless world! HB Key available: ${hbKey ? 'yes' : 'no'}. This handler is honeybadgerWrapped!`,
+      event,
+    });
+  }),
   syncError: honeybadgerWrapper(async (event) => {
     const willReport = event.body && event.body.report === 'yes';
     if (willReport) {
--- a/packages/js/examples/aws-lambda/package-lock.json
+++ b/packages/js/examples/aws-lambda/package-lock.json
@@ -1,8 +1,1270 @@
 {
   "name": "aws-lambda",
   "version": "1.0.0",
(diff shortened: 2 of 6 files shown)


joshuap (commented):
Looks reasonable to me. I'll let @shalvah review+approve. 👍

shalvah (approved):
Good catch.

Btw, I think we could fix this with a single-word change: adding an `await` before the `handler(...)` call (`await` automatically wraps the result in a Promise). But that would mean making the outer function async, which may have other side effects.

subzero10:
> Btw, I think we could fix this with a single-word change: adding an `await` before the `handler(...)` call (`await` automatically wraps the result in a Promise). But that would mean making the outer function async, which may have other side effects.

Thanks, I didn't know about this! Unfortunately, that's not that simple. The call to `handler` is inside the ALS callback which is not async. That's why I used a `Promise` class instead of `async`/`await`.

PR #891, opened 2022-09-17
feat: raise a repository dispatch event on new package releases

### LINKED ISSUE #882: Automatically update versions on honeybadger-io/honeybadger, honeybadger-io/docs and honeybadger-io/www repositories when CDN version changes
When the `js` package is released with a major or minor version update, we should change the update repositories to point to the updated version:
- honeybadger
- docs
- www

Possible solutions:
1. create issues on each repository after a successful major/minor release of the js package
2. trigger workflows on each that update links to point to the updated version of the major/minor release of the js package

subzero10:
## Status
**WIP**

## Description
Raise a repository dispatch event when a new js package is released.
Closes #882.

## TODO

- [x] Github Action to read package versions from all packages in monorepo
- [x] Repository Dispatch event to be consumed by target repositories
- [ ] Do not trigger event if no new package was released
- [ ] Allow calling multiple repositories (use [matrix](https://github.com/peter-evans/repository-dispatch#dispatch-to-multiple-repositories))
- [ ] Call reusable workflow from lerna-publish workflow
- [ ] Remove workflow trigger on `push`
- [ ] Replace `PAN_TEMP_TOKEN` with `LERNA_PUBLISH_TOKEN` (I used a personal token that has access to a personal repository)

### DIFF
--- a/.github/actions/get-package-versions/.gitignore
+++ b/.github/actions/get-package-versions/.gitignore
@@ -1,0 +1,2 @@
+node_modules
+!dist
--- a/.github/actions/get-package-versions/README.md
+++ b/.github/actions/get-package-versions/README.md
@@ -1,0 +1,33 @@
+# Get Monorepo Package Versions Github Action
+
+This action reads package versions from the monorepo and stores them as an array to a variable.
+
+## Outputs
+
+### `packages`
+
+A stringified array in the structure of:
+
+```json
+[
+  {
+    "name": "@honeybadger-io/core",
+    "version": "4.2.1"
+  },
+  {
+    "name": "@honeybadger-io/js",
+    "version": "4.3"
+  }
+]
+```
+
+## Example usage
+
+uses: .github/actions/get-package-versions
+
+## Developing
+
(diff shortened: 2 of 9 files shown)


subzero10:
@joshuap Here's a POC to raise an event that triggers a workflow in a separate repository. The event contains the current version of the js package. The target workflow (in the target repository) would then update any local files and create a PR.

What do you think? If you are OK with this, I can proceed to polish the logic (see TODO above) and then move on to the target workflows.

image
image

joshuap:
FWIW I think this decentralized approach is cool, but I like the centralized one better just because we have so many packages in different repositories, and it's nice not to spread the code around for such a small feature.

subzero10:
I agree! At first I got excited to try out these Github Actions features, but I quickly realized that I'd have to duplicate code in different repos. Plus, I'd have to update different sets (publisher and subscriber) of workflows in case of improvements/changes. That's a lot of maintenance. The other solution is centralized, all in one workflow and pretty easy to maintain.

PR #906, opened 2022-10-05
feat: rework store API

### LINKED ISSUE #905: Errors caught in middleware error handler cause process crash
1. Write code that configures HoneyBadger (v4.3.1), making sure not to disable breadcrumbs: 

```
Honeybadger.configure({
    environment,
    enableUncaught: true,
    enableUnhandledRejection: true,
    breadcrumbsEnabled: true,
    debug: false,
    revision: ver,
  });
```

3. Use the HB middleware as the last in the stack:
```
 app.use(Honeybadger.errorHandler);
``` 

## What happens?
Calling `Honeybadger.notify()` works as expected. 
However, when errors occur in Express controllers, there is an error in the HB client library that in turn causes an uncaught exception and hence process termination.

HB does not cause process to crash and successfully sends along breadcrumbs for errors caught in the middleware

```
[2022-10-03T18:30:38.701Z] [ERROR] Invalid token. {"request":{"path":"/api/jobs","method":"POST","query":{},"params":{},"startTime":"Mon Oct 03 2022 15:30:38 GMT-0300 (Chile Summer Time)"}}
[2022-10-03T18:30:38.707Z] [ERROR] TypeError: Cannot read property 'breadcrumbs' of undefined
    at Honeybadger.Client.addBreadcrumb (/our-app/node_modules/@honeybadger-io/js/dist/server/honeybadger.js:991:33)
    at Honeybadger.Client.notify (/our-app/node_modules/@honeybadger-io/js/dist/server/honeybadger.js:847:14)
    at Honeybadger.errorHandler (/our-app/node_modules/@honeybadger-io/js/dist/server/honeybadger.js:1447:10)
    at Layer.handle_error (/our-app/node_modules/express/lib/router/layer.js:71:5)
    at trim_prefix (/our-app/node_modules/express/lib/router/index.js:315:13)
    at/our-app/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/our-app/node_modules/express/lib/router/index.js:335:12)
    at next (/our-app/node_modules/express/lib/router/index.js:275:10)
    at Layer.handle_error (/our-app/node_modules/express/lib/router/layer.js:73:5)
    at trim_prefix (/our-app/node_modules/express/lib/router/index.js:315:13)
    at/our-app/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/our-app/node_modules/express/lib/router/index.js:335:12)
    at next (/our-app/node_modules/express/lib/router/index.js:275:10)
    at Layer.handle_error (/our-app/node_modules/express/lib/router/layer.js:67:12)
    at trim_prefix (/our-app/node_modules/express/lib/router/index.js:315:13)
    at/our-app/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/our-app/node_modules/express/lib/router/index.js:335:12)
    at next (/our-app/node_modules/express/lib/router/index.js:275:10)
    at Strategy.strategy.error (/our-app/node_modules/passport/lib/middleware/authenticate.js:343:9)
    at verified (/our-app/node_modules/passport-http-bearer/lib/strategy.js:118:28)
    at/our-app/config/authentication.js:79:5
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {"source":"console","stacktrace":"Stacktrace\n    at Function.logerror (/our-app/node_modules/express/lib/application.js:630:43)\n    at processImmediate (internal/timers.js:463:21)\n    at process.topLevelDomainCallback (domain.js:138:15)\n    at process.callbackTrampoline (internal/async_hooks.js:124:14)"}
/our-app/node_modules/@honeybadger-io/js/dist/server/honeybadger.js:991
        var breadcrumbs = store.breadcrumbs;
                                ^

TypeError: Cannot read property 'breadcrumbs' of undefined
    at Honeybadger.Client.addBreadcrumb (/our-app/node_modules/@honeybadger-io/js/dist/server/honeybadger.js:991:33)
    at Honeybadger.Client.notify (/our-app/node_modules/@honeybadger-io/js/dist/server/honeybadger.js:847:14)
    at process.honeybadgerUncaughtExceptionListener (/our-app/node_modules/@honeybadger-io/js/dist/server/honeybadger.js:1383:28)
    at process.emit (events.js:326:22)
    at process.EventEmitter.emit (domain.js:483:12)
    at process._fatalException (internal/process/execution.js:165:25)

Process finished with exit code 7

``` 

## Any other comments?
We worked around the issue for now turning off breadcrumbs:
```
Honeybadger.configure({
    ...
    breadcrumbsEnabled: false,
    ...
  });
```

**Operating System:** …
MacOS (local env) and Ubuntu 20.04 (Server)

**Package Name:** …
**Package Version:** …
4.3.1

Front logo [Front conversations](https://app.frontapp.com/open/top_c9rkx)

shalvah:
Rework of https://github.com/honeybadger-io/honeybadger-js/pull/845, which should probably fix #905.

Going with @subzero10 's suggestion (thanks!), I've reworked the store API.
- No more singletons
- The store API is no longer just an extension of/workaround for AsyncLocalStorage, but an expression of where the Honeybadger client stores context and breadcrumbs. This means there's now a full-fledged Store interface:

```ts
interface HoneybadgerStore {
  available(): boolean
  getContents(): StoreContents
  getContents<K extends keyof StoreContents>(key: K): StoreContents[K]
  setContext(context: Record<string, unknown>): void
  addBreadcrumb(breadcrumb: BreadcrumbRecord): void
  clear(): void
  run<R>(callback: () => R, ...opts: unknown[]): R
}
```
This gives us better encapsulation. Previously, the client would manually mutate the store's contents; now, it can just call`addBreadcrumb()`, `setContext()`, etc, and the store can figure out what to do. It also allows us to get rid of methods like `__setStore()` and `__getStoreContentsOrDefault()`, which, tbh, were hacks. Plus we can avoid mutability traps by having the store return copies to the client.

- The `core` package includes only the GlobalStore, as before.
- The `js` package comes with two extra stores, `AsyncStore` and `StackedStore`. `AsyncStore` retains the old ALS approach, while `StackedStore` is a wrapper that combines the two stores, using ALS if available, otherwise falling back to global store.

(Oh, there is one singleton—the `contents` object is shared by all three stores. I left it this way so that we don't need to manually "combine" stores. You can add stuff to the global store and then when you enter an async context, it is initialised with a copy of the AsyncStore's `contents`, which is the same as the global one's. Might change it later if it leads to problems, but I found it cleaner this way, as the async store doesn't need to know about the existence of a global store.)

### DIFF
--- a/packages/core/src/store.ts
+++ b/packages/core/src/store.ts
@@ -1,18 +1,45 @@
-import { HoneybadgerStore } from './types';
+import { HoneybadgerStore, StoreContents } from './types';
+import { merge } from './util';
 
-export class GlobalStore<T> implements HoneybadgerStore<T> {
-  private store: T
+export class GlobalStore implements HoneybadgerStore {
+  private readonly contents: StoreContents;
+  private readonly breadcrumbsLimit: number;
 
-  constructor(store: T) {
-    this.store = store;
+  constructor(contents: StoreContents, breadcrumbsLimit: number) {
+    this.contents = contents;
+    this.breadcrumbsLimit = breadcrumbsLimit;
   }
 
-  getStore(): T {
-    return this.store
+  static create(contents: StoreContents, breadcrumbsLimit: number) {
+    return new GlobalStore(contents, breadcrumbsLimit);
   }
 
-  run<R, TArgs extends unknown[]>(store: T, callback: (...args: TArgs) => R, ...args: TArgs): R {
-    this.store = store;
-    return callback(...args);
+  available() {
+    return true
+  }
+
+  getContents(key?: keyof StoreContents) {
+    const value = key ? this.contents[key] : this.contents;
+    return JSON.parse(JSON.stringify(value));
+  }
+
+  setContext(context) {
+    this.contents.context = merge(this.contents.context, context || {});
+  }
+
+  addBreadcrumb(breadcrumb) {
+    if (this.contents.breadcrumbs.length == this.breadcrumbsLimit) {
+      this.contents.breadcrumbs.shift()
+    }
+    this.contents.breadcrumbs.push(breadcrumb);
+  }
+
+  clear() {
+    this.contents.context = {}
+    this.contents.breadcrumbs = []
+  }
+
+  run(callback) {
+    return callback();
   }
 }
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1,5 +1,4 @@
 import { Client } from './client'
-import type { AsyncLocalStorage } from 'async_hooks';
 
 export interface Logger {
   log(...args: unknown[]): unknown
@@ -163,5 +162,29 @@ export interface Transport {
   send(options: TransportOptions, payload?: NoticeTransportPayload | undefined): Promise<{ statusCode: number; body: string; }>
 }
 
-export type HoneybadgerStore<T> = Pick<AsyncLocalStorage<T>, 'getStore' | 'run'>
-export type DefaultStoreContents = {context: Record<string, unknown>, breadcrumbs: BreadcrumbRecord[]}
+export interface HoneybadgerStore {
+  /**
+   * Is the store available for writing and reading?
+   */
+  available(): boolean
+
+  /**
+   * With no arguments, returns a copy of the store contents for reading.
+   * When a key is supplied, returns the value of that key in the store.
+   */
+  getContents(): StoreContents
+  getContents<K extends keyof StoreContents>(key: K): StoreContents[K]
+
+  setContext(context: Record<string, unknown>): void
+
+  addBreadcrumb(breadcrumb: BreadcrumbRecord): void
+
+  clear(): void
+
+  run<R>(callback: () => R, ...opts: unknown[]): R
+}
+
+export type StoreContents = {
+  context: Record<string, unknown>,
+  breadcrumbs: BreadcrumbRecord[]
+}
--- a/packages/js/src/server/async_store.ts
+++ b/packages/js/src/server/async_store.ts
@@ -1,13 +1,86 @@
-import { Types, Store as StoreModule } from '@honeybadger-io/core'
+import { Types } from '@honeybadger-io/core';
+import { AsyncLocalStorage } from 'async_hooks';
 
-let Store: Types.HoneybadgerStore<{ context: Record<string, unknown>; breadcrumbs: Types.BreadcrumbRecord[]; }>
-try {
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const { AsyncLocalStorage } = require('async_hooks')
-  Store = new AsyncLocalStorage()
-}
-catch (e) {
-  Store = new StoreModule.GlobalStore({ context: {}, breadcrumbs: [] })
-}
+const kHoneybadgerStore = Symbol.for('kHoneybadgerStore');
 
-export const AsyncStore = Store
+export class AsyncStore implements Types.HoneybadgerStore {
+  private als: AsyncLocalStorage<Types.StoreContents>;
+  private readonly contents: Types.StoreContents;
+  private readonly breadcrumbsLimit: number;
+
+  constructor(
+    asyncLocalStorage: AsyncLocalStorage<Types.StoreContents>,
+    contents: Types.StoreContents, breadcrumbsLimit: number
+  ) {
+    this.als = asyncLocalStorage;
+    this.contents = contents;
+    this.breadcrumbsLimit = breadcrumbsLimit;
+  }
+
+  /**
+   * Attempt to create a new AsyncStore instance
+   */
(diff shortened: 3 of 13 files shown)


joshuap (approved):
I read the code and this looks good to me. If users are waiting on it, I'm OK with merging now as long as you're around for bugfixes. Pan is on vacation right now, but might give a more thorough review if you need/want one.

shalvah:
Okay, do you know when he's coming back? Because we'll need to release this as well.

packages/core/src/store.ts
```diff
@@ -1,18 +1,45 @@
...
 
-export class GlobalStore<T> implements HoneybadgerStore<T> {
-  private store: T
+export class GlobalStore implements HoneybadgerStore {
+  private readonly contents: StoreContents;
+  private readonly breadcrumbsLimit: number;
 
-  constructor(store: T) {
-    this.store = store;
+  constructor(contents: StoreContents, breadcrumbsLimit: number) {
```
subzero10, line 8: Since we have a `static create` method, I think this constructor should be private. Otherwise, anyone can ignore can work around the `create` method and instantiate a `GlobalStore` with the constructor.

packages/core/src/types.ts
```diff
@@ -163,5 +162,29 @@ export interface Transport {
   send(options: TransportOptions, payload?: NoticeTransportPayload | undefined): Promise<{ statusCode: number; body: string; }>
 }
 
-export type HoneybadgerStore<T> = Pick<AsyncLocalStorage<T>, 'getStore' | 'run'>
-export type DefaultStoreContents = {context: Record<string, unknown>, breadcrumbs: BreadcrumbRecord[]}
+export interface HoneybadgerStore {
```
subzero10, line 165: ❤️

packages/js/src/server/async_store.ts
```diff
@@ -1,13 +1,86 @@
...
-  Store = new StoreModule.GlobalStore({ context: {}, breadcrumbs: [] })
-}
+const kHoneybadgerStore = Symbol.for('kHoneybadgerStore');
+
+export class AsyncStore implements Types.HoneybadgerStore {
+  private als: AsyncLocalStorage<Types.StoreContents>;
+  private readonly contents: Types.StoreContents;
+  private readonly breadcrumbsLimit: number;
+
+  constructor(
```
subzero10, line 11: As mentioned above, I think this should be `private`.

packages/js/src/server/async_store.ts
```diff
@@ -1,13 +1,86 @@
...
+    // This allows every request to start from the same state
+    // and add data specific to them, without overlapping.
+
+    if (!request) {
+      return this.als.run(this.getContents(), callback);
+    }
+
+    let contents: Types.StoreContents;
+    if (request[kHoneybadgerStore]) {
+      contents = request[kHoneybadgerStore] as Types.StoreContents
```
subzero10, line 78: Can you give an example where this could happen?
We call `run` for each request only once, right? If yes, why would we ever reach to L78?

subzero10 (commented):
This is great! That's exactly what I wanted to do, but I wasn't sure of the implementation details. I think you nailed it!
I have some questions/comments :)

packages/js/src/server/async_store.ts
shalvah: In the error handling middleware.

```js
// Honeybadger.requestHandler
app.use((req, res, next) => {
  hb.withRequest(next);
});

// User's request handler
app.get('/', (req, res, next) => {
  // ...
});

// Honeybadger.errorHandler
app.use((err, req, res, next) => {
  hb.withRequest(req, () => { hb.notify(err); next(err); });
});
```

packages/js/src/server/async_store.ts
```diff
@@ -1,13 +1,86 @@
-import { Types, Store as StoreModule } from '@honeybadger-io/core'
+import { Types } from '@honeybadger-io/core';
+import { AsyncLocalStorage } from 'async_hooks';
```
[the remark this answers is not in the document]
shalvah: Right, good catch!

packages/core/src/store.ts
shalvah: Hmmm. Honestly, the `create` was added as more of a convenience than a restriction. I made a `create` method so we could "safely" instantiate the AsyncStore (return null if ALS not supported). Then I added it to the global store for consistency. We could make them private, but I don't care too much either way.

packages/js/src/server/async_store.ts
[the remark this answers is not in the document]
shalvah: Fixed in my other PR.

packages/core/src/store.ts
subzero10: OK, I guess this is fine since this class is really not used directly outside the package. If at some point we decide to do more in the `create` method before returning an instance, we could then consider making the constructor private.

Issue #907, opened 2022-10-11
@types/express is required by server middleware 


anonyo:
## What are the steps to reproduce this issue?

1. cd into https://github.com/honeybadger-io/honeybadger-js/tree/master/packages/js/examples/aws-lambda
2. yarn install
3. sls offline start

## What happens?
```
➜  aws-lambda-typescript git:(master) ✗ sls offline start

Compiling with Typescript...
Using local tsconfig.json
/Users/some_user/src/honeybadger-js/packages/js/examples/aws-lambda-typescript/node_modules/@honeybadger-io/js/dist/server/middleware.d.ts (1,49): Cannot find module 'express' or its corresponding type declarations.
Typescript compiled.
Watching typescript files...
Starting Offline: dev/us-east-1.
```

Also happening on version 
```
    "@honeybadger-io/js": "^4.3.1",
```

## What were you expecting to happen?

- Shouldn't have to add `@types/express`. 
- Be able to support applications without express. 

## Any logs, error output, etc?

```
/Users/some_user/src/honeybadger-js/packages/js/examples/aws-lambda-typescript/node_modules/@honeybadger-io/js/dist/server/middleware.d.ts (1,49): Cannot find module 'express' or its corresponding type declarations.
```

## What versions are you using?
**Operating System:** MacOS 12.6
**Package Name:** @honeybadger-io/js
**Package Version:** 4.3.1
**Browser Version:** Version 106.0.5249.103 (Official Build) (x86_64)

## Any other comments?

Is there a way to ommit `dist/server/middleware.d.ts `?

Thanks in advance!

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_cnvmp)

joshuap:
Thanks @anonyo! We're working on this one, but one of our maintainers is on vacation so it might be a few weeks. In the meantime, I think you can fix this error by adding the `@types/express` package:

```
// npm:
npm add @types/express --save

// yarn:
yarn add @types/express
```

subzero10:
@anonyo This should be fixed with [v4.6.2](https://github.com/honeybadger-io/honeybadger-js/releases/tag/v4.6.2)!

PR #910, opened 2022-10-13
fix: remove unneeded dependencies

shalvah:
Removes our dependency on `@types/express`, fixing #907. (Also removed `@types/aws_lambda`.( Inlined the types directly; we don't really need the in-depth type info, since we aren't the end user.

### DIFF
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -50,8 +50,7 @@
     "postpublish": "./scripts/release-cdn.sh"
   },
   "dependencies": {
-    "@honeybadger-io/core": "^4.3.0",
-    "@types/aws-lambda": "^8.10.89"
+    "@honeybadger-io/core": "^4.3.0"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


joshuap (approved):
LGTM

subzero10:
I'm being completely honest here, please don't hate me :)

This feels like a solution that will come to bite us in the future.
- We are going down a path where we will have to manually mirror the work of the open-source community. 
- We are losing any type-checking that we could have done with dependabot and our CI checks.

I also replied on [Slack](https://honeybadger.slack.com/archives/C7ET3UGD6/p1666057005606229?thread_ts=1665423918.943069&cid=C7ET3UGD6) with more detailed thoughts.

joshuap:
I'll defer to @subzero10 on this! @shalvah let's wait to discuss further until after Pan's back from parental leave. In the meantime I think that we can just tell users to add the type dependencies if they need them, right?

PR #918, opened 2022-10-28
build!: update to lerna 6

subzero10:
## Status
**READY**

## Description
The automated update from dependabot is failing so I'm trying it manually.
Lerna 6 brings performance improvements.


subzero10:
CI is failing with node v12. I didn't see anything in Lerna's changelog about removing support. I created an [issue](https://github.com/lerna/lerna/issues/3394) in the lerna repository.

subzero10:
So, I looked a bit deeper in Lerna's changelog and it seems they [dropped support for node v10 and v12](https://github.com/lerna/lerna/releases/tag/v5.0.0) in v5. It looks though that lerna really stopped working with node v12 at v6.

@joshuap What's our convention/policy about node version support? node v12 reached [end of life 6 months ago](https://endoflife.date/nodejs).
FYI lerna v6 upgrade can be delayed, but we'll probably face issues like this one with more packages pretty soon.

joshuap:
> So, I looked a bit deeper in Lerna's changelog and it seems they [dropped support for node v10 and v12](https://github.com/lerna/lerna/releases/tag/v5.0.0) in v5. It looks though that lerna really stopped working with node v12 at v6.
> 
> @joshuap What's our convention/policy about node version support? node v12 reached [end of life 6 months ago](https://endoflife.date/nodejs). FYI lerna v6 upgrade can be delayed, but we'll probably face issues like this one with more packages pretty soon.

Our CI should build on officially maintained/supported versions. We can drop for support for anything that is EOL.

subzero10:
@joshuap @shalvah I marked this as breaking change (major version bump) because our CI won't test against node v12 anymore.

This requirement comes from Lerna v6, a monorepo dependency, **not** a dependency to any of our packages. This means the packages will still work with node v12, at least for now.

joshuap (commented):
The changes look good to me, but I'm not sure about the version bump—do we really need to bump the major version just because we're dropping an old version in CI?

subzero10:
> The changes look good to me, but I'm not sure about the version bump—do we really need to bump the major version just because we're dropping an old version in CI?

Good point. I'm not sure to be honest, but here's the argument for it:
- We are going to stop testing on Node v12.
- Any new code introduced will not be tested against Node v12.
- When some change breaks on Node v12, we won't know about it until we get an issue reported (but we don't know if that's going to happen, considering Node v12 has already reached EOL).

I guess we could decide not to do a major version bump, and accept that there is some risk involved for future releases (this release will still work with Node v12).

joshuap:
I could go either way—I'm fine doing the major bump if you're more comfortable with that. Maybe @shalvah could weigh in on this as well.

shalvah:
I'm also fine either way, but is there a chance we're introducing other breaking changes soon? If so, I'd hold off the major until that.

subzero10:
> I'm also fine either way, but is there a chance we're introducing other breaking changes soon? If so, I'd hold off the major until that.

Done! We now have 2 changes for the next major release 😉

subzero10:
See #959.

PR #945, opened 2022-11-12
feat: support non-error type "error.cause"

### LINKED ISSUE #934: [feat] Support non-error cause property

Throw an error with data inside `cause` property.

```
throw new Error('Name', { cause: { code: 'NonInteger' } })
```

## What happens?

Error `cause` is not visible on HoneyBadger web UI.

Error `cause` to be visible on Honeybadger web UI.

## Any other comments?

> The value can be of any type. You should not make assumptions that the error you caught has an Error as its cause

From [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause).

**Operating System:** macOS
**Package Name:** @honeybadger-io/js
**Package Version:** 4.6.0
**Browser Version:** Chrome 107.0.0.0

subzero10:
## Status
**READY**

Closes: #934

## Description
Support non-error type `cause`.

### DIFF
--- a/packages/js/test/unit/server/middleware.server.test.ts
+++ b/packages/js/test/unit/server/middleware.server.test.ts
@@ -100,7 +100,7 @@ describe('Express Middleware', function () {
       const initialContext = client.__getContext();
       client.setContext({ reqId: req.params.reqId });
       setTimeout(() => {
-        res.json({
+        res.status(200).json({
           initial: initialContext,
           final: client.__getContext()
         });
@@ -137,7 +137,7 @@ describe('Express Middleware', function () {
     app.use(client.errorHandler)
 
     app.use(function(_err, _req, res, _next) {
-      res.json(client.__getContext(), 500);
+      res.status(500).json(client.__getContext());
     })
 
     return Promise.all([80, 90].map((i) => {
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -377,7 +377,7 @@ export abstract class Client {
         backtrace: notice.backtrace,
         fingerprint: notice.fingerprint,
         tags: notice.tags,
-        causes: getCauses(notice),
+        causes: getCauses(notice).map(c => filter(c, this.config.filters)),
       },
       request: {
         url: filterUrl(notice.url, this.config.filters),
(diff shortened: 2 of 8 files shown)


packages/js/test/unit/server/middleware.server.test.ts
```diff
@@ -100,7 +100,7 @@ describe('Express Middleware', function () {
       const initialContext = client.__getContext();
       client.setContext({ reqId: req.params.reqId });
       setTimeout(() => {
-        res.json({
+        res.status(200).json({
```
subzero10, line 103: `res.json` is deprecated, it should be `res.status().json()`

subzero10:
@joshuap I modified the js library to send non-error type `cause` to the backend, but it seems that the Honeybadger web interface expects to render an Error object.

Here's how it looks when the `cause` object has `class`, `message` and `backtrace` properties:
Screenshot 2022-11-12 at 5 56 31 PM

And here's how it looks when the `cause` object is missing those properties:
Screenshot 2022-11-12 at 6 06 01 PM
Screenshot 2022-11-12 at 6 05 53 PM

joshuap (changes requested):
> @joshuap I modified the js library to send non-error type `cause` to the backend, but it seems that the Honeybadger web interface expects to render an Error object.
> 
> Here's how it looks when the `cause` object has `class`, `message` and `backtrace` properties: Screenshot 2022-11-12 at 5 56 31 PM
> 
> And here's how it looks when the `cause` object is missing those properties: Screenshot 2022-11-12 at 6 06 01 PM Screenshot 2022-11-12 at 6 05 53 PM

Edit: I'm still catching up on this, obviously. Let's chat about implementation on Slack—maybe we can figure out a way to also support the structured data use case. :)

joshuap:
Sorry, there should have been extra clarification on #934 before starting on it.

joshuap:
You could also ask if context would solve the use case in #934. If so, we could always skip this feature for now.

shalvah (approved):

subzero10:
Closing because of inactivity. We may come back to this again.

Issue #946, opened 2022-11-14
`enableUnhandledRejection` silently swallows errors instead of exiting in nodejs


thatsmydoing:
## What are the steps to reproduce this issue?
```
const Honeybadger = require('@honeybadger-io/js');

Honeybadger.configure({
  ...
});

async function main() {
  throw new Error('AHHHH');
}

main();
```

## What happens?
The program exits successfully.

## What were you expecting to happen?
The program should exit with an error because of an unhandled promise rejection which is the standard behavior in nodejs.

## What versions are you using?
**Operating System:** Linux
**Package Name:** @honeybadger-io/js
**Package Version:** 4.7.1
**Browser Version:** n/a

subzero10:
@thatsmydoing This should be fixed with `v4.7.4`!

thatsmydoing:
Thanks! I can confirm it's fixed.

PR #950, opened 2022-11-17
fix: exit process on unhandledRejection

subzero10:
## Status
**READY**

## Description
Fixes: #946.
I wanted to write tests but it seems that jest will exit the process without calling `process.on('unhandledRejection')`.
So, I tested manually with an example project.

## Bonus
I applied similar logic to `unhandledException`s for nodejs.

### DIFF
--- a/packages/js/rollup.config.js
+++ b/packages/js/rollup.config.js
@@ -45,7 +45,8 @@ export default [
     external: ['http', 'https', 'url', 'os', 'fs', 'util', 'domain', 'async_hooks'],
     output: {
       file: pkg.main,
-      format: 'cjs'
+      format: 'cjs',
+      sourcemap: true
     },
     plugins: sharedPlugins
   }
--- a/packages/js/src/server/integrations/uncaught_exception.ts
+++ b/packages/js/src/server/integrations/uncaught_exception.ts
@@ -14,28 +14,53 @@ function removeAwsLambdaListener() {
   removeAwsDefaultUncaughtExceptionListener()
 }
 
+/**
+ * If there are no other uncaughtException listeners,
+ * we want to report the exception to Honeybadger and
+ * mimic the default behavior of NodeJs,
+ * which is to exit the process with code 1
+ */
+function hasUncaughtExceptionListeners() {
+  return process.listeners('uncaughtException').length !== 0
+}
+
 export default function (): Types.Plugin {
   return {
     load: (client: typeof Client) => {
-      if (!client.config.enableUncaught) { return }
+      if (!client.config.enableUncaught) {
+        return
+      }
 
       removeAwsLambdaListener()
 
+      const hasOtherListeners = hasUncaughtExceptionListeners();
       process.on('uncaughtException', function honeybadgerUncaughtExceptionListener(uncaughtError) {
-        // Prevent recursive errors
-        if (count > 1) { fatallyLogAndExit(uncaughtError) }
+        if (!client.config.enableUncaught) {
+          client.config.afterUncaught(uncaughtError)
+          if (!hasOtherListeners) {
+            fatallyLogAndExit(uncaughtError)
+          }
+          return
+        }
 
-        if (client.config.enableUncaught) {
+        // report only the first error - prevent reporting recursive errors
+        if (count < 1) {
           client.notify(uncaughtError, {
             afterNotify: (_err, _notice) => {
-              count += 1
+              count++
               client.config.afterUncaught(uncaughtError)
+              if (!hasOtherListeners) {
+                fatallyLogAndExit(uncaughtError)
+              }
             }
           })
-        } else {
-          count += 1
-          client.config.afterUncaught(uncaughtError)
         }
+        else {
+          if (!hasOtherListeners) {
+            fatallyLogAndExit(uncaughtError)
+          }
+        }
+
       })
     }
   }
--- a/packages/js/src/server/integrations/uncaught_exception.ts
+++ b/packages/js/src/server/integrations/uncaught_exception.ts
@@ -4,7 +4,7 @@ import Client from '../../server'
 import { removeAwsDefaultUncaughtExceptionListener } from '../aws_lambda'
 
 let isReporting = false
-let reportedOnce = false
+let handlerAlreadyCalled = false
 
 function removeAwsLambdaListener() {
   const isLambda = !!process.env.LAMBDA_TASK_ROOT
@@ -45,24 +45,28 @@ export default function (): Types.Plugin {
         }
 
         // report only the first error - prevent reporting recursive errors
-        if (!reportedOnce) {
-          isReporting = true
-          client.notify(uncaughtError, {
-            afterNotify: (_err, _notice) => {
-              isReporting = false
-              reportedOnce = true
-              client.config.afterUncaught(uncaughtError)
-              if (!hasOtherUncaughtExceptionListeners()) {
-                fatallyLogAndExit(uncaughtError)
-              }
-            }
-          })
-        }
-        else {
-          if (!hasOtherUncaughtExceptionListeners() && !isReporting) {
+        if (handlerAlreadyCalled) {
+          if (!hasOtherUncaughtExceptionListeners()) {
             fatallyLogAndExit(uncaughtError)
           }
+          return
         }
+
+        if (isReporting) {
+          return
+        }
+
+        isReporting = true
+        client.notify(uncaughtError, {
+          afterNotify: (_err, _notice) => {
+            isReporting = false
+            handlerAlreadyCalled = true
+            client.config.afterUncaught(uncaughtError)
+            if (!hasOtherUncaughtExceptionListeners()) {
+              fatallyLogAndExit(uncaughtError)
+            }
+          }
+        })
       })
     }
   }
--- a/packages/js/src/server/integrations/uncaught_exception.ts
+++ b/packages/js/src/server/integrations/uncaught_exception.ts
@@ -4,7 +4,7 @@ import Client from '../../server'
 import { removeAwsDefaultUncaughtExceptionListener } from '../aws_lambda'
 
 let isReporting = false
-let count = 0
+let reportedOnce = false
 
 function removeAwsLambdaListener() {
   const isLambda = !!process.env.LAMBDA_TASK_ROOT
@@ -21,8 +21,8 @@ function removeAwsLambdaListener() {
  * mimic the default behavior of NodeJs,
  * which is to exit the process with code 1
  */
-function hasUncaughtExceptionListeners() {
-  return process.listeners('uncaughtException').length !== 0
+function hasOtherUncaughtExceptionListeners() {
+  return process.listeners('uncaughtException').length > 1
 }
 
 export default function (): Types.Plugin {
@@ -34,32 +34,32 @@ export default function (): Types.Plugin {
 
       removeAwsLambdaListener()
 
-      const hasOtherListeners = hasUncaughtExceptionListeners();
+
       process.on('uncaughtException', function honeybadgerUncaughtExceptionListener(uncaughtError) {
         if (!client.config.enableUncaught) {
           client.config.afterUncaught(uncaughtError)
-          if (!hasOtherListeners) {
+          if (!hasOtherUncaughtExceptionListeners()) {
             fatallyLogAndExit(uncaughtError)
           }
           return
         }
 
         // report only the first error - prevent reporting recursive errors
-        if (count < 1) {
+        if (!reportedOnce) {
           isReporting = true
           client.notify(uncaughtError, {
             afterNotify: (_err, _notice) => {
               isReporting = false
-              count++
+              reportedOnce = true
               client.config.afterUncaught(uncaughtError)
-              if (!hasOtherListeners) {
+              if (!hasOtherUncaughtExceptionListeners()) {
                 fatallyLogAndExit(uncaughtError)
               }
             }
           })
         }
         else {
-          if (!hasOtherListeners && !isReporting) {
+          if (!hasOtherUncaughtExceptionListeners() && !isReporting) {
             fatallyLogAndExit(uncaughtError)
           }
         }
--- a/packages/js/src/server/integrations/uncaught_exception.ts
+++ b/packages/js/src/server/integrations/uncaught_exception.ts
@@ -3,6 +3,7 @@ import { fatallyLogAndExit } from '../util'
 import Client from '../../server'
 import { removeAwsDefaultUncaughtExceptionListener } from '../aws_lambda'
 
+let isReporting = false
 let count = 0
 
 function removeAwsLambdaListener() {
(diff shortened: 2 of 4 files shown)


shalvah:
Yes, we should have the same behaviour for both handlers.

subzero10:
@joshuap @shalvah Please review the following changes in logic:
- The default behavior in Nodejs documentation for `unhandledRejection` and `unhandledException` is to exit the process with code 1.
- So, ideally the js client should respect this behavior by reporting the error to Honeybadger and exiting the process.
- However, again per the nodejs documentation, if there is at least one listener for any of these events, the default behavior changes and the process does not exit with code 1.

I modified the code to do the above. Let me know what you think.

packages/js/src/server/integrations/uncaught_exception.ts
```diff
@@ -14,27 +15,53 @@ function removeAwsLambdaListener() {
...
+                fatallyLogAndExit(uncaughtError)
+              }
             }
           })
-        } else {
-          count += 1
-          client.config.afterUncaught(uncaughtError)
+        }
+        else {
+          if (!hasOtherListeners && !isReporting) {
```
subzero10, line 62: Don't exit the process if we are in the middle of sending the notice to Honeybadger.

subzero10:
Finally, we should decide whether we should consider this a breaking change and bump the major version (i.e. release v5). Even though it's a bug fix, it may produce unexpected behavior for some people.

joshuap (commented):
This looks good here. I will let the others review/approve.

Regarding the breaking change, it feels to me like this is more of a bug fix, since it's only changing what happens when the process exits—and it's fixing that behavior, right?

joshuap:
> @joshuap should we do the same for `uncaughtException` as well? According to [Nodejs docs](https://nodejs.org/api/process.html#event-uncaughtexception), the default behaviour is to exit the process with error code 1:
> 
> > By default, Node.js handles such exceptions by printing the stack trace to stderr and exiting with code 1, overriding any previously set [process.exitCode](https://nodejs.org/api/process.html#processexitcode_1). Adding a handler for the 'uncaughtException' event overrides this default behavior. Alternatively, change the [process.exitCode](https://nodejs.org/api/process.html#processexitcode_1) in the 'uncaughtException' handler which will result in the process exiting with the provided exit code. Otherwise, in the presence of such handler the process will exit with 0.

As far as I recall, the `uncaughtException` handler *should* behave the same as the default, with the addition of reporting the error (unless it's replacing another custom handler). You might want to look into that to confirm.

shalvah:
Yes, I don't see it as a bug fix, but I wouldn't make it a new major either. Maybe a minor.

packages/js/rollup.config.js
```diff
@@ -45,7 +45,8 @@ export default [
     external: ['http', 'https', 'url', 'os', 'fs', 'util', 'domain', 'async_hooks'],
     output: {
       file: pkg.main,
-      format: 'cjs'
+      format: 'cjs',
+      sourcemap: true
```
shalvah, line 49: Why this?

packages/js/src/server/integrations/uncaught_exception.ts
```diff
@@ -14,27 +15,53 @@ function removeAwsLambdaListener() {
...
 
-        if (client.config.enableUncaught) {
+        // report only the first error - prevent reporting recursive errors
+        if (count < 1) {
+          isReporting = true
           client.notify(uncaughtError, {
             afterNotify: (_err, _notice) => {
-              count += 1
+              isReporting = false
+              count++
```
shalvah, line 53: What is this the count of? I know this variable wasn't introduced by your PR, but can we rename this so it's clearer? The poor name makes it more difficult to follow.

packages/js/src/server/integrations/uncaught_exception.ts
```diff
@@ -14,27 +15,53 @@ function removeAwsLambdaListener() {
...
   return {
     load: (client: typeof Client) => {
-      if (!client.config.enableUncaught) { return }
+      if (!client.config.enableUncaught) {
+        return
+      }
 
       removeAwsLambdaListener()
 
+      const hasOtherListeners = hasUncaughtExceptionListeners();
```
shalvah, line 37: Isn't there a slight bug here? If you only check for other listeners when this module is loaded, what happens if a listener is registered after ours? I think there `hasUncaughtExceptionListeners` should be called on demand. In which case, it should check if the length > 1 and probably be renamed to `hasOtherUncaughtExceptionListeners`.

subzero10: True. I would expect that this kind of setup happens quite early in any app. But you are right, I will improve! Thanks!

packages/js/src/server/integrations/uncaught_exception.ts
subzero10: I think we want to send only the first exception that lands in this handler, probably covering the scenario of an infinite loop?

But as you said, it's not my code, so it's just an assumption.

packages/js/rollup.config.js
subzero10: No. I put it here to help me with debugging. I don't remember if it did, but there's no harm letting it there.

packages/js/src/server/integrations/uncaught_exception.ts
subzero10: Done!

packages/js/src/server/integrations/unhandled_rejection.ts
```diff
@@ -19,18 +19,17 @@ export default function (): Types.Plugin {
         return
       }
 
-      const hasOtherListeners = hasUnhandledRejectionListeners();
       process.on('unhandledRejection', function honeybadgerUnhandledRejectionListener(reason, _promise) {
         if (!client.config.enableUnhandledRejection) {
-          if (!hasOtherListeners) {
+          if (!hasOtherUnhandledRejectionListeners()) {
```
shalvah, line 24: Whoops, sorry, another bug. This is being called inside `if (!client.config.enableUnhandledRejection) {`, which means checking that `listeners.length > 1` might not be accurate. I'm not sure why we even register a handler when that is turned off. Isn't that counter to user's expectations? IMO, we should invert the conditions on the previous lines, leading to:

```js
      if (!client.config.enableUnhandledRejection) {
        process.on('unhandledRejection', function honeybadgerUnhandledRejectionListener(reason, _promise) {
          if (!hasOtherUnhandledRejectionListeners()) {
```

Also not your code, but definitely something I think we can do better (and maybe might be worth packaging it into the major at this point.)

packages/js/src/server/integrations/uncaught_exception.ts
```diff
@@ -34,32 +34,32 @@ export default function (): Types.Plugin {
...
 
         // report only the first error - prevent reporting recursive errors
-        if (count < 1) {
+        if (!reportedOnce) {
           isReporting = true
           client.notify(uncaughtError, {
             afterNotify: (_err, _notice) => {
               isReporting = false
-              count++
+              reportedOnce = true
```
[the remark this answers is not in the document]
subzero10: Good points!

The early return is a bit tricky though, I tried it and it's not as clean as you show it above:
- we need to handle the case where `handlerAlreadyCalled` is `false` but we don't want to report to HB because `isReporting` is `true`

But I may have a solution with multiple early returns. Incoming commit!

packages/js/src/server/integrations/unhandled_rejection.ts
subzero10: > This is being called inside if (!client.config.enableUnhandledRejection) {, which means checking that listeners.length > 1 might not be accurate

Why is it not accurate? It's inside our registered handler, which means we know there's at least one registered handler (this one).

By the way, I had the same question when I first read this code. Here's how I understand it (assuming this is intended behaviour):

First, on HB init, check if unhandled rejections should be reported to HB.
- If not, return.
- If yes, proceed to register handler:
  - Handler called, **check if config is still enabled. Why? Because we allow modifying the config anytime (`Honeybadger.configure`)**
    - If not enabled, check if there are any other handlers before deciding whether to exit or not.
    - If enabled, proceed to report to HB.

packages/js/src/server/integrations/uncaught_exception.ts
```diff
@@ -14,28 +15,58 @@ function removeAwsLambdaListener() {
...
+        if (handlerAlreadyCalled) {
+          if (!hasOtherUncaughtExceptionListeners()) {
+            fatallyLogAndExit(uncaughtError)
+          }
+          return
+        }
+
+        if (isReporting) {
+          return
         }
```
subzero10, line 57: @shalvah

packages/js/src/server/integrations/unhandled_rejection.ts
shalvah: Oh, okay. You're right. I missed the first check, and I thought that was what this inner check was trying to do.

packages/js/src/server/integrations/uncaught_exception.ts
shalvah: Nice👍

shalvah: (Do you know if there's a reason why the rejection handler doesn't do this?)

shalvah (approved):

subzero10:
> (Do you know if there's a reason why the rejection handler doesn't do this?)

Are you referring to the `isReporting` check? I will add it! Do you think we should have a `handlerAlreadyCalled` check as well? Infinite loops are less common in async calls right?

packages/js/src/server/integrations/unhandled_rejection.ts
```diff
@@ -1,14 +1,43 @@
...
   return {
     load: (client: typeof Client) => {
-      if (!client.config.enableUnhandledRejection) { return }
+      if (!client.config.enableUnhandledRejection) {
+        return
+      }
+
+      process.on('unhandledRejection', function honeybadgerUnhandledRejectionListener(reason, _promise) {
+        if (!client.config.enableUnhandledRejection) {
+          if (!hasOtherUnhandledRejectionListeners() && !isReporting) {
```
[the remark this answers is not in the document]
shalvah: Yeah, fair.

shalvah (approved):

KonnorRogers:
Not to kind of cherry-pick here, but I wonder if rather than hooking into `uncaughtException` and changing its behavior, it seems like the better place to insert ourselves is `uncaughtExceptionMonitor`


https://nodejs.org/api/process.html#event-uncaughtexception
> It is possible to monitor 'uncaughtException' events without overriding the default behavior to exit the process by installing a 'uncaughtExceptionMonitor' listener.


https://nodejs.org/api/process.html#event-uncaughtexceptionmonitor

> Installing an 'uncaughtExceptionMonitor' listener does not change the behavior once an 'uncaughtException' event is emitted. The process will still crash if no 'uncaughtException' listener is installed.

```js
import process from 'node:process';

process.on('uncaughtExceptionMonitor', (err, origin) => {
  MyMonitoringTool.logSync(err, origin);
});

// Intentionally cause an exception, but don't catch it.
nonexistentFunc();
// Still crashes Node.js
```

It says `uncaughtExceptionMonitor` should also catch unhandledRejections.

It is worth noting this wasnt added until Node v12.17

shalvah:
Yeah, `uncaughtExceptionMonitor` is pretty recent. It was added in v12.17 (a very late release) _and 13.7_. So people on v12 and v13 less than these versions won't have it. It's something we should switch to in the future, but unlikely we can make use of it now, since we have to support v13 at least.

subzero10:
Hey @KonnorRogers, I learned about `uncaughtExceptionMonitor` while working on this issue!
Indeed, with this handler we can cover both `uncaughtException` and `unhandledRejection` errors, which would make things a lot easier in the code.

But as @shalvah mentioned, I wasn't sure if we should go for it now. 
However, we just decided that we will drop support for node v12 with our [next major version](https://github.com/honeybadger-io/honeybadger-js/milestone/3). We haven't decided _when_ is that going to be, but I guess pretty soon.
Then again I remember an [issue](https://github.com/honeybadger-io/honeybadger-js/issues/905#issuecomment-1267191741) from a user recently when they talked about a bug they discovered while upgrading from node v10 to v12.

Based on the above, I think we can release v5 soon, and replace `unhandledRejection` and `uncaughtException` with `uncaughtExceptionMonitor` in the near future.

What do you all think?

shalvah:
> Implement the uncaughtExceptionMonitor in the near future, after we release v5.

I'm okay, but note that uncaughtExceptionMonitor was added to v13 only in .7. Doing this means we're also dropping support for v13.0-13.6.

subzero10:
> > Implement the uncaughtExceptionMonitor in the near future, after we release v5.
> 
> I'm okay, but note that uncaughtExceptionMonitor was added to v13 only in .7. Doing this means we're also dropping support for v13.0-13.6.

Do you think this will be an issue? Like, would it require another major version bump? If I understand correctly, node js versions that aren't going to be LTS are not supposed to be used on production.

shalvah:
> If I understand correctly, node js versions that aren't going to be LTS are not supposed to be used on production.

No, LTS just means they have longer support periods. But there are a lot of folks (including myself) that have never cared about LTS and just use what's available.

But, idk, maybe we can drop support? 🤔 There were only three months between 13.0 and 13.7. (But note that things move fast in Node.js.  We're currently up to v18, but v13 was only released three years ago.) And if we're going to make bigger changes like separating the packages, we might as well go for it.

Which reminds me, do we have `engines` set in our package.json? Even though npm only considers it a warning, we can help users avoid surprises.

subzero10:
> No, LTS just means they have longer support periods. But there are a lot of folks (including myself) that have never cared about LTS and just use what's available.

I didn't look for the official blog post, but [here](https://www.geeksforgeeks.org/what-are-lts-versions-of-node-js/#:~:text=LTS%20version%20is%20an%20abbreviation,optimized%20performance%2C%20and%20bug%20fixes.) they claim that non LTS should not be used on production 🤷 :
>According to the Node.js blog, the “LTS version guarantees that the critical bugs will be fixed for a total of 30 months and Production applications should only use Active LTS or Maintenance LTS releases”.

> But, idk, maybe we can drop support? 🤔 There were only three months between 13.0 and 13.7. (But note that things move fast in Node.js. We're currently up to v18, but v13 was only released three years ago.) And if we're going to make bigger changes like separating the packages, we might as well go for it.

No objection, we could drop up to and including node v13 in our v5 milestone.

> Which reminds me, do we have engines set in our package.json? Even though npm only considers it a warning, we can help users avoid surprises.

Good point, we don't have. I will create an issue and include in the v5 milestone.

shalvah:
> they claim that non LTS should not be used on production

Ah, okay. Still, human behaviour doesn't conform to that. I can bet there's tons of applications running non-LTS versions in production, so we should make our decision irrespective of that. (I personally think the LTS distinction is just confusing. I'm glad Laravel is dispensing with it.)

Issue #686, opened 2022-01-20
Chrome Extension: `window is not defined` for background scripts


subzero10:
Looks like `window` is not available in background scripts.

Bug first reported [here](https://github.com/honeybadger-io/honeybadger-js/pull/665#issuecomment-1017604994).

joshuap:
Hmmm. I would love to avoid a situation where we have to guard access to `window` in the browser client. 🤔 

That said, something similar came up w/ someone who was trying to integrate with a Cloudflare worker (which is also on my integration wishlist).

christianopaets:
Instead of window we can use self. I have tried to to use smth like `global.window = self` and the error disapeared. But there is another problem)) `document` also doesn't exists

christianopaets:
@subzero10 any updates on it?

subzero10:
Hey @christianopaets, I haven't looked into this since January. It's in my todo list, I will try to bump it up.

loic-usertesting:
It would be really useful when working with a Browser extension as manifest v3 is becoming the norm...

zhanfengh:
Hello, we run a popular Chrome extension. And Honeybadger is an essential component of error reporting for us.

Google Chrome has mandated all Chrome extensions to be migrated to Manifest V3 and one of the notable changes of MV3 is that it [replaces background pages with service workers](https://developer.chrome.com/docs/extensions/mv3/mv3-migration/#man-sw). The DOM and `XMLHttpRequest()` [are not available in service workers](https://developer.chrome.com/docs/extensions/mv3/mv3-migration/#background-service-workers), which makes Honeybadger to error out.

We tried to fork and patch the library to make it compatible with service workers by replacing `window` with `self` when it's possible and `XMLHTTPRequest()` with `fetch()`. But we kept running into issues with a lot of different `window`/`document` calls within the library and couldn't figure out a clean way to fix them all.

It would be nice if you can look into this issue and make Honeybadger compatible with Manifest V3 soon as Google Chrome will not accept any extensions without MV3 after June 2023 and will actively remove them after Jan 2024. https://developer.chrome.com/blog/more-mv2-transition/

subzero10:
Hey @zhanfengh, thank you for your insights and feedback. Then, I guess I should look into this before the EOY!
cc @joshuap

guest271314:
You can alternatively substitute `globalThis` for `self` `console.assert(globalThis === self)`.

[...]

PR #958, opened 2022-11-23
feat: add window / document checks, refactor XMLHttpRequest to Fetch

KonnorRogers:
## Status

** Ready **

## Description

Reduce our calls to windows object, use fetch instead of XMLHTTPRequest, and other niceties to allow Chrome Extensions and WebWorkers to use Honeybadger.

## Related Issues

#686 
#956 

## Todos
- [x] Tests
- [x] Documentation

## Additional notes

In some light testing of cloudflare workers I discovered the following:\
-  "document", "window", and the "Location API" are not present.
- catching things like `onunhandledpromiserejection` doesn't actually work. 

Example:

```js
Honeybadger.configure({
 // ...
})

export default {
  async fetch(): Promise<Response> {
     throw Error("im an error")
     return new Response("Hello World!");
  },
};
```

The above will just straight up error and it doesnt get caught by honeybadger.

There are some docs on handling errors in CF workers, but they feel rather manual and I havent seen anything about catching global errors.

https://developers.cloudflare.com/workers/learning/logging-workers/

Happy to do some more digging, but you can at least call Honeybadger.notify() from a CF worker and it wont error.

```js
export default {
  async fetch(): Promise<Response> {
    Honeybadger.notify()
  },
};
```


packages/core/src/types.ts
```diff
@@ -123,10 +123,16 @@ export type ProcessStats = {
...
 export type TransportOptions = {
   method: 'GET' | 'POST',
+  // @TODO: This should probably follow the fetch "HeadersInit"
+  // headers?: HeadersInit
   headers?: Record<string, number | string | string[] | undefined>,
-  endpoint: string,
+  endpoint: string | URL,
   maxObjectDepth?: number,
   logger: Logger
+  /** @deprecated
```
subzero10, line 132: Any objection to remove `async` entirely?

packages/js/src/browser.ts
```diff
@@ -1,169 +1,200 @@
-import { Types, Util, Client } from '@honeybadger-io/core'
-import { encodeCookie, decodeCookie, preferCatch } from './browser/util'
-import { onError, ignoreNextOnError } from './browser/integrations/onerror'
-import onUnhandledRejection from './browser/integrations/onunhandledrejection'
-import breadcrumbs from './browser/integrations/breadcrumbs'
-import timers from './browser/integrations/timers'
-import eventListeners from './browser/integrations/event_listeners'
+import { Types, Util, Client } from '@honeybadger-io/core';
```
subzero10, line 1: Hmm, we have to decide on a convention about semi-colons (`;`) at the end of the statements.
So far, we've been avoiding them, but you may find some in some places.
Let's decide on it and we can create a separate PR with an eslint rule to enforce it.

Note: I didn't take this decision, I just went along with the convention :)

cc @joshuap @BethanyBerkowitz @stympy

packages/js/src/browser.ts
```diff
@@ -1,169 +1,200 @@
...
 
   constructor(opts: Partial<Types.BrowserConfig> = {}) {
-    super({
-      async: true,
-      maxErrors: null,
-      projectRoot: window.location.protocol + '//' + window.location.host,
-      ...opts
-    }, new BrowserTransport())
+    const global = globalThisOrWindow();
+    let projectRoot = ""
```
subzero10: You can run `npm lint` locally to fix lint errors.

packages/js/src/browser/integrations/breadcrumbs.ts
```diff
@@ -1,267 +1,319 @@
...
+              metadata: {
+                selector,
+                text,
+                event,
+              },
+            });
+          },
+          // In CloudFlare workers this must be false. This is a hacky way to detect it.
+          // CF workers require useCapture to be false.
+          _window.location ? true : false
```
subzero10, line 113: Shouldn't you `return` here?

subzero10:
Hey @KonnorRogers, I started going through the PR (I know it's not ready, just wanted to have an idea) but it's difficulty to follow with all the formatting changes. Can you please remove the changed semi-colons and we can decide on a convention at a different PR?

packages/js/src/browser.ts
KonnorRogers: @subzero10 I piped these through prettier, but repiped them back through ts-standard.

IIRC most of our repos use either "standard" or "js-standard"

KonnorRogers:
@subzero10 Was already looking at fixing the diff to be easier to follow 🙈

packages/js/src/browser.ts
subzero10: The monorepo has a global eslint configuration. I created the rules based on different styles I picked up from our repos while implementing it. It's quite possible I missed or mixed some. Nothing's set it stone though, we can always improve.

subzero10:
> @subzero10 Was already looking at fixing the diff to be easier to follow 🙈

Awesome, thanks! I will come back to this later then 😄

packages/js/src/browser.ts
KonnorRogers: @subzero10 the problem was I've recently remove "expandtab" from my Vim config and for some reason ESLint doesnt fix the "mixed spaces + tabs" issues and going in and fixing it manually is a pain. I just reverted my Vim config to "expandtab" for now

KonnorRogers:
@subzero10 should be much easier to parse now.

packages/js/src/browser/integrations/breadcrumbs.ts
KonnorRogers: Shouldnt need one, this is a parameter being passed to the event listener.

subzero10: Oh, ignore what I said. I thought this was the return of a function 🤦

packages/core/src/types.ts
```diff
@@ -123,11 +123,12 @@ export type ProcessStats = {
 
 export type TransportOptions = {
   method: 'GET' | 'POST',
+  // @TODO: This should probably follow the fetch "HeadersInit"
+  // headers?: HeadersInit
```
subzero10, line 127: Hmm, this type is also used by the server transport. Isn't `HeadersInit` a browser type?

packages/js/jest-browser-setup.js
```diff
@@ -0,0 +1,2 @@
+const fetchMock = require('jest-fetch-mock')
```
subzero10, line 1: I think it would better if this file was inside the test folder. Is there any reason to have it in the root folder?

packages/js/src/browser/transport.ts
```diff
@@ -1,27 +1,46 @@
...
+
+    try {
+      const response = await globalThisOrWindow().fetch(options.endpoint, requestInit)
+
+      const body = await response.text()
+
+      return Promise.resolve({ statusCode: response.status, body })
+    } catch (err) {
+      return Promise.reject(err)
+    }
```
subzero10, line 44: If you are not doing anything inside the catch, you could just remove the try / catch entirely right?

packages/js/test/integration/worker.js
```diff
@@ -0,0 +1,20 @@
+self.importScripts('/base/dist/browser/honeybadger.js')
```
subzero10, line 1: ❤️ Nice! Love the integration test with web workers!

packages/js/test/unit/browser.test.ts
```diff
@@ -1,24 +1,19 @@
...
-  let xhr: typeof useFakeXMLHttpRequest
 
   beforeEach(function () {
     client = Singleton.factory({
       logger: nullLogger()
-    });
+    })
 
-    // @ts-ignore - no need to test this in here
+    // @ts-expect-error - no need to test this in here
```
subzero10, line 13: 👍

subzero10 (dismissed):
Great work!
Did you get a chance to also test manually? What did you test with (i.e. browser, nodejs, chrome extension, web worker)?
I think it's necessary here because our integration tests mock the http call, which is something that we are changing with the PR.

packages/core/src/types.ts
KonnorRogers: Sounds good. I'll remove this, as it'd be a pain. Node 18 will be coming with native fetch + `Headers`, but thats a long ways away!

packages/js/src/browser/transport.ts
KonnorRogers: Hmmm...my preference is to always try / catch to avoid unhandledpromiserejections...but I don't see why it'd matter...I don't know that I have a good gauge on 1 vs the other. Will revert it.

packages/js/jest-browser-setup.js
KonnorRogers: No reasons, was just easier to have it live next to `jest.config.js`

KonnorRogers: Fixed!

KonnorRogers:
> Great work! Did you get a chance to also test manually? What did you test with (i.e. browser, nodejs, chrome extension, web worker)? I think it's necessary here because our integration tests mock the http call, which is something that we are changing with the PR.

I did get a chance to test manually in the following environments:

- Chrome Extension
- Browser
- Web Worker
- Cloudflare worker

I basically tested that `Honeybadger.configure()` and `Honeybadger.notify()` didnt produce errors and were able to send reports.

I did not test NodeJS since I didnt touch anything in core or in server.

packages/js/src/browser/transport.ts
subzero10: In this case, the promise is caught and immediately re-thrown, so it has the same behavior as not having a catch block at all, no?

subzero10:
> > Great work! Did you get a chance to also test manually? What did you test with (i.e. browser, nodejs, chrome extension, web worker)? I think it's necessary here because our integration tests mock the http call, which is something that we are changing with the PR.
> 
> I did get a chance to test manually in the following environments:
> 
> * Chrome Extension
> * Browser
> * Web Worker
> * Cloudflare worker
> 
> I basically tested that `Honeybadger.configure()` and `Honeybadger.notify()` didnt produce errors and were able to send reports.
> 
> I did not test NodeJS since I didnt touch anything in core or in server.

Sounds great! Yeah I don't see a reason to manually test server-side. We are good to go then!

subzero10 (dismissed):

subzero10:
@KonnorRogers Btw, I think this is more of a feature rather than a fix. What do you think?

KonnorRogers:
Id agree with feature here.

KonnorRogers:
Fun reminder of how annoying this problem is with detecting where a library is being called from:

https://twitter.com/TkDodo/status/1598992656226344960

https://twitter.com/alexdotjs/status/1598996196357738497

subzero10 (approved):

thatsmydoing:
Given the switch to using `fetch` that means IE is fully unsupported, should there be a list of supported browsers / runtimes in the README or documentation?

subzero10:
> Given the switch to using `fetch` that means IE is fully unsupported, should there be a list of supported browsers / runtimes in the README or documentation?

Actually, there is a page in our [docs](https://docs.honeybadger.io/lib/javascript/reference/supported-versions/), but it's outdated (it says that IE 11 is supported).
We will update it and probably create a table with more details on supported versions per browser.

Thank you for pointing it out!

PR #972, opened 2022-12-06
ci: ignore eslint imports in example folders

BethanyBerkowitz:
## Status

**READY**

## Description
Currently, running `npm run lint` fails locally unless you remember to install all the dependencies in each of the example folders, plus we have to install all those dependencies in CI. Since these are just the examples and not the main code, we can save time in CI and have fewer local errors during development if we selectively ignore these. 

## Steps to Test or Reproduce
```bash
npm run lint
```
1. You should not see errors, even if you neglected to `cd` into each example folder and `npm install`
2. CI should succeed without having to install these extra node modules just for linting

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -90,22 +90,5 @@ jobs:
       - name: Build
         run: npm ci
 
-      - name: Build Typescript example projects
-        run: |
-          cd packages/js/examples/aws-lambda-typescript
-          npm ci
-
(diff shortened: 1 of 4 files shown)


BethanyBerkowitz:
Note: reduces the lint step of CI from 2 minutes to 1 minute (ish)

subzero10 (approved):
🎉 Super clean! Nicely done.

Issue #979, opened 2022-12-09
Unrelated errors getting grouped together


subzero10:
Hello,

We are currently using honeybadger with a node/express application, and continue to see an issue that we're not sure of the best way to solve. Unrelated errors get grouped under a single error item within honeybadger, which is obscuring some errors, and just in general gets them mixed up. Here's one example: https://app.honeybadger.io/projects/103978/faults/91269777

And a screenshot of what it looks like when you click through:
Screen Shot 2022-12-07 at 10.34.16 AM.png

Note that these are two totally different errors from different parts of the application. Can you help us to understand why this would happen, or if there are any steps we can take on our side to prevent this when sending errors to honeybadger?

Thank you!
Eric

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_fk0pt)

subzero10:
Update on this:
- I went through the notice in example and I realised that these notices were not reported with a stack trace (i.e. notice was reported with a message:`hb.notify('Error creating index for model "adChoices":')`)
- When this happens, we [generate a stack trace internally](https://github.com/honeybadger-io/honeybadger-js/blob/a5fd0e973b02bf075afb85752ce71839b216ae44/packages/core/src/client.ts#L317) and then we just remove the first 2 frames of the stack. **This is what broke.**
- The 2 frames made the assumption that only 2 functions calls would happen inside our js client. This has changed with the introduction of `notifyAsync` and the extraction to a core js package.

I am working on a way to filter out the first frames that are pointing to the honeybadger source code, without specifying a hardcoded number. This way, we will have a more future-proof solution.
For example, we could loop through the stack and remove the frames that have source file from `'@honeybadger-io'`:
```
/**
 * If {@link generateStackTrace} is used, we want to exclude frames that come from
 * honeybadger's source code.
 */
function calculateBacktraceShift(backtrace: BacktraceFrame[]) {
  let shift = 0
  for (const line of backtrace) {
    if (line.file.indexOf('@honeybadger-io') > -1) {
      shift++
    }
    else {
      break
    }
  }

  return shift
}
```

I will have a PR for review soon (probably tomorrow).

subzero10:
OK, PR #982 is ready with a fix, but I'm pretty sure it won't resolve the user's issue.

Upon further inspection I noticed the following:
[...]

PR #982, opened 2022-12-16
fix: filter out honeybadger source code from stack trace

subzero10:
## Status
**READY**

## Description
Fixes: #979

## Todos
- [x] Fix
- [x] Tests
- [x] Manual Tests

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -16,7 +16,6 @@ List related PRs against other branches:
 ## Todos
 - [ ] Tests
 - [ ] Documentation
-- [ ] Changelog Entry (unreleased)
 
 ## Steps to Test or Reproduce
 Outline the steps to test or reproduce the PR here.
--- a/packages/core/src/util.ts
+++ b/packages/core/src/util.ts
@@ -94,7 +94,7 @@ export function calculateBacktraceShift(backtrace: BacktraceFrame[]) {
       continue
     }
 
-    if (frame.file || frame.file === '<anonymous>') {
+    if (!frame.file || frame.file === '<anonymous>') {
       const nextFrame = backtrace[i + 1]
       if (nextFrame && isFrameFromHbSourceCode(nextFrame)) {
         shift++
--- a/packages/core/src/util.ts
+++ b/packages/core/src/util.ts
@@ -39,7 +39,7 @@ export function objectIsExtensible(obj): boolean {
   return Object.isExtensible(obj)
 }
 
-export function makeBacktrace(stack: string, filterHbSourceCode = false): BacktraceFrame[] {
+export function makeBacktrace(stack: string, filterHbSourceCode = false, logger: Logger = console): BacktraceFrame[] {
   try {
     const backtrace = stackTraceParser
       .parse(stack)
@@ -56,17 +56,26 @@ export function makeBacktrace(stack: string, filterHbSourceCode = false): Backtr
     }
 
     return backtrace
-  } catch (_err) {
-    // TODO: log error
+  } catch (err) {
+    logger.debug(err)
     return []
   }
 }
 
-const DEFAULT_BACKTRACE_SHIFT = 3
-
 function isFrameFromHbSourceCode(frame: BacktraceFrame) {
-  return frame.file.indexOf('@honeybadger-io') > -1 || frame.method.indexOf('@honeybadger-io') > -1
+  let hasHbFile = false
+  let hasHbMethod = false
+  if (frame.file) {
+    hasHbFile = frame.file.toLowerCase().indexOf('@honeybadger-io') > -1
+  }
+  if (frame.method) {
+    hasHbMethod = frame.method.toLowerCase().indexOf('@honeybadger-io') > -1
+  }
+
+  return hasHbFile || hasHbMethod
 }
+
+export const DEFAULT_BACKTRACE_SHIFT = 3
 
 /**
  * If {@link generateStackTrace} is used, we want to exclude frames that come from
@@ -108,7 +117,7 @@ export function calculateBacktraceShift(backtrace: BacktraceFrame[]) {
   return shift || DEFAULT_BACKTRACE_SHIFT
 }
 
-export function getCauses(notice: Partial<Notice>) {
+export function getCauses(notice: Partial<Notice>, logger: Logger) {
   if (notice.cause) {
     const causes =[]
     let cause = notice as Error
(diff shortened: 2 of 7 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -16,7 +16,6 @@ List related PRs against other branches:
 ## Todos
 - [ ] Tests
 - [ ] Documentation
-- [ ] Changelog Entry (unreleased)
```
subzero10, line 19: Changelog is generated automatically, so this is redundant.

packages/core/src/util.ts
```diff
@@ -51,14 +51,63 @@ export function makeBacktrace(stack: string, shift = 0): BacktraceFrame[] {
...
+ * - Exit when frame encountered is not from Honeybadger source code
+ *
+ * Note: this will not always work, especially in browser versions where code
+ *       is minified, uglified and bundled.
+ *       For those cases we default to 3:
+ *       - generateStackTrace
+ *       - makeNotice
+ *       - notify
+ */
+export function calculateBacktraceShift(backtrace: BacktraceFrame[]) {
```
subzero10, line 88: I'm not excited with this, but I think it's better than the default approach (i.e. always remove first 3 frames)

packages/core/test/client.test.ts
```diff
@@ -309,18 +311,6 @@ describe('client', function () {
       expect(payload.error.backtrace.length).toBeGreaterThan(0)
     })
 
-    it('generates a backtrace when there isn\'t one', function () {
```
subzero10, line 312: Moved further down.

stympy (dismissed):

packages/core/src/util.ts
```diff
@@ -51,14 +51,63 @@ export function makeBacktrace(stack: string, shift = 0): BacktraceFrame[] {
...
   } catch (_err) {
     // TODO: log error
     return []
   }
 }
 
+const DEFAULT_BACKTRACE_SHIFT = 3
+
+function isFrameFromHbSourceCode(frame: BacktraceFrame) {
+  return frame.file.indexOf('@honeybadger-io') > -1 || frame.method.indexOf('@honeybadger-io') > -1
```
BethanyBerkowitz, line 68: Perhaps more readable as `frame.file.includes('@honeybadger-io') || frame.method.includes('@honeybadger-io')`

Also can we assume lower case?

packages/core/src/util.ts
```diff
@@ -51,14 +51,63 @@ export function makeBacktrace(stack: string, shift = 0): BacktraceFrame[] {
...
+export function calculateBacktraceShift(backtrace: BacktraceFrame[]) {
+  let shift = 0
+  for (let i = 0; i < backtrace.length; i++) {
+    const frame = backtrace[i]
+    if (isFrameFromHbSourceCode(frame)) {
+      shift++
+      continue
+    }
+
+    if (frame.file || frame.file === '<anonymous>') {
```
shalvah, line 97: What's the scenario where a frame's file is `<Anonymous>`? THat's missing in the explanation.

BethanyBerkowitz: Confused by this... if `frame.file === '<anonymous>'` is true, then `frame.file` is also truthy. Typo or am I missing something?

packages/core/src/util.ts
```diff
@@ -51,14 +51,63 @@ export function makeBacktrace(stack: string, shift = 0): BacktraceFrame[] {
...
+    }
+
     return backtrace
   } catch (_err) {
     // TODO: log error
     return []
   }
 }
 
+const DEFAULT_BACKTRACE_SHIFT = 3
```
BethanyBerkowitz, line 65: Are there other constants like this? Would be nice to store them all in one place, either at the top of the file or in a separate constants file.

packages/core/test/client.test.ts
```diff
@@ -1,5 +1,6 @@
 import { nullLogger, TestClient, TestTransport } from './helpers'
 import { Notice } from '../src/types'
+import { makeBacktrace,  } from '../src/util'
```
BethanyBerkowitz, line 3: Excess comma?

packages/core/src/util.ts
subzero10: Your comment saved me! I missed a negation (it should be `if (!frame.file ...`).

To answer your question, this came up from the fault I was debugging. I don't know when it happens, but I know it can happen. That's why I came up with the [not-so-straight-forward] approach to ignore these frames if they happen to be in-between our source code. Otherwise, the algorithm would stop here and would not see that the next frame is still honeybadger source code.

See attached image.

image

packages/core/test/client.test.ts
```diff
@@ -338,6 +328,80 @@ describe('client', function () {
...
+      const backtrace = makeBacktrace(error.stack, true)
+      expect(backtrace[0]).toEqual({
+        file: '/var/www/somebody/node_modules/@somebody/logger/HoneybadgerTransport.js',
+        method: 'HoneybadgerTransport.log',
+        number: 18,
+        column: 19
+      })
+    })
+
+    it('filters out top frames that come from @honeybadger-io (browser)', function () {
```
BethanyBerkowitz, line 389: Might be helpful to also add a test where your code fails to find the honeybadger entries and just cuts the default number, mostly as a way to document the behavior.

BethanyBerkowitz (commented):
Minor questions / comments but looking good!

packages/core/src/util.ts
```diff
@@ -51,14 +51,63 @@ export function makeBacktrace(stack: string, shift = 0): BacktraceFrame[] {
...
         }
       })
-    backtrace.splice(0, shift)
+    if (filterHbSourceCode) {
+      backtrace.splice(0, calculateBacktraceShift(backtrace))
+    }
+
     return backtrace
   } catch (_err) {
     // TODO: log error
```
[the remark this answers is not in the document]
subzero10: I think it's because we don't want to log internal errors to our users' console.
I will add a logger.debug() 👌 .
Thanks for pointing it out!

packages/core/src/util.ts
subzero10: I usually go for the `indexOf` because it has better browser support, but I guess this shouldn't be a concern anymore. I will change it!

> Also can we assume lower case?

I think so. But it doesn't hurt to add a `toLowerCase()`! Will do!

packages/core/src/util.ts
subzero10: Yes, typo! Fixed!

packages/core/src/util.ts
subzero10: We have _some_ constants and so far we've been placing them near their usage (or where they most make sense).
For me, it works OK so far. But if you have a strong opinion on refactoring them (i.e. moving them to separate file, creating a convention), we can discuss it more 😄

subzero10:
@BethanyBerkowitz All review feedback applied. I've just requested another round of review :)

BethanyBerkowitz (approved):
🚀 🌔 🪐

PR #984, opened 2022-12-22
feat: allow async beforeNotify functions to modify the notice object

### LINKED ISSUE #981: Is it possible to have async beforeNotify handler?
I would like to be able to do something like:
```
Honeybadger.beforeNotify(async (notice) => {
  if (notice) {
    notice.context = {
      state: await dumpState(),
    };
  }

  return true;
});
```
So that we can call async methods to add data to the notice.

## What happens?
Typescript complains because the beforeNotify handler doesn't allow Promises.

I would like to be able to use async functions in beforeNotify.

n/a

**Package Name:** honeybadger-js
**Package Version:** 4.8.0

KonnorRogers:
## Status
**READY**

## Description
Fixes #981.
Allow async beforeNotify handlers to modify the `notice` object before sending to Honeybadger.

## Todos
- [x] `async beforeNotifyHandler`
- [x] `async afterNotifyHandler`
- [x] Tests (fix integration tests)
- [x] [Documentation](https://github.com/honeybadger-io/docs/pull/316)


packages/core/src/util.ts
```diff
@@ -95,15 +95,25 @@ export async function getSourceForBacktrace(backtrace: BacktraceFrame[],
   return result
 }
 
-export function runBeforeNotifyHandlers(notice: Notice | null, handlers: BeforeNotifyHandler[]): boolean {
+export function runBeforeNotifyHandlers(notice: Notice | null, handlers: BeforeNotifyHandler[]): { results: ReturnType<BeforeNotifyHandler>[], result: boolean } {
```
subzero10, line 98: `ReturnType`! I always forget that this exists! ❤️

packages/core/src/util.ts
```diff
@@ -95,15 +95,25 @@ export async function getSourceForBacktrace(backtrace: BacktraceFrame[],
...
+  }
+
+  return {
+    results,
+    result
   }
-  return result
 }
 
 export function runAfterNotifyHandlers(notice: Notice | null, handlers: AfterNotifyHandler[], error?: Error): boolean {
```
subzero10, line 119: Wondering, if we should support async afterNotifyHandlers as well? It would be a lot easier, considering we don't need to read the result later on. Right?

packages/core/test/client.test.ts
```diff
@@ -597,6 +597,56 @@ describe('client', function () {
...
+        client.afterNotify((_err, notice) => {
+          expect(notice.name).toEqual(funkyName)
+          resolve()
+        })
+
+        expect(client.notify('Should report and modify notice')).toEqual(true)
+      })
+    })
+
+    it('modifies the notice when an async function is provided', function () {
```
subzero10, line 626: Typo? This is the same test as above. Perhaps you need another test to confirm that notice won't be sent if an async handler returns false.

packages/core/src/util.ts
KonnorRogers: So handy, and I usually forget as well!

packages/core/src/client.ts
```diff
@@ -151,89 +151,32 @@ export abstract class Client {
...
+        }
+        return this.__send(notice, sourceCodeData)
+      })
 
-    getSourceForBacktrace(sourceCodeData, this.__getSourceFileHandler)
-      .then(sourcePerTrace => {
-        sourcePerTrace.forEach((source, index) => {
-          notice.backtrace[index].source = source
-        })
+      return true
```
subzero10, line 176: In case of an async beforeNotify handler, we return true regardless of the result.

packages/core/src/client.ts
```diff
@@ -418,4 +361,108 @@ export abstract class Client {
...
+      this.logger.log('honeybadger.js is in development mode; the following error report will be sent in production.', notice)
+      preConditionError = new Error('honeybadger.js is in development mode')
+    }
+
+    if (!this.config.apiKey) {
+      this.logger.warn('could not send error report: no API key has been configured', notice)
+      preConditionError = new Error('missing API key')
+    }
+
+    const beforeNotifyResult = runBeforeNotifyHandlers(notice, this.__beforeNotifyHandlers)
```
BethanyBerkowitz, line 387: If there's a `preConditionError`, why do we still run the `beforeNotifyHandlers`?

packages/core/src/util.ts
```diff
@@ -154,15 +158,23 @@ export async function getSourceForBacktrace(backtrace: BacktraceFrame[],
   return result
 }
 
-export function runBeforeNotifyHandlers(notice: Notice | null, handlers: BeforeNotifyHandler[]): boolean {
+export function runBeforeNotifyHandlers(notice: Notice | null, handlers: BeforeNotifyHandler[]): { results: ReturnType<BeforeNotifyHandler>[], result: boolean } {
```
BethanyBerkowitz, line 161: Hrm, it's confusing to me that `result` here will be `true` if the handlers return promises that eventually reject. However, I'm not sure of a better solution. 

One option could be something like an enum `status`... 

```javascript
function runBeforeNotifyHandlers(...) {
  const results = handlers.map(handler => handler(notice))

  if (results.find(result => result === false) {
    return { results, status: 'failed' }
  }

  if (results.find(result => result instanceof Promise) {
    return { results, status: 'pending' }
  }
  
  return { results, status: 'passed' }
}
```
Just an idea... the larger problem of mixing sync and async functions here is just thorny and I'm trying to think of ways it could be cleaner.

packages/core/src/client.ts
BethanyBerkowitz: I understand why this is working this way, but this is smelly. A user would definitely not intuitively know how to interpret this boolean. 

I took a quick look through the documentation but I couldn't find anywhere where we explain this return value or give an example of its use. Do we know if users rely on it at all? 

If the return value simply indicates that we ran the precondition checks (which are all sync) and does not provide any information about the results of the `beforeNotify` handlers, this could be understandable for a user. But imo the idea that it returns a meaningful result of the `beforeNotify` handlers, but only if all of them are sync is confusing. Getting `true` back when it actually means 'promises are pending'... I think it's worse than getting nothing back at all. 

I think I'd be in favor of just returning the result of the precondition checks and clearly documenting that behavior somewhere. I'm not sure if this would constitute a "breaking change" or not if the prior behavior wasn't documented... theres a chance that savvy users are relying on knowing if their `beforeNotify` handler failed even if its not in the docs.

packages/core/src/client.ts
```diff
@@ -151,89 +151,32 @@ export abstract class Client {
...
-    }
     // we need to have the source file data before the beforeNotifyHandlers,
     // in case they modify them
     const sourceCodeData = notice && notice.backtrace ? notice.backtrace.map(trace => shallowClone(trace) as BacktraceFrame) : null
-    const beforeNotifyResult = runBeforeNotifyHandlers(notice, this.__beforeNotifyHandlers)
-    if (!preConditionError && !beforeNotifyResult) {
-      this.logger.debug('skipping error report: beforeNotify handlers returned false', notice)
-      preConditionError = new Error('beforeNotify handlers returned false')
-    }
+    const preConditionsResult = this.__runPreconditions(notice)
```
BethanyBerkowitz, line 159: Big fan of moving the preconditions to their own function! 

I think it would be more intuitive if the `beforeNotify` handlers were not run within the precondition function -- conceptually they are pretty different since the preconditions are defined by us and the `beforeNotify` handlers are defined by the user.

packages/core/src/client.ts
```diff
@@ -418,4 +361,108 @@ export abstract class Client {
...
+          if (preConditionError) {
+            return preConditionError
+          }
+        })
+    }
+
+    return preConditionError
+  }
+
+  private __send(notice: Notice, originalBacktrace: BacktraceFrame[]) {
```
BethanyBerkowitz, line 410: Big fan of having this broken out into its own function, thanks!

packages/core/src/types.ts
```diff
@@ -38,12 +38,12 @@ export interface ServerlessConfig extends Config {
...
 
 export interface BeforeNotifyHandler {
-  (notice?: Notice): boolean | void
+  (notice?: Notice): boolean | void | Promise<boolean | void>
 }
 
 export interface AfterNotifyHandler {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  (error: any, notice?: Notice): boolean | void
+  (error: any, notice?: Notice): boolean | void | Promise<void>
```
BethanyBerkowitz, line 46: Can the `AfterNotifyHandler` resolve to a boolean? Just `void`?

packages/core/src/util.ts
```diff
@@ -40,6 +40,10 @@ export function objectIsExtensible(obj): boolean {
 }
 
 export function makeBacktrace(stack: string, filterHbSourceCode = false, logger: Logger = console): BacktraceFrame[] {
+  if (typeof stack === 'undefined') {
```
BethanyBerkowitz, line 43: Same behavior if `stack` is an empty string? If so, you could just go with `if (!stack)`

packages/core/src/util.ts
BethanyBerkowitz: If we decide not to try to return the status of the `beforeNotify` handlers to the user, perhaps this could be an async function and we could just wait for `Promise.allSettled()` before returning the result.

BethanyBerkowitz (commented):
Not trying to gum up progress here, but I don't feel great about how the return value of `notify()` is working in this PR.. let's discuss further?

packages/core/src/client.ts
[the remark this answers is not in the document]
subzero10: Good question! I had this discussion with @joshuap in the [past](https://github.com/honeybadger-io/honeybadger-js/pull/659), and we agreed to follow the convention applied in the Ruby package, which is to always call the beforeNotify and afterNotify hooks.

I think the point here is that users _may_ want to do something with the error even if it's not going to be sent to Honeybadger.

packages/core/src/client.ts
subzero10: > I took a quick look through the documentation but I couldn't find anywhere where we explain this return value or give an example of its use. Do we know if users rely on it at all?

Knowing the answer to this question would greatly help us to decide whether we should be trying to stay backwards compatible or not. Actually, that's why it took us so long to work on this feature request. Unfortunately, we do not have this kind of information.

> If the return value simply indicates that we ran the precondition checks (which are all sync) and does not provide any information about the results of the beforeNotify handlers, this could be understandable for a user.

@joshuap explained that the boolean result can be used as an indicator to know whether the notice will be reported to HB (in a synchronous manner). I think it's a convention that came from other HB client integrations, but I'm not sure how much is used today.

> But imo the idea that it returns a meaningful result of the beforeNotify handlers, but only if all of them are sync is confusing.

True, I agree with you. Since we are missing usage data (how is the result of notify being used), we tried to implement the new feature (async beforeNotify) without breaking the existing functionality:
- the result of notify() remains consistent if sync beforeNotify handlers are used
- the result of notify() is always true if async beforeNotify handlers are used <- this is the new feature and we can document it to manage expectations

> Getting true back when it actually means 'promises are pending'...

You'd be getting that only with the new behavior, which we can document today.

> I think I'd be in favor of just returning the result of the precondition checks and clearly documenting that behavior somewhere. I'm not sure if this would constitute a "breaking change" or not if the prior behavior wasn't documented... theres a chance that savvy users are relying on knowing if their beforeNotify handler failed even if its not in the docs.

If we go this route, we'll have to make notify an async function, since we support async precondition checks, right? This goes against our approach of "fire-and-forget" for error reporting.

Btw, I'm loving the fact that you have the same concerns I had in the past when I first encountered this feature request 😄 .

packages/core/src/types.ts
subzero10: Hmm, I missed the fact that we allow a boolean to be returned from this handler. I looked at the code and we are not really looking at the result of these handlers, so I think we _could_ remove the `boolean` return option as well.

packages/core/src/util.ts
subzero10: Yes! Will do that! This is unrelated to this PR but I noticed that I was getting some debug logs that were unnecessary.

subzero10: Done!

packages/core/src/types.ts
subzero10: Done!

packages/core/src/util.ts
subzero10: Let's continue this discussion [here](https://github.com/honeybadger-io/honeybadger-js/pull/984#discussion_r1271868376).

packages/core/src/client.ts
subzero10: 🤔 Hmm, the way I thought about this is that `beforeNotify` handlers are a subset of the precondition checks.
And we even have beforeNotify handlers that are defined by us (see `browser.ts` and `server.ts` in the js package).

BethanyBerkowitz: Reasonable either way!

packages/core/src/client.ts
subzero10: I added documentation about [Honeybadger.notify() return value](https://github.com/honeybadger-io/docs/pull/316/files#diff-8658674704fa4cb438d9f97f2ca0233be7cba8012c578766ebf88ecf20c48ea6R69).

BethanyBerkowitz (approved):
The documentation helps, thanks. I still think it's not ideal, but I understand wanting to keep backwards-compatibility!

packages/core/src/client.ts
joshuap: I sorta get what's going on here... sounds correct. :grin:

We used to use the return value pattern in our Ruby gem (and may still elsewhere, I forget), but at least for Ruby we've since moved to calling a method on the notice to prevent it from being sent:

https://docs.honeybadger.io/lib/ruby/getting-started/ignoring-errors/#ignore-programmatically

subzero10:
I will go ahead and merge this to get this feature out, but here's what I'm thinking:
- we could ditch the return value entirely
- knowing that one can get this information by setting an `afterNotify` handler (if a notice is not going to be sent, an `afterNotify` handler will be called with the notice and an error)
- this should probably be a major version bump

What do you think?

joshuap:
> * we could ditch the return value entirely

How would a user [programmatically stop a notice from being sent](https://docs.honeybadger.io/lib/javascript/guides/reducing-noise/#ignoring-errors)?

subzero10:
> > * we could ditch the return value entirely
> 
> How would a user [programmatically stop a notice from being sent](https://docs.honeybadger.io/lib/javascript/guides/reducing-noise/#ignoring-errors)?

I am referring to the return value of the `notify()` function, not the `beforeNotify` handlers. Does this make sense?
(I edited my comment above to make it more clear)

joshuap:
> > > * we could ditch the return value entirely
> > 
> > 
> > How would a user [programmatically stop a notice from being sent](https://docs.honeybadger.io/lib/javascript/guides/reducing-noise/#ignoring-errors)?
> 
> I am referring to the return value of the `notify()` function, not the `beforeNotify` handlers. Does this make sense? (I edited my comment above to make it more clear)

Gotcha—yeah, I think that makes sense.

subzero10:
> Gotcha—yeah, I think that makes sense.

Done! See #1162.
cc @BethanyBerkowitz

PR #1000, opened 2023-01-13
fix: improve generateStackTrace function to hide vue warn message

### LINKED ISSUE #996: [Vue warn]: Property ("_isVue" | "constructor") was accessed during render but is not defined on instance.
I'm using honeybadger with Laravel, Inertia and Vue3.

Honeybadger seems to be accessing a Vue2 deprecated property.

I found the equivalent issue in Sentry while searching the error
https://github.com/getsentry/sentry-javascript/issues/4170

And the PR that fixed it:
https://github.com/getsentry/sentry-javascript/pull/4461/files

1. The error happened when calling `includes` on `null` in a computed property.

## What happens?
Some errors but not others make it so honeybadger throws its own error. 
`[Vue warn]: Property "_isVue" was accessed during render but is not defined on instance.`

…

Following the stacktrace in the browser points to this line:
`Function.prototype.apply.call(original, _window.console, arguments);`

## Any other comments?

Here are the places where it's used:

https://github.com/honeybadger-io/honeybadger-js/blob/317220310f07edd18283214fe8901cab31218416/packages/vue/src/vue-debug.js#L17

https://github.com/honeybadger-io/honeybadger-js/blob/317220310f07edd18283214fe8901cab31218416/packages/vue/src/vue-debug.js#L44

**Operating System:** Windows 10
**Package Name:**  @honeybadger-io/vue  
**Package Version:** 4.8.2
**Browser Version:** Chrome

subzero10:
## Status
**READY**
Fixes: #996

## Description
Improves `generateStackTrace` function to hide `Vue[warn]` message about undefined property `_isVue`.

*Bonus*: adds `jest-fetch-mock` to remove some errors logged in tests

### DIFF
--- a/packages/vue/src/vue-debug.js
+++ b/packages/vue/src/vue-debug.js
@@ -1,5 +1,5 @@
 /**
- * This was taken from https://github.com/vuejs/vue/blob/master/src/core/util/debug.js.
+ * This was originally taken from https://github.com/vuejs/vue/blob/master/src/core/util/debug.js.
  * The method generateStackTrace is used to log errors the same way as Vue logs them when errorHandler is not set.
  */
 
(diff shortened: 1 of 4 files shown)


packages/vue/src/vue-debug.js
```diff
@@ -8,15 +8,27 @@ const classify = str => str
...
   if (vm.$root === vm) {
-    return '<Root>'
+    return ROOT_COMPONENT
+  }
+
+  if (!vm.$options) {
+    return ANONYMOUS_COMPONENT
+  }
+
+  const options = vm.$options
```
subzero10, line 27: I simplified this using Sentry's implementation as a reference point.

stympy (dismissed):

packages/vue/src/vue-debug.js
```diff
@@ -8,15 +8,27 @@ const classify = str => str
...
+  if (!options) {
+    return ANONYMOUS_COMPONENT
```
BethanyBerkowitz, line 29: Duplicate of lines 23-25 above?

BethanyBerkowitz (dismissed):
One minor comment, but no need to re-review

packages/vue/src/vue-debug.js
subzero10: Yes, nice catch! Thanks!

PR #1002, opened 2023-01-18
fix: replace instrument with safer alternative addEventListener

### LINKED ISSUE #1001: Uncaught Error: Permission denied to access property "__hb_original"
I haven't figured out how to reliably reproduce the issue yet, but it seems to be some combination of honeybadger.js + the [Readwise Highlighter](https://addons.mozilla.org/en-US/firefox/addon/readwise-highlighter/) plugin in Firefox.

## What happens?
The following error is in the console, and in at least one case it prevented the customer's app from loading:

```
Uncaught Error: Permission denied to access property "__hb_original"
  instrument honeybadger.js:580
  load honeybadger.js:1716
  load honeybadger.js:1737
  configure honeybadger.js:830
  configure honeybadger.js:830
  configure honeybadger.js:2045
```

We've also seen this error in our own projects—[here's an error report](https://titanic.honeybadger.io/projects/5/faults/33117#notice-context).

Here's a user agent from that titanic error report:

```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:108.0) Gecko/20100101 Firefox/108.0
```

That instance was with @honeybadger-io/js 4.8.2. Another customer is using 3.2.8 and has the same error (that's the one that causes their app to crash).

Front logo [Front conversations](https://app.frontapp.com/open/top_henxd)

subzero10:
## Status
**READY**

## Description
Fixes: #1001
- Replaces `instrument` with the native `addEventListener`.
- Adds error control when accessing `__hb_original`.

### DIFF
--- a/packages/js/src/browser/integrations/breadcrumbs.ts
+++ b/packages/js/src/browser/integrations/breadcrumbs.ts
@@ -247,15 +247,11 @@ export default function (_window = globalThisOrWindow()): Types.Plugin {
           })
         }
 
-        // https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onpopstate
-        instrument(_window, 'onpopstate', function (original) {
-          return function () {
+        if (typeof addEventListener === 'function') {
+          addEventListener('popstate', (_event) => {
             recordUrlChange(lastHref, _window.location.href)
-            if (original) {
-              return original.apply(this, arguments)
-            }
-          }
-        })
+          })
+        }
 
         // https://developer.mozilla.org/en-US/docs/Web/API/History/pushState
         // https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState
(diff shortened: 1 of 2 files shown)


packages/js/src/browser/integrations/breadcrumbs.ts
```diff
@@ -247,15 +247,11 @@ export default function (_window = globalThisOrWindow()): Types.Plugin {
           })
         }
 
-        // https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onpopstate
-        instrument(_window, 'onpopstate', function (original) {
-          return function () {
+        if (typeof addEventListener === 'function') {
+          addEventListener('popstate', (_event) => {
```
subzero10, line 251: I think this is a better approach. `addEventListener` does not replace the function.

joshuap (approved):
LGTM, thanks for the quick fix!

subzero10:
> LGTM, thanks for the quick fix!

Well, you did most of the work 😉

Issue #992, opened 2023-01-04
Vite Plugin for Source Maps


subzero10:
Is there any documentation / plugin to upload source maps with Vite instead of webpack? I’m unsure how to use the GitHub action with Vite (unsure where source maps are uploaded, the minified url, minified file, etc).

Any help would be appreciated. Using Vue3 + Vite.

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_go3tt)

subzero10:
Note:
- It should be under this repo (as a package of the monorepo)
- It should be published on npm with the name `@honeybadger-io/vite-plugin`

BethanyBerkowitz:
@subzero10 I'm tempted to try https://github.com/unjs/unplugin to see if we can get rollup and esbuild as well, and have only one main piece of plugin code. I've never used it before so I'm not sure how easy/difficult it will be to make the existing webpack plugin code work for all these different options. 

If not, I'd still consider building a rollup plugin and then seeing if we can just use it for Vite, since on a first pass through the documentation I think our use case fits their criteria for rollup compatibility: https://vitejs.dev/guide/api-plugin.html#rollup-plugin-compatibility

subzero10:
That's a great idea, @KonnorRogers mentioned `unplugin` about a month ago!

I think you should at least give it a go, probably make it a time-bound exercise if you are concerned that it might be too difficult.

BethanyBerkowitz:
Ok, so I started playing around with `unplugin`... here's where I got to: https://github.com/honeybadger-io/honeybadger-js/compare/BKB/unplugin?expand=1

I started by just trying to re-create the basics of the existing Webpack plugin, along with a small webpack example project using it. The problem I ran into is that our existing Webpack plugin relies on getting the bundle with the file names in it in order to upload the sourcemaps. However, for some reason `unplugin` decided not to pass the args that normally go along with the rollup `writeBundle` hook... I have no idea why. They hook into the timing correctly, but just don't pass the args, so we don't have access to the bundle. 

It should be possible to re-write how we locate the sourcemaps, rather than just getting them directly from the bundle. However, I'm inclined to take a pass at just writing a rollup plugin directly, since unplugin doesn't seem to be as convenient as I expected. 

Very open to suggestions if anyone has a better idea for getting the sourcemap files without the `writeBundle` args though!

[...]

PR #1006, opened 2023-01-26
feat(rollup-plugin): add rollup/vite plugin that uploads sourcemaps to honeybadger API

BethanyBerkowitz:
## Status
WIP

## Description
This PR introduces a new rollup/vite plugin to upload sourcemaps to Honeybadger. Addresses https://github.com/honeybadger-io/honeybadger-js/issues/992. I have tested that it uploads sourcemaps and that those sourcemaps are applied correctly in the Honeybadger UI. 

It is not yet feature-complete as compared to the webpack plugin. The main functionality missing is sending deploy notifications. The webpack plugin also has configurable parallelism, which this does not currently have.

We also may want to add a check to make sure we're only uploading sourcemaps on production builds. 

## Todos
- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce
There are two example projects added here. One uses rollup to package some simple JS, and the other is a Vite/Vue setup. To test:
* run `npm run build` in the plugin folder to build the plugin
* run `npm run build` in the example folder to build the example project
If you haven't configured your API key correctly, you should get a nice, useful error. 
Configure your env variables using `.env.local`. Each example project includes a `.env.example` you can copy from.


packages/react-native/index.d.ts
```diff
@@ -52,10 +52,10 @@ declare namespace honeybadger {
        * Sets the logging level for the Honeybadger library.
        * @param {('debug'|'warning'|'error')} [level=warning] - The logging level.
        */
-    function setLogLevel(level?: "error" | "warning" | "debug"): void;
```
BethanyBerkowitz, line 55: Just from eslint, not sure why this wasn't caught elsewhere.

packages/rollup-plugin/src/index.js
```diff
@@ -0,0 +1,15 @@
...
+import { uploadSourcemaps } from './hbUtils.js'
+
+export default function honeybadgerRollupPlugin(options) {
+  const hbOptions = cleanOptions(options)
+
+  return {
+    name: 'honeybadger', 
+    writeBundle: async (outputOptions, bundle) => {
+      const sourcemapData = extractSourcemapDataFromBundle({ outputOptions, bundle })
+      await uploadSourcemaps({ sourcemapData, hbOptions })
```
subzero10, line 12: Shouldn't we report success (and possibly some stats, like number of files uploaded)?

packages/rollup-plugin/src/hbUtils.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  const sourcemapUploadPromises = sourcemapData.map(data => {
+    return uploadSourcemap({ 
+      ...hbOptions,
+      ...data
+    })
+  })
+  
+  return Promise.all(sourcemapUploadPromises)
```
subzero10, line 31: I'm wondering if we can use [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) here.

packages/rollup-plugin/package.json
```diff
@@ -0,0 +1,44 @@
+{
```
subzero10, line 1: Should we specify the files to published on npm here? This is what we do with the other packages (we publish only `dist` folder and type definitions).

subzero10 (commented):
This looks great and very thorough!
I like the inclusion of examples and there's good coverage on the tests!

I have some general notes/feedback:
- Can you please add an `MIT-LICENSE` file in the plugin folder? You can take one from the other packages.
- With the js monorepo, we've tried to adopt Typescript as much as we can, to gain static type checking and easier maintainability. I can see that you are already using jsDoc to achieve similar value, but I think Typescript is better. What are your thoughts?
- Is there a reason you went for mocha/chai and not jest? I'm just curious, is it because our webpack plugin is using mocha as well? (I wanted to switch all packages to one testing suite, again to improve maintainability, but didn't get the time to do that yet)

subzero10:
> The main functionality missing is sending deploy notifications. The webpack plugin also has configurable parallelism, which this does not currently have.

As soon as this PR ready, we can publish it and create github issues to implement deploy notifications and parallelism 👌 .

subzero10:
> We also may want to add a check to make sure we're only uploading source-maps on production builds.

+1. I think we should have this from the first version.
Source-maps on non-production builds can become tricky to address and most of the times they are unnecessary.

BethanyBerkowitz:
> * Can you please add an `MIT-LICENSE` file in the plugin folder? You can take one from the other packages.
Yes, good catch, thanks! 

> * With the js monorepo, we've tried to adopt Typescript as much as we can, to gain static type checking and easier maintainability. I can see that you are already using jsDoc to achieve similar value, but I think Typescript is better. What are your thoughts?
I'm happy to use Typescript if we're settling on that! TS definitely helps avoid some categories of bugs. 

> * Is there a reason you went for mocha/chai and not jest? I'm just curious, is it because our webpack plugin is using mocha as well? (I wanted to switch all packages to one testing suite, again to improve maintainability, but didn't get the time to do that yet)
I saw we had a mixed bag.. some with mocha/chai and some with jest. I just picked mocha/chai since it was in the webpack repo. Happy to switch if you want to settle on jest? I don't have a preference for one vs the other.

packages/rollup-plugin/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+```
+# npm
+npm install @honeybadger-io/rollup-plugin --save-dev
+
+# yarn
+yarn add @honeybadger-io/rollup-plugin --dev
+```
+
+### rollup.config.js
```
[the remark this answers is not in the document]
BethanyBerkowitz: I went back and forth on that... I think this info is more "basic setup" kind of info than the detailed config so I'd prefer to read it first if I were setting it up myself. But either way makes sense to me, so I'm happy to just match the layout that the webpack one has.

packages/rollup-plugin/src/hbUtils.js
BethanyBerkowitz: Good point! I wasn't sure if we want the failure of an upload to cause the user's build to fail, or if we want to just log to the console any failures. Current behavior will cause the build to fail if any of the uploads fail. `allSettled` would allow the build to pass, and then we could just log out what happened. I could see an argument for either behavior... having the builds fail is actually pretty helpful when you're first setting it up because it's super obvious that something's broken. But if your CI started failing just because of sourcemap upload, it would probably be pretty annoying.

BethanyBerkowitz:
@subzero10 I had a go at switching to typescript. I was able to convert the code itself to TS pretty easily. However, I could not get the tests running after the conversion. 

You can see my progress on this branch: https://github.com/honeybadger-io/honeybadger-js/compare/BKB/rollup-plugin...BKB/rollup-typescript?expand=1

Here's what I understand of the situation so far: 

1. I set this code up as an esm module because `node-fetch` is now esm-only. Because of this, I ended up using `testdouble` for mocks and dependency injection. `testdouble` makes dependency injection work by using a node loader like `mocha --loader testdouble` (loaders are still technically experimental, but since it's just for running tests I wasn't too worried about that). 
2. To get mocha to play nice with typescript, I needed to use `ts-node`. Getting this to work required adding a loader like `mocha --loader ts-node/esm`. 
3. Multiple loaders is a super recent addition to node, I don't think I can reliably get both loaders working at once without a lot of effort.
4. Jest does a better job with typescript than mocha, but I don't think switching to Jest would work well with the testdouble esm functionality, based on [this blog post](https://blog.testdouble.com/posts/2021-10-29-testing-esm-with-mocha-and-testdouble/). 

There are a couple of ways I can think of to address this: 

1. Go forward with the existing code without adding typescript, circle back to typescript later. 
2. Re-work the code to avoid being esm-only. This would mean I'd have to either downgrade the node-fetch major version from 3 to 2 and lose some features I'm depending on. I think I'd need to rewrite at least generating the body of the request and some of the testing. 

What do you think? Any other solutions I'm missing here?

packages/rollup-plugin/src/options.js
```diff
@@ -0,0 +1,40 @@
+export const MAX_RETRIES = 10
+export const DEFAULT_RETRIES = 3
+export const DEFAULT_ENDPOINT = 'https://api.honeybadger.io/v1/source_maps'
+export const DEFAULT_REVISION = 'master'
```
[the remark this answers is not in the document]
subzero10: It's already done, but not deployed yet! We have a [v5 milestone](https://github.com/honeybadger-io/honeybadger-js/milestone/3?closed=1) that should be going out soon.

packages/rollup-plugin/src/hbUtils.js
subzero10: Interesting points 🤔 . I think we can let the build fail if the source map file upload fails.
To me, this is an important part of the deployment; otherwise reported errors with the new release will not show up as expected, and it might take time until people realise that the upload is failing.

However, my original suggestion for `Promise.allSetlled` is whether we want to show more detailed error messages. For example, `Promise.all` will reject with the first promise that failed, but we won't know the state of the other promises that haven't resolved/rejected yet. It may be unnecessary information, though. Like, is there any additional value if we tell "these 10 source map file uploads failed" or just that "source map files upload failed" ?

subzero10:
Hey @BethanyBerkowitz, first of all let me say that I apologize for having you go through all this trouble, even with all this amazing work you've done so far! I should have listed typescript as a requirement from the beginning.

> I set this code up as an esm module because node-fetch is now esm-only.

This might be a problem, since our current stand is that we support node.js as low as v12. ES Modules weren't native to Node.js until v14. However, we can choose to release this plugin with our [v5 milestone](https://github.com/honeybadger-io/honeybadger-js/milestone/3), which drops support of Node v12 and v13. Here's what you can try:
- switch to `releases/v5` branch, rebase `master`
- push the rollup plugin changes in that branch (merge/rebase/cherry-pick). The important bit is that we have one commit with this name `feat(rollup-plugin): add rollup/vite plugin that uploads sourcemaps to honeybadger API` (the rest of the commits can be anything **but** `feat`. This is because we want the commit to be picked up automatically by lerna and conventional changelog). Feel free to reach out to me if this is not clear!
- let the CI run and see if it passes. You can ignore the _Check PR title_ step. That's expected to fail.

> There are a couple of ways I can think of to address this:
>
> 1. Go forward with the existing code without adding typescript, circle back to typescript later.
> 2. Re-work the code to avoid being esm-only. This would mean I'd have to either downgrade the node-fetch major version > from 3 to 2 and lose some features I'm depending on. I think I'd need to rewrite at least generating the body of the request > and some of the testing.
>
> What do you think? Any other solutions I'm missing here?

Considering the current status, I would say that we can go ahead with the existing code (assuming all the PR review comments are resolved):
- the code is clean and easy to read
- the code is thoroughly tested
- documentation is detailed and easy to read
- there are examples
- people can start using the plugin while we work on our internal conventions (use Typescript)

The only blocker here for me is the CI. If we get that resolved by pushing in the v5 milestone, then we can go ahead and release it with v5!

When we release, I suggest we create an issue to move to typescript.

packages/rollup-plugin/src/hbUtils.js
BethanyBerkowitz: Just added logging for the number of successes, and re-threw the failures in order to cause a failure of the build!

packages/rollup-plugin/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "name": "@honeybadger-io/rollup-plugin",
+  "version": "1.0.0",
+  "description": "A rollup plugin to upload sourcemaps to Honeybadger",
+  "main": "index.js",
```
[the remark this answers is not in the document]
BethanyBerkowitz: I think you're right! Removed.

packages/rollup-plugin/package.json
BethanyBerkowitz: Just added a 'files' key like I saw here: https://github.com/honeybadger-io/honeybadger-js/blob/c6970426f1c38b6dd07c1d355fc20bd2730521c8/packages/js/package.json#L32

packages/rollup-plugin/src/index.js
BethanyBerkowitz: Added!

packages/rollup-plugin/src/hbUtils.js
```diff
@@ -0,0 +1,132 @@
...
+      if (body && body.error) {
+        details = `${res.status} - ${body.error}`
+      } else {
+        details = `${res.status} - ${res.statusText}`
+      }
+    } catch (parseErr) {
+      details = `${res.status} - ${res.statusText}`
+    }
+
+    throw new Error(`Failed to upload sourcemap ${sourcemapFilename} to Honeybadger: ${details}`)
```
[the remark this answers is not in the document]
BethanyBerkowitz: If we get a parse error, it's probably because the body of the response that came back was not JSON. So I think the contents of the parse error are not going to be that interesting... something like "SyntaxError: Unexpected end of JSON input". If this happens, I think the most important data the user needs is just the response status and statusText.

BethanyBerkowitz:
@subzero10 Ok, I moved this feature into the `releases/v5` branch like you recommended and got CI passing with a few minor changes. Please double-check that my commits are ok for the Lerna setup when you get a chance. 

In the meanwhile I'm going to go ahead and create some follow-up issues based on the notes above.

packages/rollup-plugin/src/hbUtils.js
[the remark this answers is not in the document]
BethanyBerkowitz: Aaaand turns out that `error.cause` wasn't introduced until node 16, so I just cut it and added additional info into the `message`.

subzero10:
> @subzero10 Ok, I moved this feature into the `releases/v5` branch like you recommended and got CI passing with a few minor changes. Please double-check that my commits are ok for the Lerna setup when you get a chance.
> 
> In the meanwhile I'm going to go ahead and create some follow-up issues based on the notes above.

All looks good! We can go ahead with the release as soon as I get a PR approval (I am the author of the PR so I can't do it)!

PR #1019, opened 2023-02-13
Update peer dependency @honeybadger-io/js to 5.1.1

kelkes:
I guess that was missed with the latest 5.x.x release. npm install failed without the force option.

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -23,7 +23,7 @@
     "deploy": "gh-pages -d example/build"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


subzero10:
Thanks @kelkes! I was going to approve and merge, but I just realized that we missed the same thing for the react package as well. I created this [PR](https://github.com/honeybadger-io/honeybadger-js/pull/1021) that takes care of both packages.

kelkes:
Thanks @subzero10

Issue #1020, opened 2023-02-13
Version 5.1.1 is broken on Next.js


bymoe:
## What are the steps to reproduce this issue?
1. Install both `@honeybadger-io/react` & `@honeybadger-io/js` @ v `5.1.1`
2. Try to deploy to something like vercel or even locally
3. You'll get `TypeError: _window.addEventListener is not a function` 

## What happens?
I am getting `TypeError: _window.addEventListener is not a function` after upgrading to v5

## What were you expecting to happen?
To support SSR? Not sure why it was working fine before.

## Any logs, error output, etc?
`TypeError: _window.addEventListener is not a function`

## What versions are you using?
**Operating System:** Mac
**Package Name:** `@honeybadger-io/react` & `@honeybadger-io/js` @ v `5.1.1`
**Package Version:** 5.1.1

![Screenshot from 2023-02-13 20-37-40](https://user-images.githubusercontent.com/20565199/218531279-5c0d753b-c8c8-4c49-9f4f-a0615fea6000.png)

bymoe:
Thanks @subzero10 :+1: 

Also another thing I'd like to mention is the docs are not very clear about which is the best way to integrate or even to upgrade from version to version. 
On the previous version `@honeybadger-io/react": "^2.0.1` it took me 5min to setup. Now it just for some reason 10x more complicated.
The docs for example have multiple integration ways:
e.g. latest [React docs](https://docs.honeybadger.io/lib/javascript/integration/react/) tells you to install `@honeybadger-io/js @honeybadger-io/react`. But the [Next.js example](https://github.com/honeybadger-io/nextjs-with-honeybadger) tells to install `"@honeybadger-io/js"  "@honeybadger-io/webpack"` is the  `@honeybadger-io/react` deprecated ? Since nothing indicates that in the [package readme](https://www.npmjs.com/package/@honeybadger-io/react). Is Next.js not considered part of "React" docs? Just not sure, It's a bit confusing! 

Thanks!

subzero10:
Hey @bymoe, apologies for all the trouble caused!

> Also another thing I'd like to mention is the docs are not very clear about which is the best way to integrate or even to upgrade from version to version.

We use lerna versioning and keep all the versions aligned, so sometimes a package may get a major version bump, but it doesn't necessarily mean that there are breaking changes. In any case, I've taken note of this and we will more communicative with version upgrades.

> The docs for example have multiple integration ways:
[...]

PR #1024, opened 2023-02-17
fix(react): import correctly @honeybadger-io/js

subzero10:
Fixes: #1020

## Status
**READY**

## Description
This was a brain teaser 😓 .
This PR fixes a bug I introduced with the release of the Collect User Feedback feature. The bug was introduced because I couldn't run the tests 🤦 .

### Some context:
- With the "Collect User Feedback", I added a new property in the react component to show the form automatically in case of an error (`showUserFeedbackForm`)
- I created a test for it, but it was failing with HTTP 403, even though I tried mocking the fetch api. I realised that it was importing the _server_ implementation of Honeybadger (as it is the main entry point of the `@honeybadger-io/js` package.
- So, since we always need the client side implementation, I changed the import to be explicit and imported `@honeybadger-io/js/dist/browser`.
- Hooray! My tests passed!

### This is what I missed:
- Some frameworks use SSR, and in this case importing the browser implementation will create errors; and this is what happens with the next.js app.

### The fix:
- Always import `@honeybadger-io/js` and let bundlers decide which entry point to use depending on the environment. I tested this in @BethanyBerkowitz's [repo](https://github.com/BethanyBerkowitz/hb-nextjs-react-test) by adding a line in `_app.js`:
```
// honeybadger.errorHandler should exist only on server implementation
console.log(typeof window === 'undefined' ? 'server' : 'browser', honeybadger.errorHandler) 
```
When running `npm run dev`, I get this on the terminal:
```
server [Function: bound errorHandler]
```
and this on the browser console:
```
browser undefined
```
Which indicates that everything is imported correctly.
- Fixing the tests: jest supports a `resolver` config which can be used for these cases (to make it respect the browser field in package.json). However, it's not supported in react-scripts so I explicitly import the browser implementation.

### DIFF
--- a/packages/react/src/HoneybadgerErrorBoundary.tsx
+++ b/packages/react/src/HoneybadgerErrorBoundary.tsx
@@ -1,5 +1,5 @@
 import React, { Component, ReactNode } from 'react'
-import Honeybadger from '@honeybadger-io/js/dist/browser/honeybadger'
+import Honeybadger from '@honeybadger-io/js'
 import DefaultErrorComponent, { DefaultErrorComponentProps } from './DefaultErrorComponent'
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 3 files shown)


BethanyBerkowitz (approved):
This was a tricky one! Glad you were able to find it.

Issue #1010, opened 2023-02-02
Convert rollup-plugin to typescript


BethanyBerkowitz:
Our rollup plugin was written in Javascript, and may be a little bit of a pain to convert to Typescript. The code itself is easy to convert, but the testing setup was painful. 

See https://github.com/honeybadger-io/honeybadger-js/pull/1006#issuecomment-1410897881

BethanyBerkowitz:
I remain curious about this problem, so I tried a few ideas: 

1. Write a custom loader that just combines the `testdouble` loader and `ts-node/esm` loader. Chaining loaders is available in node 18.6.0, but we want to be compatible back to node 14. Writing a loader was useful for learning about how loaders work, but ultimately I got some errors related to `quibble` not being set up correctly and it seemed like more trouble than it was worth, especially for a solution that's a bit hacky and undocumented. 
2. Transpile to JS first (`npx tsc -p tsconfig.test.json`, for example), then run mocha tests against the JS. This seems like it will work, and I think it would probably be fine to add a pretest step, even if it's not the standard way people usually do it. 

@subzero10 If you think adding a transpile step before testing is ok, I can go ahead with converting to TS. Or if you have any better ideas!

subzero10:
Hey @BethanyBerkowitz,
I considered suggesting transpiling earlier, but the major drawback for me is the generated js files that we'd have to cleanup every time we finish testing (we don't want to commit those). Plus, testing with the IDE will require additional configuration but I guess that's OK.

Have you tried importing testdouble in your test files instead of using the loader?
```
import * as td from 'testdouble';
```

BethanyBerkowitz:
> Hey @BethanyBerkowitz, I considered suggesting transpiling earlier, but the major drawback for me is the generated js files that we'd have to cleanup every time we finish testing (we don't want to commit those). Plus, testing with the IDE will require additional configuration but I guess that's OK.

If we put the generated files into a specific folder, we can gitignore them. But I agree that it's not an elegant solution!

>
> Have you tried importing testdouble in your test files instead of using the loader?
> 

Unfortunately in order to mock out imported esm modules, I need to use the loader. And I need to mock imported modules, otherwise our tests will make real API calls. Ultimately I think this comes down to the fact that Node made it sound like esm modules are fully operational and ready to go, so some tools like `node-fetch` went ahead and became esm-only, but actually the tooling is not very good yet 😬 

[...]

PR #1027, opened 2023-02-22
chore(rollup): downgrade from esm-only and use typescript

BethanyBerkowitz:
## Status
WIP

## Description
Closes https://github.com/honeybadger-io/honeybadger-js/issues/1010

In order to deal with issues related to node-fetch being esm-only (see https://github.com/honeybadger-io/honeybadger-js/pull/1006#issuecomment-1410897881), I downgraded node-fetch to version 2. They say they will continue to maintain v2. Downgrading meant I lost some built-in functionality, so I needed to introduce the `form-data` package and change how files were uploaded. 

Removing the esm-only setup allowed me to get Typescript happily working with Mocha. 

From our users' perspective, I don't think there are any changes here. If they were using esm, they still can. 

## Todos
- [x] Tests
- [ ] Documentation

## Steps to Test or Reproduce
1. `npm run build:clean` in the rollup package to build the plugin
2. Test the plugin in the two example projects by running `npm run build`. You may need to set up your API key and assetsUrl.


BethanyBerkowitz:
@subzero10 I have an outstanding question on this PR, but I thought it was ready for you to take a look at anyway. 

The rollup example now seems to want to use the cjs version of the build, rather than the es version of the build. I don't know why, since the config file is `.mjs`. I need to do a bit more testing on this.

subzero10:
Hey @BethanyBerkowitz, I just tried your branch and I found a way to resolve the error you are getting above:
- open the rollup example folder
- do `npm i ../../` -> this will create a link to the local plugin package (if you make changes to it and rebuilt, you don't have to re-install)
- import the rollup plugin with `import honeybadgerRollupPlugin from '@honeybadger-io/rollup-plugin'` -> you can verify that the `es` version is being imported by going to `node_modules/@honeybadger-io/rollup-plugin/dist` and removing the `cjs` folder

I don't really know why we have to do that, but my guess would be that node confuses which _package.json_ to load when importing a file that is outside the package root.

BethanyBerkowitz:
@subzero10 thanks for this! Seems to work great. I changed both examples to import this way, which is more in line with how users will do it anyway. 

I think this is ready for a full review, now that that issue is resolved.

packages/rollup-plugin/src/hbUtils.ts
```diff
@@ -0,0 +1,176 @@
+import originalFetch from 'node-fetch'
+import FormData from 'form-data'
+import { promises as fs } from 'fs'
+import fetchRetry from 'fetch-retry'
+// @ts-ignore
```
subzero10, line 5: I assume you had to add this because there's a type mismatch between the packages.
I would recommend replacing with [`@ts-expect-error`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments). This way, you are telling typescript that you expect this to error, but if it is resolved in future package updates, it will give a warning message and you can remove this line.
```suggestion
// @ts-expect-error
```

packages/rollup-plugin/src/hbUtils.ts
```diff
@@ -0,0 +1,176 @@
...
+  hbOptions: HbPluginOptions
+): Promise<Response> {
+  const body = await buildBodyForSourcemapUpload(sourcemapData, hbOptions)
+  
+  let res: Response
+
+  try {
+    res = await fetch(hbOptions.endpoint, {
+      method: 'POST',
+      // @ts-ignore
```
subzero10, line 57: Same as above.

subzero10 (approved):
This looks good to me! I left a couple of comments, but obviously not blocking.

Finally, I have one more thing to note, that we should be aware of:

You are doing the typescript transpilation with rollup, which is convenient and the expected approach. But the typescript plugin of rollup does not support [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html). We use this feature in our monorepo. This would be necessary if the rollup plugin had dependency on other packages of the monorepo. 
So, something to keep in mind when we decide to create a shared package for our webpack and rollup plugins. We would have to adjust the build to be as it is in the other packages. I wouldn't ask you to do it now though, we spent enough time refactoring this package already :)

PR #1028, opened 2023-02-24
chore(react-native): new example projects to test on

BethanyBerkowitz:
## Status
READY

## Description
In getting ready to work on react-native sourcemap integration, I tried to set up the example project. I ran into install issues that I could not easily resolve. Instead, I just made 2 new example projects (one with expo and one without) that we can use to test against. 

## Todos
- [ ] Tests
- [x] Documentation

## Steps to Test or Reproduce
See README update in this PR for instructions


subzero10 (changes requested):
No objection from my side for the new example projects! I've never tried running our react-native project (yet!).

Can you please make sure the `examples` folder is excluded from the _npm_ release? (either with an `.npmignore` or the `files` property in `package.json`)

BethanyBerkowitz:
> Can you please make sure the `examples` folder is excluded from the _npm_ release? (either with an `.npmignore` or the `files` property in `package.json`)

Great catch, thank you! I included most of the .gitignore file in the .npmignore because I saw that by [default if you don't include `.npmignore`, npm ignores whatever is in `.gitignore`](https://docs.npmjs.com/cli/v9/using-npm/developers#keeping-files-out-of-your-package).

BethanyBerkowitz:
Also wanted to point out that we need to update the documentation page if this PR is accepted. I didn't want to do it yet in case @joshuap has any reason why making new example projects for this is not a good idea. 

https://docs.honeybadger.io/lib/react-native/#example-project

joshuap (approved):
I have two issues w/ example projects in general, neither of which should stop you from making one. 😁 

1. They tend to become stale quickly (so they should avoid boilerplate if possible, but I know that's not always feasible)
2. They can stand in place of good docs

My ultimate goal is to improve the docs to the point where the example projects are no longer necessary (or [easily reproducible](https://www.notion.so/honeybadger/549be6376a6246e19d6291bf8bdcee69?v=02c2f9dfbe41498ebd8f43880fb4e782&pvs=4)). That said, I think having some examples in this case would help both us and our customers figure out what should actually go in the docs in the first place. 😅

BethanyBerkowitz:
> I have two issues w/ example projects in general, neither of which should stop you from making one. 😁
> 
> 1. They tend to become stale quickly (so they should avoid boilerplate if possible, but I know that's not always feasible)
> 2. They can stand in place of good docs
> 
> My ultimate goal is to improve the docs to the point where the example projects are no longer necessary (or [easily reproducible](https://www.notion.so/honeybadger/549be6376a6246e19d6291bf8bdcee69?v=02c2f9dfbe41498ebd8f43880fb4e782&pvs=4)). That said, I think having some examples in this case would help both us and our customers figure out what should actually go in the docs in the first place. 😅

I see the argument about staleness, especially since I needed to create new examples since the existing one was stale! On the other hand, we need examples to develop against.. so the question is just whether to check them in or not. I lean toward checking them in to make the dev workflow easier, but could definitely be convinced otherwise.

joshuap:
I'm good with checking them in!

BethanyBerkowitz:
@subzero10 I need your re-review on this -- I think I addressed your concern re: keeping examples out of the package. I'm still fairly stuck on getting sourcemaps working, or even getting a reasonable backtrace showing up in the HB UI, but I still think these are an improvement over what was there before.

subzero10 (approved):

Issue #1023, opened 2023-02-17
Next.js dedicated package


subzero10:
## Description

Create `@honeybadger-io/nextjs` package that encapsulates all necessary configuration to have honeybadger reporting errors both on server (`@honeybadger-io/js`) and client side (`@honeybadger-io/react`). This package should simplify setting up a project to use with nextjs, as described [here](https://github.com/honeybadger-io/nextjs-with-honeybadger/blob/master/pages/_app.js).

## Todo/Requirements:

- Typescript
- Tests
- Allow separate config for client vs server side
- Default config should: setup client side, server side, edge(?) and upload source maps
- [Dedicated docs page](https://github.com/honeybadger-io/docs/issues/190)

PR #1030, opened 2023-03-06
feat: nextjs integration

subzero10:
## Status
**READY**

## Description
Closes: #1023
A nextjs dedicated package that simplifies integration with nextjs:
- server integration
- client integration
- source map upload

## Todos
- [x] Publish alpha version of the package so it can be installed on a Vercel deployment
- [x] Webpack config to configure Honeybadger based on runtime
- [x] Honeybadger Edge API client (Server client may not work) - made it work with the browser client
- [x] Source map upload with webpack plugin
- [ ] Tests
- [x] README.md
- [x] [Documentation](https://github.com/honeybadger-io/docs/issues/190) - include known limitations
- [x] cli command to generate config files
- [x] Manual Testing with deployments to Vercel
- [ ] Manual Testing with deployments to Heroku
- [ ] Manual Testing with deployments to self-hosted server (?)
- [x] Add @honeybadger-io/nextjs in [pull-package-versions workflow](https://github.com/honeybadger-io/.github/blob/main/.github/workflows/pull-package-versions.yml)
- [x] Create issues on github to address limitations (#1055, #1056)
- [x] Verify that HoneybadgerErrorBoundary React component provides additional React contextual information
- [ ] [Update nextjs example repo to use this package](https://github.com/honeybadger-io/nextjs-with-honeybadger/issues/12)


subzero10:
Just realized that I will have to find a way to publish an alpha version of the package without Lerna. 
Our current Lerna setup uses fixed mode, which uses same versioning strategy for all packages and I don't want to change that at the moment.
It should be easily done with the npm cli.

subzero10:
I'm sad to realize that it's no easy task to install a package from the monorepo without an npm publish.
I came across [gitpkg](https://gitpkg.vercel.app/), but it's not working for me. They have an [open issue](https://github.com/EqualMa/gitpkg/issues/25#issuecomment-1468161033) and I commented.

subzero10:
Update:

- The HB client can now report errors either from server, browser or vercel edge.
- However, it won't always work because of some strange caveats of Nextjs/Vercel. For example, [this](https://nextjs.org/docs/advanced-features/custom-error-page#caveats) one makes it difficult to report an error that comes from an error thrown in an api route, because Nextjs renders a 404 status code 🤷 . Additionally, there is no information about the error at all. As a result, a different way is to be implemented to handle those cases, which I plan to do with a separate issue/PR.

I will pause here as far as "the reporting to Honeybadger" task goes and focus on the next points in the todo list above. I will document known limitations, so we can create issues to improve them.

subzero10:
Update:
I integrated the webpack plugin so now source maps are automatically uploaded. They are working for server and browser runtimes. **They are not working for the vercel edge runtime**. 
I haven't decided if I should work on this or if I should focus on getting the first version of this package out there (remaining todo items); I'm leaning towards the latter.

subzero10:
Hey @BethanyBerkowitz, @joshuap  and @stympy,

I think this is ready for review. I have some more testing to do, but you should be able to get started with the review.

subzero10:
I will merge this so that the package will be published on npm. Then, I will be able to do end-to-end testing with our example repository.

PR #1034, opened 2023-03-08
chore(react-native): Minor updates to example projects 

BethanyBerkowitz:
## Status
READY

## Description
Recording updates I've made that make it easier to use the test projects. 

Also removing an unused plugin (react-native-dotenv) that was causing an error. I'd uninstalled the dependency but forgot to remove the plugin. 

## Todos
- [ ] Tests
- [ ] Documentation

## Steps to Test or Reproduce
Run the example projects -- you should see the ability to configure your revision, and no unexpected errors.


packages/react-native/examples/react-native-cli/ios/AwesomeProject.xcodeproj/project.pbxproj
```diff
@@ -498,7 +499,8 @@
 					"-ObjC",
 					"-lc++",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = "org.reactjs.native.example.$(PRODUCT_NAME:rfc1034identifier)";
+				PRODUCT_BUNDLE_IDENTIFIER = "honeybadger.react-native-cli.example.--PRODUCT-NAME-rfc1034identifier-";
```
BethanyBerkowitz, line 502: This is an auto-generated file... I'm not totally sure why this changed but it seems to work fine.

subzero10 (approved):

Issue #926, opened 2021-04-28
Depend on honeybadger-js or core


joshuap:
A while back we [rebuilt our honeybadger.js JavaScript client](https://www.honeybadger.io/blog/universal-honeybadger-js/) as a universal JavaScript client to support both browser JS and Node. Currently honeybadger-react-native implements its own limited JavaScript client that is used only in React Native projects.

Would it be possible to integrate with honeybadger.js so that we can share the core modules and features it provides? For an example of how this might be done, see the server and browser implementations [here](https://github.com/honeybadger-io/honeybadger-js/tree/master/src).

PR #1049, opened 2023-03-22
feat(core): accept a backtrace passed in the notice

BethanyBerkowitz:
## Status
READY

## Description
In working on https://github.com/honeybadger-io/honeybadger-js/issues/926, I found it would be helpful if the core package would accept a backtrace passed in, rather than exclusively either parse the `notice.stack` string into a backtrace array or create its own backtrace. The existing react-native package has custom code to make nice backtraces for native iOS and android errors, which are not formatted the same as javascript errors, so they wouldn't be parsed correctly from `notice.stack`. 

## Todos
- [x] Tests
- [ ] Documentation

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -314,12 +314,15 @@ export abstract class Client {
       tags: uniqueTags,
     })
 
-    if (typeof notice.stack !== 'string' || !notice.stack.trim()) {
-      notice.stack = generateStackTrace()
-      notice.backtrace = makeBacktrace(notice.stack, true, this.logger)
-    }
-    else {
-      notice.backtrace = makeBacktrace(notice.stack, false, this.logger)
+    // If we're passed a custom backtrace array, use it
+    // Otherwise we make one. 
+    if (!Array.isArray(notice.backtrace) || !notice.backtrace.length) {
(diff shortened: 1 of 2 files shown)


BethanyBerkowitz:
@subzero10 @joshuap Since this is part of the core package (eek!) I want to check with both of you that this seems safe to do. I don't think we'd generally want users to pass in a custom backtrace like this, it would really just be for our other packages to use. The [current documentation](https://docs.honeybadger.io/lib/javascript/guides/reporting-errors/#additional-options) doesn't indicate that you can or should pass in a backtrace, and I don't see a reason to change that.

subzero10 (approved):
I can't think of anything against this.
I think it's OK to accept customization at this level, since this is the _core_ package.
Maybe an alternative would be to make the `makeNotice()` method abstract, and have each client come up with its implementation, but that's not a good idea at this point: all current clients would duplicate almost all of this logic (browser, node and now react-native).

joshuap (approved):
Yeah, this seems fine to me.

We do expose `notice.backtrace` as a [notice property](https://docs.honeybadger.io/lib/javascript/reference/configuration/#notice-properties) already, and users can modify it in `beforeNotify` handlers (e.g. to optimize grouping). So it's a far cry to allow the property to be passed in—although I don't think we should recommend it to end-users.

Issue #1050, opened 2023-03-23
Android error is not caught and reported


BethanyBerkowitz:
Native android errors are being caught and reported to Honeybadger when thrown [this delayed manner](https://github.com/honeybadger-io/honeybadger-js/blob/f9f596e82c7ee3a9a7df74a10703c699314707a9/packages/react-native/examples/react-native-cli/android/app/src/main/java/com/awesomeproject/ThrowErrModule.java#L25), but not when [thrown directly](https://github.com/honeybadger-io/honeybadger-js/blob/f9f596e82c7ee3a9a7df74a10703c699314707a9/packages/react-native/examples/react-native-cli/android/app/src/main/java/com/awesomeproject/ThrowErrModule.java#L37). 

It appears that this directly thrown error never makes it to the javascript listener on `native-exception-event`, whereas the delayed error does. 

We theorize that the “direct” error crashes the app/process, and that the Android OS won’t let you report the error. In this case, the only way to work around that is by storing the exception info (i.e. persist on device) and reporting it to honeybadger when the app starts back up again. This is the experience @subzero10 had with Firebase Crashlytics, and it seems that this is still the case, as you can see [here](https://firebase.google.com/docs/crashlytics/test-implementation?authuser=3&platform=android). Andrey found that this is the case with many of the other providers as well. 

If we take the approach of persisting the error locally and then reporting it on next launch, care must be taken to make we don't make assumptions about when the exception took place via when it’s being reported. It seems like a non-optimal approach, but it is a better alternative to either not sending it or pushing it into a background thread and hoping it is kept in memory long enough to send.

andreybutov:
TLDR: This isn't so much a bug, as a limitation of the way React Native integrates itself into Android. All of the other crash handers that I tested (Rollbar, Bugsnag, Sentry) also contain this issue, and none of them figured out a way to address it properly. We can't resolve it definitively, but we can implement some additional functionality to minify the pain.

OK. An explanation.

For a React Native application running on Android, we have three "origins" for exceptions. 

1. Errors thrown from JavaScript.
2. Errors thrown from native Android code, on the main thread.
3. Errors thrown from native Android code on a background thread.

1 and 3 are working fine. But 2, where errors are thrown by native Android code, on the main thread are never reported.

The reason they aren't reported is because the only avenue we have, as a third-party library, to get a foothold into a exceptions is to set up a global handler for uncaught exceptions.

Which is what we do.

[...]

PR #1051, opened 2023-03-23
chore(react-native): throw native errors in react-native example project

BethanyBerkowitz:
## Status
READY

## Description
This update to the example project is something I'm using in my branch while testing #926. It is also useful for documenting https://github.com/honeybadger-io/honeybadger-js/issues/1050, so I think it's worth adding it now. This will also make my eventual PR for #926 less bulky.  

## Todos
- [ ] Tests
- [ ] Documentation


packages/react-native/examples/react-native-cli/android/app/src/main/java/com/awesomeproject/ThrowErrModule.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    ThrowErrModule(ReactApplicationContext context) {
+        super(context);
+    }
+
+    @ReactMethod
+    public void throwErr() {
+        final Handler handler = new Handler(Looper.getMainLooper());
+        handler.postDelayed(new Runnable() {
```
subzero10, line 27: @BethanyBerkowitz I guess this is the "delayed" error you were talking about. Looking at the code, this error won't crash the app, but the direct error will probably crash it.

subzero10 (approved):

Issue #1053, opened 2023-03-29
[rollup-plugin] - Sourcemaps failing to upload when directory containing them has been deleted prior to upload happening


joekrump:
## What are the steps to reproduce this issue?
1. Set up an app which runs `vite build` in a [Heroku Nodejs buildpack](https://github.com/heroku/heroku-buildpack-nodejs)
2. Have the build command run in the buildpack

## What happens?
- Most sourcemap files that are output are uploaded
- Sourcemap files that are found in the ephemeral `tmp/build_...` directory are not uploaded because they no longer exist, even though their entries show up in the bundle object which this plugin reads entries from. The build fails because this results in an Error being thrown.

## What were you expecting to happen?
- No issues. Sourcemaps are uploaded successfully.

## Any logs, error output, etc?
<img width="1027" alt="image" src="https://user-images.githubusercontent.com/3317231/228605068-bcc104e5-7e46-46dd-b560-eeae571fa9bd.png">

## What versions are you using?
**Operating System:** 
Building on the Heroku-20 stack
-----> Using buildpacks:
       1. heroku/nodejs
       2. heroku/ruby
**Package Name:** `@honeybadger-io/rollup-plugin`
**Package Version:** 5.1.5
**Browser Version:** N/A

joekrump:
It would be great if I could set something like an `ignorePath` regex or glob which I could set to `"tmp/build_*"` in the plugin config options. That would then be able to be used in `isSourcemap`: https://github.com/honeybadger-io/honeybadger-js/blob/a10f68655e4949e32f0fbe0ed8cdb763232d6497/packages/rollup-plugin/src/rollupUtils.ts#LL21C25-L21C25

BethanyBerkowitz:
Hi @joekrump, thanks for reaching out! Just to clarify, you believe the files in `tmp/` should not be uploaded, correct? I want to understand why they're even being generated since I don't see that happening when building locally with Vite. When I build, it looks like I just get assets in `/dist/assets`. Do you know if this is a Heroku thing?

Your suggestion of setting an `ignorePath` seems very reasonable, I'm just wondering if dumping files into `/tmp` and then deleting them is expected behavior.. if so, perhaps we should exclude those files by default rather than adding additional config. 

Are you able to tell what folder those assets that were successfully uploaded were located in? 

Also, would you mind sharing your `vite.config.js` just in case there's anything relevant?

subzero10:
[...]

PR #1059, opened 2023-04-25
feat: add ignorePaths option

subzero10:
Fixes: #1053

## Status
**READY**

## Description
Add an option `ignorePaths` to allow ignoring certain paths from being uploaded as source maps.

## TODO
- [x] Add `ignorePaths` option
- [x] Ignore source maps that have `sourcesContent === []`
- [x] Tests for `ignorePaths`
- [x] Tests for `sourcesContent`
- [x] Documentation

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -36,8 +36,24 @@ jobs:
     - name: Run unit tests
       run: npm test
 
+  changes:
+    runs-on: ubuntu-latest
+    # Set job outputs to values from filter step
+    outputs:
+      core: ${{ steps.filter.outputs.core }}
+    steps:
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            core:
+              - 'packages/core/**'
+              - 'packages/js/**'
+
   integration:
     # integration tests from "js" package fail if they are executed at the same time
+    needs: changes
+    if: needs.changes.outputs.core == 'true'
     concurrency: js-concurrency
     runs-on: ubuntu-latest
     steps:
--- a/packages/rollup-plugin/README.md
+++ b/packages/rollup-plugin/README.md
@@ -42,6 +42,13 @@ These plugin parameters correspond to the Honeybadger [Source Map Upload API](ht
   <dd>This package implements fetch retry functionality via the <a href="https://github.com/vercel/fetch-retry">fetch-retry</a> package. Retrying helps fix issues like `ECONNRESET` and `SOCKETTIMEOUT` errors.
   </dd>
 
+  <dt><code>ignorePaths</code> (optional &mdash; default: [])</dt>
+  <dd>An array of paths to ignore when uploading sourcemaps. Regex is also supported.
+  </dd>
+
+  <dt><code>deployEndpoint</code> (optional &mdash; default: "https://api.honeybadger.io/v1/deploys")</dt>
+  <dd>Where to send deployment notifications.</dd>
+
   <dt><code>deploy</code> (optional &mdash; default: false)</dt>
   <dd>
   Configuration for <a href="https://docs.honeybadger.io/api/reporting-deployments/">deployment notifications</a>. To disable deployment notifications, ignore this option. To enable deployment notifications, set this to <code>true</code>, or to an object containing any of the fields below. Your deploy's <code>revision</code> will be set to the same value as for your sourcemaps (see above). 
--- a/packages/rollup-plugin/README.md
+++ b/packages/rollup-plugin/README.md
@@ -43,7 +43,7 @@ These plugin parameters correspond to the Honeybadger [Source Map Upload API](ht
   </dd>
 
   <dt><code>ignorePaths</code> (optional &mdash; default: [])</dt>
-  <dd>An array of paths to ignore when uploading sourcemaps. Regex is also supported.
+  <dd>An array of paths (glob patterns) to ignore when uploading sourcemaps. Uses <a href="https://github.com/micromatch/picomatch">picomatch</a> to match against paths. 
   </dd>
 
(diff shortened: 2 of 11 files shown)


packages/rollup-plugin/src/rollupUtils.ts
```diff
@@ -6,30 +6,53 @@ import type { SourcemapInfo } from './types'
...
 function isSourcemap(file: OutputAsset | OutputChunk): file is OutputAsset {
-  return file.type === 'asset' && file.fileName.endsWith('.js.map')
+  if (file.type !== 'asset' || !file.fileName.endsWith('.js.map') || !file.source) {
+    return false
+  }
+
+  const source = typeof file.source === 'string' ? file.source : file.source.toString()
+  const json = JSON.parse(source)
+
+  return json.sourcesContent && json.sourcesContent.length > 0
```
BethanyBerkowitz, line 29: Since this function should return a boolean, you may want to coerce json.sourcesContent to be a boolean (eg `!!json.sourcesContent`).

packages/rollup-plugin/src/rollupUtils.ts
```diff
@@ -6,30 +6,53 @@ import type { SourcemapInfo } from './types'
...
+}
+
+function isNotIgnored(sourceMapInfo: SourcemapInfo, ignorePaths: Array<string | RegExp>) {
+  for (const ignorePath of ignorePaths) {
+    let regex = ignorePath
+    if (!(regex instanceof RegExp)) {
+      regex = new RegExp(regex)
```
BethanyBerkowitz, line 36: I'm wondering if treating strings as RegExp here could be confusing to the user. I wouldn't necessarily expect that passing `'foo.js'` would ignore `subfolder1/foo.js` and `subfolder2/foo.js`. I think I'd expect to have to pass a wildcard like `*/foo.js`. Here if you did pass in a wildcard like `'*/foo.js'` or `subfolder/*/foo`, we'd accept it but I don't think it would work. 

For example 
```
> let r = new RegExp('subfolder/*/foo.js')
undefined
> r.test('subfolder/bar/foo.js')
false
> 
```
This could be cleared up with documentation. I'm just thinking that using a wildcard for path matching rather than a regex has become pretty common, so people might expect it?

packages/rollup-plugin/README.md
```diff
@@ -42,6 +42,13 @@ These plugin parameters correspond to the Honeybadger [Source Map Upload API](ht
   <dd>This package implements fetch retry functionality via the <a href="https://github.com/vercel/fetch-retry">fetch-retry</a> package. Retrying helps fix issues like `ECONNRESET` and `SOCKETTIMEOUT` errors.
   </dd>
 
+  <dt><code>ignorePaths</code> (optional &mdash; default: [])</dt>
+  <dd>An array of paths to ignore when uploading sourcemaps. Regex is also supported.
+  </dd>
+
+  <dt><code>deployEndpoint</code> (optional &mdash; default: "https://api.honeybadger.io/v1/deploys")</dt>
```
subzero10, line 49: I noticed that this was missing.

.github/workflows/node.js.yml
```diff
@@ -36,8 +36,24 @@ jobs:
...
+        with:
+          filters: |
+            core:
+              - 'packages/core/**'
+              - 'packages/js/**'
+
   integration:
     # integration tests from "js" package fail if they are executed at the same time
+    needs: changes
+    if: needs.changes.outputs.core == 'true'
```
subzero10, line 56: No need to run integration tests when `js` or `core` do not have changed files.

packages/rollup-plugin/src/rollupUtils.ts
subzero10: That's a great suggestion! I switched to [picomatch](https://github.com/micromatch/picomatch) for glob pattern matching instead!

packages/rollup-plugin/test/rollupUtils.test.ts
```diff
@@ -7,33 +7,77 @@ import * as td from 'testdouble'
...
+        sourcemapFilename: 'subfolder/bar.js.map',
+        sourcemapFilePath: path.resolve('dist/subfolder/bar.js.map'),
+        jsFilename: 'subfolder/bar.js',
         jsFilePath: path.resolve('dist/subfolder/bar.js')
       },
     ])
   })
-});
+
+  const itEach = ['foo.js', '**/foo.js', 'foo.*', '**/foo.*', 'foo.js*', '**/foo.js*']
```
BethanyBerkowitz, line 37: Your readme change and TS types say that `ignorePaths` accepts regex, but I'm not sure if it actually does? If it does accept regex, you should add one to these tests.

BethanyBerkowitz (dismissed):
Thanks for tackling this! I added minor comments, I don't think I need re-review unless you prefer me to.

packages/rollup-plugin/src/rollupUtils.ts
```diff
@@ -1,35 +1,55 @@
...
+
+  const source = typeof file.source === 'string' ? file.source : file.source.toString()
+  const json = JSON.parse(source)
+
+  return !!json.sourcesContent && json.sourcesContent.length > 0
+}
+
+function isNotIgnored(sourceMapInfo: SourcemapInfo, ignorePaths: Array<string | RegExp>) {
+  for (const ignorePath of ignorePaths) {
+    const isMatch = picomatch.isMatch(sourceMapInfo.jsFilePath, ignorePath, { basename: true })
```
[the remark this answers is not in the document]
subzero10: Ah yes, good point!

packages/rollup-plugin/test/rollupUtils.test.ts
subzero10: The regex part of the readme is not there anymore, but I forgot to remove the TS types! On it!

Issue #1066, opened 2023-05-12
[react-native] Can't get source maps to work


subzero10:
A user reached out with the following issue:

> We have recently switched to React Native from Ionic and have needed to reconfig our process for uploading to HoneyBadger. I have followed these steps - https://docs.honeybadger.io/lib/react-native/ and have successfully implemented it. I'm able to generate errors, and log things in console and to the platform. I generated sourcemaps and have it as part of the build process. I can see in HB that my errors have the correct associated revision, and the sourcemaps have that same revision, but still being shown that I have no sourcemaps available for that issue.
> 
> Do I need file extensions on the sourcemap-ios / sourcemap-android? Am I uploading them incorrectly? Here is my curl request that I'm making after the maps have been generated.
> 
> curl https://api.honeybadger.io/v1/source_maps \
>   -F api_key=$PROJECT_API_KEY \
>   -F revision=$HONEYBADGER_REVISION \
>   -F minified_url=URL_TO_A_ZIP_ARCHIVE \
>   -F source_map=@$CURRENT_DIRECTORY/sourcemap-ios \
>   -F minified_file=@$CURRENT_DIRECTORY/sourcemap-ios
> Same for the android one.

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_mr5i9)

s-bridges:
Thank you for opening this for me @subzero10 !

BethanyBerkowitz:
@s-bridges, in investigating your issue, we've found some problems with how we're handling sourcemaps in the react-native package. For the moment, please hang on tight while we look into it. You should still be able to get errors reported, just without the sourcemaps for the moment. Thanks again for bringing this to our attention!

Issue #1029, opened 2023-02-27
[react-native] document sourcemap upload procedure 


BethanyBerkowitz:
We have an example React Native project in the react-native package directory, but there is no documentation for uploading source maps—and the general source maps documentation for React Native is pretty light. People tend to get confused by what the value of minified_url should be when uploading source maps. We need to improve those docs, or maybe even make a guide. 

https://github.com/honeybadger-io/honeybadger-js/tree/590895d7369a87e3f838e9e75341ab39e2bd8f69/packages/react-native#source-maps

This particular customer is trying to configure source maps for an Expo app. We have an issue related to Expo/React Native here:
https://github.com/honeybadger-io/honeybadger-js/issues/930

Issue #902, opened 2022-10-01
Set up Jest testing for React Native


subzero10:
Our React Native package does not have any unit tests. We should create some to maintain the quality of the package with updates.

PR #1069, opened 2023-05-17
feat(react-native)!: include the features of the core js project

### LINKED ISSUE #926: Depend on honeybadger-js or core

### LINKED ISSUE #928: Add Breadcrumbs support
Background:

> Breadcrumbs seems to be a missing feature in the react-native implementation of the SDK. This would be a great addition. Personally I use breadcrumbs when it comes to mobile apps for navigational events to help track down how a user got to a specific screen which could be helpful along with the stack trace of the error.

See also honeybadger-io/honeybadger-js#926 (which could give us the Breadcrumbs API for free).

## Specification
The client should have a **public** `addBreadcrumb()` method that stores an array of breadcrumbs. There should also be a public `clear()` method (see also: honeybadger-io/honeybadger-react-native#4) to reset the breadcrumbs to the default (`[]`).

- Data should be stored [local to the current thread](https://www.notion.so/Client-Library-Spec-aa891332f7874196aa0695b6d38dca66)
- Data should be sanitized, but not [filtered](https://www.notion.so/Client-Library-Spec-aa891332f7874196aa0695b6d38dca66)

**Supported config options:**

- `breadcrumbsEnabled`
    - optional
    - default: `true`
- `maxBreadcrumbs`
    - optional
    - default: `40`

**Example:**

```ruby
# Supported configuration options w/ defaults:
Honeybadger.configure({
  # optional
  breadcrumbsEnabled: true,

  # optional
  # The max number of breadcrumbs to store at one time.
  # The oldest breadcrumbs should be dropped as newer breadcrumbs are added.
	maxBreadcrumbs: 40
})

# Adding a breadcrumb:
Honeybadger.addBreadcrumb(
	"Email Sent",
	metadata: {
		user: user.id,
		message: message
	}
)

# Clearing context and breadcrumbs:
Honeybadger.clear()
```

### Resources
- [Client Library Spec](https://www.notion.so/honeybadger/Client-Library-Spec-aa891332f7874196aa0695b6d38dca66#d9a1892930694580a9f41eebb99c22bd)
- [Honeybadger Documentation](https://docs.honeybadger.io/lib/javascript/guides/breadcrumbs.html)
- [Exceptions API Documentation](https://docs.honeybadger.io/api/exceptions.html)
- [Notices JSON Schema](https://github.com/honeybadger-io/honeybadger-schemas/blob/master/exceptions/v1/notices.json)

### LINKED ISSUE #932: Document configuration method and options
Now that we have multiple configuration options available (#4), we should have a section on config. The new [Crystal docs](https://github.com/honeybadger-io/honeybadger-crystal#configuration) have a good example of what this section should look like.

### LINKED ISSUE #1054: Duplicate JS errors reported in react-native on iOS, missing on Android
1. In `examples/react-native-cli`, run `npm install && cd ios && pod install && cd ..`
2. `npm start`
3. In another terminal, run a Release build of the example app `npx react-native run-ios --configuration Release`  (I did this on the simulator)
4. In the simulator (or on your device), enter your API key and click "Configure HB". 
5. Click "Throw a JS Error!"

## What happens?
In the HB UI, I see a duplicate error reported when throwing a JS error on iOS. 
Screen Shot 2023-04-04 at 10 28 44 AM

I was having this issue in my branch, but it turns out it happens on the released version of the package as well. 

One has the expected message: "This is a test error from the react-native-cli example project!"
- it does not have a stack trace 
- https://app.honeybadger.io/projects/108285/faults/94966597/01GX6B0YZBFDWRAF0JNZY0MME0?page=0 

Other one has prepended and appended data: "Unhandled JS Exception: Error: This is a test error from the react-native-cli example project!, js engine: hermes"
- does have a stack trace 
- https://app.honeybadger.io/projects/108285/faults/94966598/01GX6B0YZ89EST3BEK6G9JM6HN?page=0

Just one error, sent using the JS handler. 

…

## Any other comments?
This is specific to production builds -- when running in dev mode (`npm run ios`), I only get one error with the expected `This is a test error from the react-native-cli example project!`. 

**Operating System:** macOS Monterey, running xcode simulator for ios app
**Package Name:** react-native
**Package Version:** 5.1.6
**Browser Version:** n/a

BethanyBerkowitz:
## Status
**READY**

## Description
This PR represents a major re-work of the react-native package to depend on the `core` package. This will make the behavior of the react-native package more consistent with the rest of our javascript packages. 
Closes #926.
Closes #1054.
Closes #932. 
Closes #928.

This is definitely a breaking change, since the config options now must be passed as an object (like the rest of the packages).

High-level changes include: 

- Rely on core package instead of using custom `notify()`, `setContext()`, etc. 
- Add `transport` specific to this package but based on [HB's js package](https://github.com/honeybadger-io/honeybadger-js/blob/6bfda1cd24abf5b9d86a5d95540cbf9895f9b17f/packages/js/src/server/transport.ts)
- Use typescript
- Build with rollup (using just `tsc` results in a messy nested structure)
- No longer pass to built-in error handler after notifying in release build. This removes the duplicate error reporting documented in https://github.com/honeybadger-io/honeybadger-js/issues/1054

- [x] Manual testing using an Expo setup (I mostly tested using `examples/react-native-cli`)
- [x] Test throwing a native error in the Expo app
- [x] Test that other aspects of core package work correctly (notify(), setContext(), beforeNotify(), afterNotify(), addBreadcrumb())
- [x] Documentation updates

The following are missing features. However they are also missing in the existing package, so they don't need to hold up this PR. 
- Sourcemaps/symbolication: https://github.com/honeybadger-io/honeybadger-js/issues/1066, https://github.com/honeybadger-io/honeybadger-js/issues/1029, https://github.com/honeybadger-io/honeybadger-js/issues/1082
- Unit testing: https://github.com/honeybadger-io/honeybadger-js/issues/902 
- Capturing Android errors that cause an immediate crash in production: https://github.com/honeybadger-io/honeybadger-js/issues/1050


packages/react-native/package.json
subzero10, line 1: I think we need to set a "files" property here, otherwise we will be bundling the example projects as well, right? (which increases the package size unnecessarily)

packages/react-native/src/transport.ts
```diff
@@ -0,0 +1,51 @@
...
+      version: pkg.version,
+    }
+
+    const params = {
+      method: options.method,
+      headers: {
+        ...options.headers, 
+        'User-Agent': this.buildUserAgent()
+      },
+      // TODO: is this sanitize helpful?
```
subzero10, line 24: I can't tell for sure here, but generally yes. It's the function we have at the client to sanitize the payload (for example, recursive structures).

subzero10 (commented):
I had a first look and this looks very promising!

packages/react-native/package.json
BethanyBerkowitz: I think the `.npmignore` should serve this purpose!

subzero10: Ah yes! I missed that!

BethanyBerkowitz:
Using the Expo example app, I tested throwing an uncaught JS error and calling hb.notify() directly. Both seem to work fine. Got stuck for a while trying to figure out how to run the Expo example project in "production" mode, but I eventually figured that out and added scripts for it to `package.json` and a note in the README. 

I did not yet test throwing a native error in the Expo app, though I probably should do that.

BethanyBerkowitz:
Tested `Honeybadger.addBreadcrumb()` and `Honeybadger.clear()` on iOS emulator. Both seem to be working! 
Screen Shot 2023-05-25 at 4 09 24 PM

packages/react-native/examples/expo-app/package.json
```diff
@@ -2,17 +2,18 @@
...
     "ios": "npx expo run:ios",
-    "web": "expo start --web"
+    "android:release": "npx expo run:android --variant release",
+    "ios:release": "npx expo run:ios --configuration Release",
+    "start": "expo start --dev-client"
   },
   "dependencies": {
-    "@honeybadger-io/react-native": "^5.1.1",
-    "expo": "~47.0.12",
+    "@honeybadger-io/react-native": "file:../../honeybadger-io-react-native-5.1.6_517150.tgz",
```
BethanyBerkowitz, line 12: I'm fine with just changing this to `"@honeybadger-io/react-native": "latest"`, however I do think my testing procedure is not super obvious and should be documented. Before I add documentation, can you think of a better way to do it @subzero10?  

Here's what I've been doing: 

1. In the `react-native` folder, run `npm run build && npm pack`
2. Rename the tarball to append some unique string and add the tarball file as the dependency in the test project(s) I'm working on (for a while I struggled because changes to the package were not coming through and it turned out that there was some caching issue. Renaming the file fixes that, however it's quite annoying)
3. `npm install` 

I tried just installing the package without making a tarball first like `"@honeybadger-io/react-native": "file:../.."`, but that never worked for me.

subzero10:
@BethanyBerkowitz I modified the PR title with the correct syntax for a breaking change.

packages/react-native/src/index.ts
```diff
@@ -0,0 +1,140 @@
...
+  private onNativeAndroidException(data:NativeExceptionData) {
+    const notice = {
+      name: `React Native Android ${data.type}`,
+      message: data.message || '',
+      details: {
+        errorDomain: data.errorDomain || '',
+        initialHandler: data.initialHandler || '',
+        userInfo: data.userInfo || {},
+        architecture: data.architecture || '',
+      },
```
subzero10, line 126: What's the difference between `details` and `context` here?
cc @joshuap 🤔

packages/react-native/src/index.ts
```diff
@@ -0,0 +1,140 @@
...
+import { Client, Types } from '@honeybadger-io/core'
+import { Transport } from './transport'
+import { backtraceAndDetailsFromIosException, errorMessageFromIosException } from './iosUtils'
+import { backtraceFromAndroidException } from './androidUtils'
+import { NativeExceptionData } from './types'
+
+class Honeybadger extends Client {
+  protected __jsHandlerInitialized:boolean
+  protected __nativeHandlerInitialized:boolean
+  protected __originalJsHandler:(error: Error, isFatal: boolean) => void
```
subzero10, line 11: Why not `private`?

packages/react-native/README.md
```diff
@@ -34,77 +36,21 @@ You can log into your [Honeybadger](https://honeybadger.io) account to obtain yo
...
-```
-
-The **reportErrors** parameter determines if errors are to be sent to Honeybadger. This is set to **true** by default. In certain environments, say, during development, it could be useful to set **reportErrors** to **false** to prevent errors from being posted to your Honeybadger account.
-
-
-## Usage Examples
-
-iOS, Android, and JavaScript errors will be automatically handled by the Honeybadger React Native library, by default. But you can also use the following API to customize error handling in your application.
-
-### Honeybadger.notify(error, additionalData)
```
subzero10, line 62: I see why you'd remove this, but I think we should include usage examples of the main features of the package (at least this is what we do with the other packages).
If you look at [React](https://docs.honeybadger.io/lib/javascript/integration/react/) or [Vue](https://docs.honeybadger.io/lib/javascript/integration/vue3/) integration guides, there are a few sections that are always included (Installation, Reporting Errors, Identifying Usage) and then a reference to the js/core package for advanced usage.

packages/react-native/examples/expo-app/package.json
subzero10: Actually, I didn't know about this trick before seeing your approach with the tarball file!

I've had the same issue when installing with `"file:../.."` so what I would usually do is install the from npm (`"@honeybadger-io/react-native": "latest"`) and then modify code in `node_modules` while developing. So yeah, I don't have a good answer 😓.

subzero10:
@BethanyBerkowitz I reviewed the part of the code I understand :)

There's a final point I want to bring up:
If we want to leverage Typescript Project References correctly, we would need to build the typescript files outside rollup.
See this [PR](https://github.com/honeybadger-io/honeybadger-js/pull/1078) for more information.
However, this PR is already quite big, so I wouldn't consider this a blocker, we can fix this with a separately.

packages/react-native/README.md
andreybutov: @BethanyBerkowitz +1 on "... I think we should include usage examples of the main features of the package"

packages/react-native/examples/expo-app/App.js
```diff
@@ -1,38 +1,62 @@
 import React, { useState } from 'react';
 import { StyleSheet, View, Button, TextInput } from 'react-native';
 import Honeybadger from '@honeybadger-io/react-native';
+import { throwNativeErr } from './modules/thow-err-module'
```
andreybutov, line 4: @BethanyBerkowitz `thow-err-module` -- thow?

packages/react-native/examples/react-native-cli/package.json
```diff
@@ -3,14 +3,18 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
+    "install:clean": "rm -rf node_modules && npm install",
     "android": "react-native run-android",
+    "android:release": "rm -rf android/app/build && react-native run-android --mode release",
     "ios": "react-native run-ios",
+    "ios:install": "cd ios && pod install && cd ..",
```
andreybutov, line 10: @BethanyBerkowitz Nice. I like the addition of "ios:install" a lot.

packages/react-native/package.json
```diff
@@ -1,36 +1,46 @@
...
+  "license": "MIT",
+  "types": "dist/index.d.ts",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "rollup -c",
+    "build:clean": "rm -rf dist && npm run build",
+    "test": "echo \"Error: no test specified\""
+  },
+  "bin": {
+    "honeybadger-generate-sourcemaps": "./honeybadger-generate-sourcemaps.sh"
```
andreybutov, line 15: I'm still not sure how npm works with all of this being moved because of the monorepo stuff. Also, how is the "generate-sourcemaps" script in bin supposed to work now? There's lots of stuff that I'm not sure about anymore :) But I leave the npm stuff to the experts.

packages/react-native/src/androidUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  function isStringWithValue(val:unknown):boolean {
+    return typeof val === 'string' && val.trim().length > 0;
+  }
+
+  return data.stackTrace.map((frame) => {
+    const method = (isStringWithValue(frame.class) && isStringWithValue(frame.method)) 
+      ? `${frame.class}.${frame.method}` 
+      : frame.method;
+    return {
```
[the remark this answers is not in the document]
andreybutov: @subzero10 Native Android land only sends back "file", "class", "method", and "line" per stack frame.

andreybutov (approved):

packages/react-native/examples/expo-app/package.json
BethanyBerkowitz: Ok, I think I will go with "latest" and add a note in the README.

packages/react-native/src/iosUtils.ts
```diff
@@ -0,0 +1,125 @@
...
+        regex.lastIndex++;
+      }
+
+      const moduleName = match && match.length > 1 ? match[1] : '';
+      const stackAddress = match && match.length > 2 ? match[2] : '';
+      const loadAddress = match && match.length > 3 ? match[3] : '';
+      const file = match && match.length > 6 ? match[6] : '';
+      const line = match && match.length > 7 ? match[7] : '';
+
+      // TODO: Why doesn't this match the BacktraceFrame type?
```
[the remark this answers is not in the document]
joshuap: We don't support a `stack_adddress` property as far as I know. Can someone write up a proposal if that's something we should consider? Keep in mind that we try to keep the backtrace representation as generic as possible so that other languages can use the property if they have a similar feature.

packages/react-native/examples/expo-app/ios/expoapp/Supporting/Expo.plist
```diff
@@ -11,6 +11,6 @@
     <key>EXUpdatesSDKVersion</key>
     <string>47.0.0</string>
     <key>EXUpdatesURL</key>
-    <string>https://exp.host/@anonymous/expo-app</string>
+    <string>https://exp.host/@bethany_b_honeybadger/expo-app</string>
```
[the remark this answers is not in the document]
BethanyBerkowitz: This got auto-generated by Expo, I think because I was logged in to Expo. However, I tried logging out and the example project still runs, so I don't think this will cause issues. I'm inclined to leave it as is, but if someone else could make sure the example projects run for them, that would be helpful!

packages/react-native/src/iosUtils.ts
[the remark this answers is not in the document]
BethanyBerkowitz: This I also copied directly from the original code. I was confused about `stack_address` and lack of `column` but happy to look further into it!

packages/react-native/src/index.ts
BethanyBerkowitz: Copied this from the original package as well -- I was trying to make as few changes as possible. I will look into it, though, I'm not sure we actually want this here.

packages/react-native/src/index.ts
BethanyBerkowitz: No reason, can go with `private`!

packages/react-native/README.md
BethanyBerkowitz: Cool, fine by me. It just seemed like copy/pasting the same instructions into all the libraries might be a maintenance headache. But it's definitely easier to read.

BethanyBerkowitz: @subzero10 In the examples you linked, those sections are actually held in the docs repo, not in the READMEs. Happy to put them in the README here, but for consistency I'd put them in `docs` instead.

packages/react-native/examples/expo-app/App.js
BethanyBerkowitz: Nice catch, thanks!

packages/react-native/examples/expo-app/ios/expoapp/Supporting/Expo.plist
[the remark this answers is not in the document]
subzero10: I will try the example project at some point, but we can consider this as resolved for now.

packages/react-native/src/iosUtils.ts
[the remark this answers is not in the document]
subzero10: I think it makes more sense to remove at this point, since we don't even know what it was supposed to be used for. Unless @andreybutov has some idea?

packages/react-native/src/index.ts
subzero10: Thanks @joshuap!

@BethanyBerkowitz I mainly wanted to know how the `details` property is used. We don't have to change anything if this works 👌.

packages/react-native/README.md
subzero10: Yes, I posted those to show that even if they inherit most of their features from the js package, they still include sections on the main features.

For the React Native documentation page, the current state is that the documentation is built based on the README. You can change that though.

So, to clarify:
- we can update the React Native README file and make sure the docs page is reflected with the content of the latest README
- we can cleanup the React Native README file (as you did) and make sure that the docs page is on par with the rest of our integration guides

I'm fine with either approach!

packages/react-native/package.json
subzero10: > Also, how is the "generate-sourcemaps" script in bin supposed to work now?

It should work in the same way, no change in behavior.

BethanyBerkowitz:
> There's a final point I want to bring up: If we want to leverage Typescript Project References correctly, we would need to build the typescript files outside rollup. See this [PR](https://github.com/honeybadger-io/honeybadger-js/pull/1078) for more information. However, this PR is already quite big, so I wouldn't consider this a blocker, we can fix it separately.

Thanks! Added an issue to the backlog. 
https://github.com/honeybadger-io/honeybadger-js/issues/1081

packages/react-native/src/iosUtils.ts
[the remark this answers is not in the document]
BethanyBerkowitz: Ok, so I spent some time looking deeper into this. I had tried to avoid needing to understand how the iOS stack traces work by just copying @andreybutov's original code, but alas I've now learned enough to know how much I don't know  😆 

The `localizedDescription` here looks something like: 
```
Exception 'Testing native iOS exception' was thrown while invoking throwErr on target ThrowErrModule with params (
)
callstack: (
	0   CoreFoundation                      0x00007ff8004288ab __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007ff80004dba3 objc_exception_throw + 48
	2   AwesomeProject                      0x00000001088b2ab0 -[AppDelegate application:didFinishLaunchingWithOptions:] + 0
	3   CoreFoundation                      0x00007ff80042f2fc __invoking___ + 140
	4   CoreFoundation                      0x00007ff80042c529 -[NSInvocation invoke] + 305
	5   CoreFoundation                      0x00007ff80042c7bc -[NSInvocation invokeWithTarget:] + 70
	6   AwesomeProject                      0x0000000108dd9aa6 -[RCTModuleMethod invokeWithBridge:module:arguments:] + 2534
	7   AwesomeProject                      0x0000000108dde69b _ZN8facebook5reactL11invokeInnerEP9RCTBridgeP13RCTModuleDatajRKN5folly7dynamicEiN12_GLOBAL__N_117SchedulingContextE + 2043
	8   AwesomeProject                      0x0000000108dddccc _ZZN8facebook5react15RCTNativeModule6invokeEjON5folly7dynamicEiENK3$_0clEv + 156
	9   AwesomeProject                      0x0000000108dddc29 ___ZN8facebook5react15RCTNativeModule6invokeEjON5folly7dynamicEi_block_invoke + 25
	10  libdispatch.dylib                   0x00007ff80013b7fb _dispatch_call_block_and_release + 12
	11  libdispatch.dylib                   0x00007ff80013ca3a _dispatch_client_callout + 8
	12  libdispatch.dylib                   0x00007ff800144346 _dispatch_lane_serial_drain + 949
	13  libdispatch.dylib                   0x00007ff800144fe3 _dispatch_lane_invoke + 400
	14  libdispatch.dylib                   0x00007ff800151b46 _dispatch_workloop_worker_thread + 870
	15  libsystem_pthread.dylib             0x00007ff836170fd0 _pthread_wqthread + 326
	16  libsystem_pthread.dylib             0x00007ff83616ff57 start_wqthread + 15
)
```
As far as I can tell, this call stack doesn't actually contain lines or columns, or even file names. Instead it has some kind of memory address. I found this helpful: https://www.bugsnag.com/blog/symbolicating-ios-crashes. I'm not sure if this has changed since Andrey worked on it, but the `file` and `line` variables in this code are undefined. 

Apparently in order to get something useful out of an error like this, we need to use a `dSYM` file. I'm not sure how to generate one automatically or how to use one, but I'm looking into it. 

I so far haven't found a good breakdown of exactly what each item in the call stack represents -- if anyone finds a good resource for this please lmk.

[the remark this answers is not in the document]
andreybutov: @BethanyBerkowitz I have some notes on this. Let me get them in some sort of presentable order and I'll post them in Slack.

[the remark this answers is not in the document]
subzero10: > I'm not sure if this has changed since Andrey worked on it, but the file and line variables in this code are undefined.

If this is not something we introduced with this PR, my suggestion remains the same: let's tackle this with a separate issue, independently.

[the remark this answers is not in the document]
BethanyBerkowitz: https://github.com/honeybadger-io/honeybadger-js/issues/1082

packages/react-native/README.md
BethanyBerkowitz: Here's the documentation PR: https://github.com/honeybadger-io/docs/pull/303

packages/react-native/README.md
```diff
@@ -138,6 +95,30 @@ npm run android
...
+```shell
+npm run ios:release
+```
+or
+```shell
+npm run android:release
+```
+
+### Development workflow
+When developing the react-native package and testing against the example projects, you will need to change the dependency in the example project's `package.json`. Rather than `"@honeybadger-io/react-native": "latest"`, you'll want to generate a tarball of the `react-native` package and install it. For example:
```
subzero10, line 108: Thanks for putting together this part here, I know it's not pretty but at least we have a guide.

subzero10 (approved):

packages/nextjs/package-lock.json
subzero10, line 1: Can you please undo the changes in this file?

I'm afraid that if this file is associated with a BREAKING CHANGE commit, it will bump the major version of the nextjs package as well!

subzero10 (changes requested):
I'm requesting changes just to revert the nextjs package-lock.json file.

packages/nextjs/package-lock.json
BethanyBerkowitz: Good catch, thank you!

Issue #1070, opened 2023-05-19
Add support for App Router


subzero10:
See https://github.com/honeybadger-io/nextjs-with-honeybadger/issues/14.

Issue #1055, opened 2023-04-10
Uncaught errors in API routes or middleware functions are not reported


subzero10:
Uncaught errors in API routes or middleware functions are not reported, because they reach our custom `_error.js` handler with [status code 404](https://nextjs.org/docs/advanced-features/custom-error-page#caveats) and have no context of the error.

Possible solution:
- Wrap all middleware functions and api handlers with a try/catch block that reports to Honeybadger. It would be great to do transparently with webpack rather than asking users to wrap all their functions with a try/catch (or a Honeybadger utility method).

PR #1071, opened 2023-05-19
feat(nextjs): app router support

subzero10:
Closes: #1070

## Status
**READY**

## Description
Adds support for the new [App Router in Next.js](https://nextjs.org/docs/app/building-your-application/routing#the-app-directory).

## Todos
- [x] Implement support for app router
- [x] Make sure that source maps are correctly uploaded for the app directory
- [x] BONUS: support projects built with `.tsx` (Typescript) or `.js` files

### DIFF
--- a/packages/nextjs/src/copy-config-files.ts
+++ b/packages/nextjs/src/copy-config-files.ts
@@ -36,10 +36,15 @@ function getTargetPath(isAppRouter = false, isGlobalErrorComponent = false) {
   return path.join(srcFolder, fileName + '.' + extension)
 }
 
-function getTemplate(isAppRouter = false) {
-  const templateName = isAppRouter ? '_error_app_router.js' : '_error.js'
+function getTemplate(isAppRouter = false, isGlobalErrorComponent = false) {
+  if (!isAppRouter && isGlobalErrorComponent) {
+    throw new Error('invalid arguments: isGlobalErrorComponent can only be true when isAppRouter is true')
+  }
 
-  return path.resolve(__dirname, '../templates', templateName)
+  const extension = isGlobalErrorComponent ? 'tsx' : 'js'
+  const templateName = isAppRouter ? '_error_app_router' : '_error'
+
+  return path.resolve(__dirname, '../templates', templateName + '.' + extension)
 }
 
 async function copyErrorJs(isAppRouter = false) {
@@ -50,7 +55,7 @@ async function copyErrorJs(isAppRouter = false) {
 }
 
 function copyGlobalErrorJs() {
-  const sourcePath = getTemplate(true)
+  const sourcePath = getTemplate(true, true)
   const targetPath = getTargetPath(true, true)
 
   return copyFileWithBackup(sourcePath, targetPath)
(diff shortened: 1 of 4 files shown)


subzero10:
I also have an example project that I will push into the nextjs package with #1072.

packages/nextjs/src/copy-config-files.ts
```diff
@@ -5,18 +5,57 @@ const fs = require('fs')
 
 const debug = process.env.HONEYBADGER_DEBUG === 'true'
 
+function usesTypescript() {
+  return fs.existsSync('tsconfig.json')
+}
+
+function usesAppRouter() {
```
BethanyBerkowitz, line 12: I saw in the Next documentation that they're supporting `app` and `pages` concurrently.. is that going to cause any issues ?

subzero10: That's a nice catch! I will have to handle that case as well 😄

subzero10:
Error reported automatically while rendering component. Source maps applied successfully.

image

subzero10:
Error reported automatically on the server side, during a data fetching method. However, most of the context is missing because [Next.js does not propagate server errors](https://nextjs.org/docs/app/building-your-application/routing/error-handling#handling-server-errors) to the Honeybadger React Error Boundary (since it's on the client side).

We can look into improving this with #1055.

image

subzero10:
@BethanyBerkowitz Can I please get another round of review 🙏 ?

Thanks!

BethanyBerkowitz:
> Error reported automatically while rendering client-side component. Source maps applied successfully.
> 
> image

This shows that the sourcemap was applied, but the file names are still minified... shouldn't they be the original file names and not minified ones?

packages/nextjs/src/copy-config-files.ts
```diff
@@ -5,17 +5,76 @@ const fs = require('fs')
...
+  return fs.existsSync('app')
+}
+
+function getTargetPath(isAppRouter = false, isGlobalErrorComponent = false) {
+  if (!isAppRouter && isGlobalErrorComponent) {
+    throw new Error('invalid arguments: isGlobalErrorComponent can only be true when isAppRouter is true')
+  }
+
+  const extension = usesTypescript() ? 'tsx' : 'js'
```
BethanyBerkowitz, line 25: Why is it only either `.tsx` or `.js`? Are `.ts` and `.jsx` not allowed? (pardon my ignorance on this)

BethanyBerkowitz (commented):
Code looks good to me, I left minor questions / comments. Did you want me to test this in some way?

packages/nextjs/src/copy-config-files.ts
subzero10: It's a good question! I am following Next.js examples and these are the extensions used. So I used just these for now.

subzero10:
> This shows that the sourcemap was applied, but the file names are still minified... shouldn't they be the original file names and not minified ones?

Are you referring to `react-dom.production.min.js`? This is a dependency, i.e. not part of the application code.
`page.tsx` is application code and appears to be shown correctly.

subzero10:
> Did you want me to test this in some way?

I don't think it's necessary at this point, but if you want to play with Next.js, feel free to do it!

subzero10:
@BethanyBerkowitz Can you please approve, if you are happy with my replies 😄  ?

PR #1078, opened 2023-05-26
chore: use project references correctly

subzero10:
## Status
**READY**

## Description
I don't remember why, but the tsconfig.json files that have package dependencies within the monorepo were missing project references.
Project References are a handy Typescript feature that makes development easier by keeping the dependency package up-to-date with the latest changes while development.
For example, if I make a change in the core package, it should be available in the js package, just by building the js package (tsc will see that core is a dependency of the js package and would build it as well).

Note:
This is why we transpile typescript files before passing on to rollup; rollup does not work with project references.

### DIFF
--- a/packages/core/package-lock.json
+++ b/packages/core/package-lock.json
@@ -20,6 +20,9 @@
         "sinon": "^14.0.0",
         "ts-jest": "^27.1.5",
         "typescript": "^4.5.3"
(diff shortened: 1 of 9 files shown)


BethanyBerkowitz (dismissed):
Cool, I didn't know about this feature!

Issue #1075, opened 2023-05-25
Invalid types on HoneybadgerErrorBoundary and ErrorCompon


dvictory:
## What are the steps to reproduce this issue?
1. use Typescript
2. Create a React Component(DefaultErrorComponent), copy contents of DefaultErrorComponent (Honeybadger version) to component, basically make a copy
3. Pass the Component as the property ErrorComponent to HoneybadgerErrorBoundary

## What happens?
Type error CustomComponent is not of type ReactNode

## What were you expecting to happen?
No Type errors

## Any logs, error output, etc?
Does not compile.

## Any other comments?
I think the type defs for the HoneybadgerErrorBoundary are incorrect.  

https://github.com/honeybadger-io/honeybadger-js/blob/master/packages/react/src/HoneybadgerErrorBoundary.tsx#L22 conflicts with https://github.com/honeybadger-io/honeybadger-js/blob/master/packages/react/src/HoneybadgerErrorBoundary.tsx#L9

ReactNode is the first, and the 2nd is PropTypes.oneOfType([PropTypes.element, PropTypes.func]).

I am no React expert, but seems like something is off. I would think you could pass a copy of the default as an override and not have issues.

I solved with:
```jsx
HoneybadgerErrorBoundary
          honeybadger={Honeybadger}
          ErrorComponent={DefaultErrorComponent as unknown as ReactNode}
        >
```
ew, this seems wrong.

Thanks

subzero10:
Hey @dvictory, thank you for submitting an issue!

Indeed, this `ErrorComponent={DefaultErrorComponent as unknown as ReactNode}` does not look good!
I tried reproducing your issue in the [example project](https://github.com/honeybadger-io/honeybadger-js/tree/master/packages/react/example), but it seems that a custom error component works for me.
Here's what I did:
- I created a `CustomErrorComponent.tsx`, which is a copy of the `DefaultErrorComponent.tsx`, with the class and file names changed.
- I pass the CustomErrorComponent to the `HoneybadgerErrorBoundary`:
```
  import CustomErrorComponent from './CustomErrorComponent';
  ...
  ReactDOM.render(
    <HoneybadgerErrorBoundary honeybadger={Honeybadger} ErrorComponent={CustomErrorComponent}>
      <App />
    </HoneybadgerErrorBoundary>,
    document.getElementById('root')
  )
```
- I start the project with `npm run start` and all works as expected.

---
I took a look at the type definitions:
[...]

PR #1083, opened 2023-06-02
fix(react): update types to work with react 18

subzero10:
## Status
**READY**

## Description
Fixes: #1075

### DIFF
--- a/packages/react/src/index.test.tsx
+++ b/packages/react/src/index.test.tsx
@@ -72,10 +72,11 @@ describe('HoneybadgerReact', () => {
       sandbox.spy(honeybadger, 'notify')
       sandbox.spy(honeybadger, 'showUserFeedbackForm')
 
-      // need to cast to unknown because ErrorComponent does not expect a jest.Mock type
-      const MyError = jest.fn(() => 'custom error view') as unknown as string
-      TestRenderer.create(<HoneybadgerErrorBoundary honeybadger={honeybadger} ErrorComponent={MyError}><Broken /></HoneybadgerErrorBoundary>)
-      expect(MyError).toBeCalledWith({
+      const MyErrComponent = () => <>custom error</>
+      const MyErrComponentMock = jest.fn(MyErrComponent) as jest.MockedFunction<typeof MyErrComponent>
+
+      TestRenderer.create(<HoneybadgerErrorBoundary honeybadger={honeybadger} ErrorComponent={MyErrComponentMock}><Broken /></HoneybadgerErrorBoundary>)
+      expect(MyErrComponentMock).toBeCalledWith({
         error: expect.any(Error),
         info: { componentStack: expect.any(String) },
         errorOccurred: expect.any(Boolean)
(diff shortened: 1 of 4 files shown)


packages/react/src/index.test.tsx
```diff
@@ -72,7 +72,8 @@ describe('HoneybadgerReact', () => {
       sandbox.spy(honeybadger, 'notify')
       sandbox.spy(honeybadger, 'showUserFeedbackForm')
 
-      const MyError = jest.fn(() => 'custom error view')
+      // need to cast to unknown because ErrorComponent does not expect a jest.Mock type
```
subzero10, line 75: I'm not happy with this, but I spend too much time trying to get it right. I don't know why I can't get it work 🤷

BethanyBerkowitz: Ah yeah, the jest mock type is getting in the way! Can you try something like:

```
const ErrComponent = () => <>custom error view</>
const MockErrComponent = ErrComponent as jest.MockedFunction<typeof ErrComponent>
      
TestRenderer.create(<HoneybadgerErrorBoundary honeybadger={honeybadger} showUserFeedbackFormOnError={true} ErrorComponent={MockErrComponent}><Broken /></HoneybadgerErrorBoundary>)
```

This is helpful: https://www.benmvp.com/blog/using-jest-mock-functions-typescript/

BethanyBerkowitz (changes requested):
I think the type issue can be fixed (see comment). 

Also, do you want to update the example project to use React 18? I could see an argument either way... I think React 18 is substantially different than prior versions.

packages/react/src/index.test.tsx
subzero10: Thanks! I tried it and it didn't work.

And actually, in the snippet above, there is no call to jest.mock or jest.fn so the ErrComponent is never mocked.
In the example you showed, there component was a separate file, and was mocked with `jest.mock('./ComponentFile')` before being used.

I already tried this and faced other issues:
```
// added these at the top of the tests file
import DefaultErrorComponent from './DefaultErrorComponent';

jest.mock('./DefaultErrorComponent');

// then, I modified the test to simply use DefaultErrorComponent
TestRenderer.create(<HoneybadgerErrorBoundary honeybadger={honeybadger} ErrorComponent={DefaultErrorComponent}><Broken /></HoneybadgerErrorBoundary>)
```

With the above code, the test is now failing with this:
```
expect(jest.fn()).toBeCalledWith(...expected)

    - Expected
    + Received

      Object {
    -   "error": Any<Error>,
    -   "errorOccurred": Any<Boolean>,
    -   "info": Object {
    -     "componentStack": Any<String>,
    -   },
    +   "error": [Error: Busted, sorry],
    +   "errorOccurred": true,
    +   "info": null,
      },
      {},

    Number of calls: 1
```

The issue here is that `info` is null and after diving into it, it seems to be because React will call render more than once (which is fine), and will eventually call it with all the necessary props. I will take another stab at it with `jest.mock('./DefaultErrorComponent')`. Perhaps we should change the way we make our assertions.

BethanyBerkowitz: Huh weird, I tested this before recommending it to you... let me try again.

BethanyBerkowitz: Ah ok mine does seem to pass, but I made a typo when I posted it to you. My apologies! Can you give this a shot? 

```
it('should render a custom error message when a component errors', (done) => {
      sandbox.spy(honeybadger, 'notify')
      sandbox.spy(honeybadger, 'showUserFeedbackForm')

      const MyErrComponent = () => <>custom error</>
      const MyErrComponentMock = jest.fn(MyErrComponent) as jest.MockedFunction<typeof MyErrComponent>
      
      TestRenderer.create(<HoneybadgerErrorBoundary honeybadger={honeybadger} ErrorComponent={MyErrComponentMock}><Broken /></HoneybadgerErrorBoundary>)
      expect(MyErrComponentMock).toBeCalledWith({
        error: expect.any(Error),
        info: { componentStack: expect.any(String) },
        errorOccurred: expect.any(Boolean)
      }, {})
      // Still want to ensure notify is only called once. The MyErrComponent component will be created twice by React.
      afterNotify(done, function () {
        assert.calledOnce(honeybadger.notify as SinonSpy)
        assert.notCalled(honeybadger.showUserFeedbackForm as SinonSpy)
      })
    })
```

subzero10: It works, huuuuuge thank you! 🔥

subzero10:
> Also, do you want to update the example project to use React 18? I could see an argument either way... I think React 18 is substantially different than prior versions.

I think it makes sense for the example projects to be on the latest version.

BethanyBerkowitz (approved):
🥳

PR #1090, opened 2023-06-12
chore(react-native): react native unit testing setup

BethanyBerkowitz:
## Status
READY

## Description
Closes https://github.com/honeybadger-io/honeybadger-js/issues/902. 

It includes: 

1. Jest setup
2. `androidUtils` testing
3. `iosUtils` testing (more would be helpful, see note in the file)
4. `transport` testing

It does not include testing on `index.ts`, which can go in a follow-up PR. 

## Steps to Test or Reproduce
```bash
cd packages/react-native
npm install
npm test
```


BethanyBerkowitz:
Checked that the new tests ran in CI. There's an issue with NextJS example app deployment which is not related to this PR.

packages/react-native/test/iosUtils.test.ts
```diff
@@ -0,0 +1,125 @@
...
+} from '../src/iosUtils'
+
+/* TODO from Bethany June 2023:
+ * The iosUtils code came from Andrey's original code and I took pains
+ * not to mess with how it works. However, in my testing, I've only found 
+ * backtraces generated using framesFromIOSCallStack. I've never seen an
+ * error reported with primaryBackTraceSource = 'ReactNativeComponentStack'
+ * or primaryBackTraceSource = 'ReactNativeIOSStackTrace'
+ * 
+ * Need to check with Andrey if he remembers how to generate an error that
```
BethanyBerkowitz, line 13: @andreybutov do you happen to remember how to trigger these code paths?

subzero10:
> Checked that the new tests ran in CI. There's an issue with NextJS example app deployment which is not related to this PR.

The nextjs example app deployments should not be running with these PRs 🤔 .
I disconnected them now so it should be OK.

packages/react-native/test/iosUtils.test.ts
subzero10: @andreybutov Could it be related to this change from #1069:
> No longer pass to built-in error handler after notifying in release build. This removes the duplicate error reporting documented in #1054

andreybutov: @subzero10 Yes, perhaps. I'll take a look.

BethanyBerkowitz: @andreybutov Yes, javascript errors are currently just going through the JS handler and not passed to the built-in handler in prod because previously those errors were getting double reported. I'm definitely open to handling this differently and appreciate any thoughts you have on it. 

This question can also be broken out into another github issue so we can make this PR just about testing setup -- I think just getting some basic Jest tests running is enough for one PR! I just didn't want to lose track of it.

andreybutov: @BethanyBerkowitz @subzero10 I agree. Let’s get the testing stuff through, and let’s open a new issue for the double-reporting / native handling, and I’ll take a look at that separately.

BethanyBerkowitz: Awesome, thanks! I added https://github.com/honeybadger-io/honeybadger-js/issues/1099 and cleaned up this code comment.

subzero10 (approved):

subzero10:
Thank you @BethanyBerkowitz!

Hopefully with the added unit tests we will go away from the discussions of "was this supposed to work like this?" or "is this functionality necessary/redundant?"

packages/react-native/test/iosUtils.test.ts
subzero10: Thank you @andreybutov. I assigned #1099 to you 👌.

PR #1094, opened 2023-06-16
chore(monorepo): detect changes that would cause major version bump

subzero10:
## Status
**READY**

## Description
Detect if PR will introduce breaking changes which will cause a major version upgrade.

## Todos
- [x] Add comment if PR title has `!` or `BREAKING CHANGE`
- [x] Install and configure [Task List Completed](https://github.com/marketplace/task-list-completed) app

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -35,7 +35,7 @@ jobs:
           echo "${{ github.event.pull_request.title }}"
 
       - name: Check PR name for breaking changes
-        if: github.event_name == 'pull_request' && contains(github.event.pull_request.title, 'BREAKING CHANGE')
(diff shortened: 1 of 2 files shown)


.github/workflows/commitlint.yml
```diff
@@ -24,3 +24,20 @@ jobs:
       - name: Verify PR title is in the correct format
         run: |
           echo "${{ github.event.pull_request.title }}" | npx commitlint -V
+
+  needs-migration-docs-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Echo event name
```
subzero10, line 31: I added this dummy step, because I noticed the job wouldn't execute 🤷

PR #1105, opened 2023-06-22
chore(react-native): add testing for the client 

### LINKED ISSUE #1103: [react-native] Write unit tests for index.ts
In #902 we set up Jest unit testing. However, there are still no tests for the main entrypoint file, `index.ts`.

BethanyBerkowitz:
## Status
READY

## Description
Closes #1103 

## Steps to Test or Reproduce
```bash
cd packages/react-native
npm test
```

### DIFF
--- a/packages/react-native/src/index.ts
+++ b/packages/react-native/src/index.ts
@@ -92,12 +92,12 @@ class Honeybadger extends Client {
 
   private onNativeException(data:NativeExceptionData) {
     switch ( Platform.OS ) {
-      case 'ios': 
-        this.onNativeIOSException(data)
-        break
-      case 'android': 
-        this.onNativeAndroidException(data)
-        break
+    case 'ios': 
+      this.onNativeIOSException(data)
+      break
+    case 'android': 
+      this.onNativeAndroidException(data)
+      break
     }
   }
 
@@ -122,12 +122,6 @@ class Honeybadger extends Client {
     const notice = {
       name: `React Native Android ${data.type}`,
       message: data.message || '',
-      details: {
-        errorDomain: data.errorDomain || '',
-        initialHandler: data.initialHandler || '',
-        userInfo: data.userInfo || {},
-        architecture: data.architecture || '',
-      },
       backtrace: backtraceFromAndroidException(data)
     }
     this.notify(notice)
--- a/packages/react-native/src/index.ts
+++ b/packages/react-native/src/index.ts
@@ -50,20 +50,24 @@ class Honeybadger extends Client {
     this.__originalJsHandler = ErrorUtils.getGlobalHandler();
 
     ErrorUtils.setGlobalHandler((err, isFatal) => {
-      this.logger.debug('JavaScript global error handler triggered.')
-      this.notify(err)
-
-      // Allowing the default error handler to process the error after
-      // we're done with it will show the useful RN red info box in dev.
-      // However in prod it causes duplicate errors to be reported on iOS
-      // since the default handler re-throws a native error
-      if (__DEV__ && this.__originalJsHandler) {
-        this.logger.debug('Passing error to previous error handler.')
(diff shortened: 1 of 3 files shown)


packages/react-native/src/index.ts
```diff
@@ -118,12 +122,6 @@ class Honeybadger extends Client {
     const notice = {
       name: `React Native Android ${data.type}`,
       message: data.message || '',
-      details: {
```
BethanyBerkowitz, line 121: This was copied from the original package, but I just removed the `details` here because I took a look at [the java code that emits the event](https://github.com/honeybadger-io/honeybadger-js/blob/f0d361720144297d20dc7d08712f0ec5208aa564/packages/react-native/android/src/main/java/io/honeybadger/HoneybadgerReactNativeModule.java#L76C28-L76C28) and I don't think any of these fields will ever be populated.

packages/react-native/src/transport.ts
```diff
@@ -33,7 +33,8 @@ export class Transport implements Types.Transport {
   }
 
   private buildUserAgent() {
-    const reactNativeVersion = `${Platform.constants.reactNativeVersion.major}.${Platform.constants.reactNativeVersion.minor}.${Platform.constants.reactNativeVersion.patch}`
+    const rnVersion = Platform.constants?.reactNativeVersion
+    const rnVersionStr = `${rnVersion?.major}.${rnVersion?.minor}.${rnVersion?.patch}`
     
     let nativePlatform:string
```
subzero10, line 39:
```suggestion
    let nativePlatform: string
```
🤔 Shouldn't this get picked up from eslint?

packages/react-native/src/transport.ts
```diff
@@ -33,7 +33,8 @@ export class Transport implements Types.Transport {
   }
 
   private buildUserAgent() {
-    const reactNativeVersion = `${Platform.constants.reactNativeVersion.major}.${Platform.constants.reactNativeVersion.minor}.${Platform.constants.reactNativeVersion.patch}`
+    const rnVersion = Platform.constants?.reactNativeVersion
+    const rnVersionStr = `${rnVersion?.major}.${rnVersion?.minor}.${rnVersion?.patch}`
```
subzero10, line 37: How about we at least check for `rnVersion`?
```suggestion
    const rnVersionStr = rnVersion ? `${rnVersion?.major}.${rnVersion?.minor}.${rnVersion?.patch}` : 'n/a'
```

subzero10 (commented):
This looks great!

Just a couple non-blocking comments:
- can we add tests for breadcrumbs and context being sent along with `notify()`?
- can we add tests with a `beforeNotifyHandler` and an `afterNotifyHandler`?

packages/react-native/src/transport.ts
BethanyBerkowitz: I confirmed that this file is getting linted. 

Looks like [type-annotation-spacing](https://typescript-eslint.io/rules/type-annotation-spacing/) is not part of `@typescript-eslint/recommended`. They want you to use Prettier or another formatter instead. I'm also partial to using Prettier, but that is too big a task for this PR, so I'll just go ahead and enable this rule for now.

BethanyBerkowitz:
@subzero10 I added tests for `beforeNotify`, `afterNotify`, `setContext` and `addBreadcrumb` -- thanks for the suggestion, I think they add confidence as well as documentation. They're really more integration tests than unit tests... I'm happy leaving them where they are, but lmk if you'd rather separate them out.

.eslintrc
```diff
@@ -22,6 +22,7 @@
       "error",
       "always"
     ],
+    "@typescript-eslint/type-annotation-spacing": 1,
```
subzero10, line 25: ❤️

subzero10 (approved):
👏

Issue #1097, opened 2023-06-20
Errors reported from different javascript clients share the same notifier


subzero10:
## What are the steps to reproduce this issue?
<img width="1129" alt="image" src="https://github.com/honeybadger-io/honeybadger-js/assets/5813382/071762fb-ea18-4175-b5b7-8956c76a67d5">
The error shown above was reported from the nextjs package.

## What happens?
The client shown in Honeybadger dashboard appears to be the js package.

## What were you expecting to happen?
The client shown in Honeybadger dashboard should be the nextjs package.

## Any logs, error output, etc?
n/a

## Any other comments?
This is most probably because our javascript integrations use the js package to report errors.

## What versions are you using?
**Package Name:**  @honeybadger-io/nextjs
**Package Version:** 5.6.1

BethanyBerkowitz:
@subzero10 This is also true for the React package -- here's an error reported from that. Is there a chance this was the desired behavior? I have no real preference... but if we update this for one package, we should check all the packages. 

<img width="856" alt="Screen Shot 2023-06-20 at 2 01 15 PM" src="https://github.com/honeybadger-io/honeybadger-js/assets/16493493/f4b67e20-fbfe-4b2b-aa18-9a5a71685e84">

The react-native package does show its own version, but that's because it depends on `core`, not on `js`. 
<img width="856" alt="Screen Shot 2023-06-20 at 2 04 04 PM" src="https://github.com/honeybadger-io/honeybadger-js/assets/16493493/be0e7453-2069-4ee1-89f4-c9d99b4ec79c">

joshuap:
I think we'd need to have a custom `notifier` entry for each sub package for this to work.

BethanyBerkowitz:
> I think we'd need to have a custom `notifier` entry for each sub package for this to work.

@joshuap agreed, we'd also need to add the ability to set a notifier on the client -- currently the `js` package doesn't accept a notifier and just adds it during `__buildPayload`. 

So @subzero10 I don't think this is a bug, it's a feature request. If this would be useful for us to have for debugging, analytics etc, I'm totally down to do it. My guess is that users don't much care what the notifier shows though.

subzero10:
Hey @BethanyBerkowitz, 

> So @subzero10 I don't think this is a bug, it's a feature request. If this would be useful for us to have for debugging, analytics etc, I'm totally down to do it. My guess is that users don't much care what the notifier shows though.

[...]

PR #1107, opened 2023-06-23
feat: specify notifier per package

subzero10:
## Status
**READY**

## Description
Closes: #1097

## Todos
- [x] js
- ~~nextjs~~ Next.js uses the react client to report errors
- [x] react
- [x] vue
- [x] react-native

### DIFF
--- a/packages/js/src/browser.ts
+++ b/packages/js/src/browser.ts
@@ -88,7 +88,10 @@ class Honeybadger extends Client {
 
   public factory(opts?: Partial<Types.BrowserConfig>): this {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    return new Honeybadger(opts) as any
+    const clone = new Honeybadger(opts) as any
+    clone.setNotifier(this.getNotifier())
+
+    return clone
   }
 
   public checkIn(_id: string): Promise<void> {
@@ -244,9 +247,7 @@ class Honeybadger extends Client {
   }
 }
 
-export { Types } from '@honeybadger-io/core'
-
-export default new Honeybadger({
+const singleton = new Honeybadger({
   __plugins: [
     onError(),
     onUnhandledRejection(),
@@ -255,3 +256,14 @@ export default new Honeybadger({
     breadcrumbs()
   ]
 })
+
+const NOTIFIER = {
+  name: '@honeybadger-io/js',
+  url: 'https://github.com/honeybadger-io/honeybadger-js/tree/master/packages/js',
+  version: '__VERSION__'
+}
+
+singleton.setNotifier(NOTIFIER)
+
+export { Types } from '@honeybadger-io/core'
+export default singleton
--- a/packages/js/test/unit/browser.test.ts
+++ b/packages/js/test/unit/browser.test.ts
@@ -165,6 +165,23 @@ describe('browser client', function () {
         })
       })
     })
+
+    it('uses the correct notifier name', function () {
+      return new Promise<void>((resolve) => {
+        client.configure({
+          apiKey: 'testing'
+        })
+
+        client.notify('testing')
+
+        setTimeout(() => {
+          // @ts-expect-error
+          const payload = JSON.parse(fetch.mock.lastCall[1].body)
+          expect(payload.notifier.name).toEqual('@honeybadger-io/js')
+          resolve()
+        })
+      })
+    })
   })
 
   describe('showUserFeedbackForm', function () {
--- a/packages/js/test/unit/browser.test.ts
+++ b/packages/js/test/unit/browser.test.ts
@@ -113,74 +113,58 @@ describe('browser client', function () {
       expect(result).toEqual(false)
     })
 
-    it('excludes cookies by default', function () {
-      return new Promise<void>((resolve) => {
-        client.configure({
-          apiKey: 'testing'
-        })
-
-        client.notify('testing')
-
-        setTimeout(() => {
-          // @ts-expect-error
-          const payload = JSON.parse(fetch.mock.lastCall[1].body)
-          expect(payload.request.cgi_data.HTTP_COOKIE).toBeUndefined()
-          resolve()
-        })
+    it('excludes cookies by default', async function () {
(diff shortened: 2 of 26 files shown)


packages/js/src/browser.ts
```diff
@@ -88,7 +88,10 @@ class Honeybadger extends Client {
 
   public factory(opts?: Partial<Types.BrowserConfig>): this {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    return new Honeybadger(opts) as any
+    const clone = new Honeybadger(opts) as any
+    clone.setNotifier(this.getNotifier())
```
subzero10, line 92: This is important. I could go for a _static_ `__notifier` that is shared for all instances of the client, but I chose this method for more flexibility.

BethanyBerkowitz: I'm fine with either solution, but is flexibility in this case actually a benefit? Would you ever want an instance of the class to have a different notifier? This is one case where less flexibility may actually be better -- we want to know for sure what package the notice came from and reduce the possible ways to mess with that.

packages/js/test/unit/browser.test.ts
```diff
@@ -165,6 +165,23 @@ describe('browser client', function () {
...
+
+    it('uses the correct notifier name', function () {
+      return new Promise<void>((resolve) => {
+        client.configure({
+          apiKey: 'testing'
+        })
+
+        client.notify('testing')
+
+        setTimeout(() => {
```
BethanyBerkowitz, line 177: Huh, interesting trick! So using `setTimeout` forces this code to execute after `notify()` is complete, even though there's no value set for the delay? This is weird and I'm not sure I like it 😆 

But looks like the rest of the tests in here use this already, so no reason to change that with this PR!

packages/react-native/src/index.ts
```diff
@@ -10,15 +10,21 @@ class Honeybadger extends Client {
   private __nativeHandlerInitialized:boolean
   private __originalJsHandler:(error: Error, isFatal: boolean) => void
 
+  protected override __notifier = {
+    name: '@honeybadger-io/react-native',
+    url: 'https://github.com/honeybadger-io/honeybadger-js/tree/master/packages/react-native',
+    version: '__VERSION__'
```
BethanyBerkowitz, line 16: Need to add a rollup plugin for this to work, no? 

Also the `react-native` package currently sets a notifier within `transport.ts` -- that code can now be removed, correct?

BethanyBerkowitz (commented):
May need changes to `react-native`, otherwise 🚀

packages/js/test/unit/browser.test.ts
subzero10: I copy-pasted an existing test, but here's what I remember:
- `client.notify()` makes the http request in a promise
- A promise, even if it resolves immediately, won't be executed on the same stack frame
- In order to make sure that the assertions are done after the promise execution we use setTimeout. The code inside setTimeout will be executed in a separate stack frame, after the promise is executed.

I came up with this workaround when I refactored the code from callbacks to promises, but there may be better solutions!

packages/js/src/browser.ts
subzero10: You are correct, the notifier is not something we will encourage users to play with (though I'm not sure if there are any negative implications if they do?).

However, I had to make the `setNotifier` method public because of the way we export the js package (a singleton instance).
By making this method publicly accessible, we are prone to inconsistencies if we make the notifier as a static shared variable.
Consider this use case:
```
import hbClient from '@honeybadger-io/js'
hbClient.configure()

const anotherHbClient = hbClient.factory()
anotherHbClient.setNotifier({name: 'manual-notifier'})
```
With a static shared variable, both clients would have the same notifier name (manual-notifier).

packages/react-native/src/index.ts
subzero10: Yes, nice catch!

packages/js/test/unit/browser.test.ts
BethanyBerkowitz: In my react-native tests, I put the assertions into an `afterNotify`... which is ok but also not as obvious and readable as I would prefer. If using `notifyAsync` is an acceptable alternative, you could make the test async and `await` it. But like I mentioned, I don't think you need to make any changes in this PR since the rest of the file is also set up like this. You could add a comment to the first test that uses this trick if you feel like being extra generous to the next person who sees this file for the first time 😆

BethanyBerkowitz:
Also @subzero10 we've been having issues with the `task-list-completed` task getting hung up. This one has been stuck for a few days. 

Screen Shot 2023-06-29 at 1 04 17 PM


The developer is aware, see https://github.com/stilliard/github-task-list-completed/issues/34#issuecomment-1609886196.

packages/js/test/unit/browser.test.ts
subzero10: Good point, we could use `notifyAsync` here. `notifyAsync` came after these tests 😅 .

I'll see if I can change these to notifyAsync.

subzero10: Done!

packages/react-native/src/index.ts
subzero10: Done!

subzero10:
@BethanyBerkowitz Can I get another review when you get the chance? Thanks!

packages/js/test/unit/browser.test.ts
subzero10, line 1: I replaced setTimeout with async/await in this test file.

subzero10:
> Also @subzero10 we've been having issues with the `task-list-completed` task getting hung up. This one has been stuck for a few days.
> 
> Screen Shot 2023-06-29 at 1 04 17 PM
> The developer is aware, see [stilliard/github-task-list-completed#34 (comment)](https://github.com/stilliard/github-task-list-completed/issues/34#issuecomment-1609886196).

Oh, it's probably because I left a checkbox unchecked 😅 
image

Issue #1129, opened 2023-07-12
`honeybadger-copy-config-files` doesn't check `src/`


mikew:
## What are the steps to reproduce this issue?
1. move next folders to `src/`
2. run `npx honeybadger-copy-config-files`

## What happens?
No `_error.tsx` is generated.

## What were you expecting to happen?
An `_error.tsx` to be generated in `src/pages/`

## Any logs, error output, etc?
…

## Any other comments?
…

## What versions are you using?
**Operating System:** …
**Package Name:** …
**Package Version:** …
**Browser Version:** …

subzero10:
Hey @mikew, thanks for reporting an issue!

If I understand correctly, you are using the optional src directory, as mentioned in the [documentation](https://nextjs.org/docs/app/building-your-application/configuring/src-directory).
We can update the script to account for that.

While you wait, you can do the following as a workaround:
- create a `pages` folder in the project root
- run `npx honeybadger-copy-config-files`
- move the new `pages/_error.tsx` file to `src/pages/_error.tsx`
- remove `pages` folder from root

subzero10:
@mikew This should be available with [@honeybadger-io/nextjs@5.7.0](https://github.com/honeybadger-io/honeybadger-js/releases/tag/%40honeybadger-io%2Fnextjs%405.7.0)!

PR #1130, opened 2023-07-14
feat(nextjs): support project structure under src folder

subzero10:
## Status
**READY**

## Description
Closes: #1129.

## TODO
- [x] Update script to check for `src` folder
- [x] Tests

### DIFF
--- a/packages/nextjs/src/copy-config-files.test.ts
+++ b/packages/nextjs/src/copy-config-files.test.ts
@@ -1,0 +1,180 @@
+import mock from 'mock-fs'
+import fs from 'fs'
+import path from 'path'
+import { copyConfigFiles } from './copy-config-files';
+describe('copy-config-files', () => {
+
+  afterEach(() => {
+    mock.restore()
+  })
+
+  it('should copy config files to a project with pages router', () => {
+    mock({
+      'templates': mock.load(path.resolve(__dirname, '..', 'templates')),
+      'pages': {
+        'index.js': 'dummy content'
+      }
+    })
+
+    return copyConfigFiles()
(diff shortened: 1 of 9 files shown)


packages/nextjs/src/copy-config-files.test.ts
```diff
@@ -0,0 +1,180 @@
...
+import fs from 'fs'
+import path from 'path'
+import { copyConfigFiles } from './copy-config-files';
+describe('copy-config-files', () => {
+
+  afterEach(() => {
+    mock.restore()
+  })
+
+  it('should copy config files to a project with pages router', () => {
```
BethanyBerkowitz, line 11: Thanks for adding tests!

Can we default to using `async/await` instead of `.then()`?

BethanyBerkowitz (dismissed):
Nice! Minor comment, don't need to re-review.

BethanyBerkowitz:
@subzero10 Do we need to continue with unit test for node 14? Looks like 14 is no longer supported: https://github.com/nodejs/release#release-schedule

packages/nextjs/src/copy-config-files.test.ts
subzero10: 👍 Done!

subzero10:
> @subzero10 Do we need to continue with unit test for node 14? Looks like 14 is no longer supported: https://github.com/nodejs/release#release-schedule

That'a good question. Seems like we can stop supporting Node 14, though we'd have to do a major version bump right? 🤔 Because as soon as we remove from CI, then we won't know when some piece of code would break on Node 14.

Issue #960, opened 2022-11-26
`uncaughtException` handler not working as expected with latest release


subzero10:
While manually testing `v4.7.4` I noticed that for `uncaughtException`, we register 2 listeners (at least that's how it appears to be so):
- `domainUncaughtExceptionClear`
- `honeybadgerUncaughtExceptionListener`

Moreover, the Honeybadger [instance sets `fatallyLogAndExit` as the `afterUncaught` handler](https://github.com/honeybadger-io/honeybadger-js/blob/master/packages/js/src/server.ts#L41), which exits the process regardless the number of listeners registered.

I would like to:
- investigate where this `domainUncaughtExceptionClear` listener comes from 
- remove the `afterUncaught: fatallyLogAndExit` assignment
- and spend some time to see if I can write some tests for our logic in these listeners.

BethanyBerkowitz:
One interesting option here - the [`uncaughtExceptionMonitor` event](https://nodejs.org/api/process.html#event-uncaughtexceptionmonitor) was added in Node 12/13 (so it should be available in all of our currently supported node versions). It doesn't override the default behavior, so it seems like our logic for when to exit vs not could be removed if we used this event. However, we'd want to do really thorough testing before making a change like this since catching the uncaught exceptions is such an important baseline feature.

BethanyBerkowitz:
The listener `domainUncaughtExceptionClear` is apparently something attached [by node itself](https://github.com/nodejs/node/blob/6432060c7ce1670467c230d2c0c925465dd311f8/lib/domain.js#L207). 

It's confusing that this listener is still attached, considering that [domains appear to have been deprecated years and years ago](https://nodejs.org/api/domain.html). 

@subzero10 do you remember where you saw this listener popping up? Was it in one of our test projects? I'm curious when and why it is still attached if domains are deprecated.

subzero10:
> One interesting option here - the [uncaughtExceptionMonitor event](https://nodejs.org/api/process.html#event-uncaughtexceptionmonitor) was added in Node 12/13 (so it should be available in all of our currently supported node versions). It doesn't override the default behavior, so it seems like our logic for when to exit vs not could be removed if we used this event. However, we'd want to do really thorough testing before making a change like this since catching the uncaught exceptions is such an important baseline feature.
> 
> [edit] I also haven't yet looked into how this works with potentially async code -- we don't want the process to shut down before the notification is sent off.

[...]

PR #1161, opened 2023-08-10
fix(js): remove default `afterUncaught` and add tests for our unhandled node error listener

BethanyBerkowitz:
## Status
READY

## Description
Part of https://github.com/honeybadger-io/honeybadger-js/issues/960. 

Adding some unit testing for the listener we attach for unhandled exceptions. This should allow us to better understand how this code works and be more confident if we decide to make changes. 

This also removes the default `afterUncaught` in the server client, which was set to `fatallyLogAndExit`. Having that default there would mean that our logic for when to exit vs not based on the number of listeners was not actually being executed most of the time, and instead we were just always exiting after notifying. The default in the base `Client` class is just `() => true`, so we'll now default to that. 

## Todos
- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce
```bash
cd packages/js
npm run test:server
```


packages/js/src/server/integrations/uncaught_exception.ts
```diff
@@ -25,49 +25,59 @@ function hasOtherUncaughtExceptionListeners() {
...
 }
 
+function handleUncaughtException(uncaughtError: Error, client: typeof Client) {
+  if (isReporting) { 
+    return 
+  }
+
+  // In the "load" function in the exported plugin below, we 
+  // never attach this listener if enableUncaught is false...
+  // so why do we need this?
```
BethanyBerkowitz, line 35: Without knowing the purpose of this piece of code I am afraid to remove it. I think it might be intended for the case in which `client.config.enableUncaught` changes after the initial setup, ie the user configured Honeybadger with default config first, then ran `honeybadger.configure({ enableUncaught: false })`. It seems that [plugins will only be loaded once](https://github.com/honeybadger-io/honeybadger-js/blob/c41074ec5e78c2992ab67a171dc2a2ef91927a65/packages/core/src/client.ts#L121-L122), so that could make sense. 

Even if this is the reasoning, I still think it's a little odd to run `afterUncaught` if `enableUncaught` is false. As a user, I don't think I'd expect `afterUncaught` to run if I had disabled uncaught error handling entirely. 

Also I don't think it would work the other way... if the user set up Honeybadger with `enableUncaught = false` and later ran `honeybadger.configure({ enableUncaught: true })`, it wouldn't start reporting errors because `load()` already ran and won't run again, so no listener is attached. 

I haven't tried testing these different scenarios. Before I go down a rabbit hole, I wanted to just see if anyone else (@subzero10?) had insight.

packages/js/src/server/integrations/uncaught_exception.ts
```diff
@@ -25,49 +25,59 @@ function hasOtherUncaughtExceptionListeners() {
...
-            }
-          }
-        })
+        handleUncaughtException(uncaughtError, client)
       })
     }
   }
 }
+
+export const exportedForTesting = {
```
BethanyBerkowitz, line 79: Not everyone likes this way of getting access to functions for unit testing -- I'm open to other suggestions. 

`rewire` or another library like that is probably an option if folks don't like this, it's just more setup.

packages/js/src/server/integrations/uncaught_exception.ts
subzero10: > I think it might be intended for the case in which client.config.enableUncaught changes after the initial setup, ie the user configured Honeybadger with default config first, then ran honeybadger.configure({ enableUncaught: false }).

Yes, I think this is the reason as well.

> Also I don't think it would work the other way... if the user set up Honeybadger with enableUncaught = false and later ran honeybadger.configure({ enableUncaught: true }), it wouldn't start reporting errors because load() already ran and won't run again, so no listener is attached.

You are correct! Plugins are loaded only once, no matter how many times you call `configure`.

Based on the above, we could:
- [my preference] change the logic to reload this plugin if `enableUncaught` is changed after plugins are loaded
- or always attach the listener and report only if `enableUncaught` is true

BUT, I would first have another look at the `uncaughtExceptionMonitor`, which: - captures both exceptions and promise rejections
- allows us to avoid checking for number of listeners and deciding whether to exit or not

packages/js/src/server/integrations/uncaught_exception.ts
subzero10: 🤔 Not a fan of this either, but I see why you have to do it. How about we just export these functions in the normal way?
```
export function handleUncaughtException() { ... }
export function setIsReporting() { ... }
export function setHandlerAlreadyCalled() { ... }
```
And you can use `setIsReporting`, `setHandlerAlreadyCalled` inside this file as well, instead of modifying it directly.

packages/js/test/unit/server/integrations/uncaught_exception.server.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        expect(listeners.length).toBe(1)
+        expect(listeners[0].name).toBe('honeybadgerUncaughtExceptionListener') 
+
+        const error = new Error('uncaught')
+        process.emit('uncaughtException', error)
+        expect(notifySpy).toHaveBeenCalledTimes(1)
+      })
+
+      it('removes the AWS lambda uncaught exception listener', () => {
+        const restoreEnv = process.env
```
subzero10, line 59: Does this give you a copy of `process.env`? Or is it a reference?

subzero10 (commented):
Finally, I think I would go for a `fix` as the commit type, since you are removing the default use of `fatallyLogAndExit`.

packages/js/src/server/integrations/uncaught_exception.ts
BethanyBerkowitz: In general it's not ideal to export anything that isn't intended to be used by other modules -- a later developer may think it's exported to be used in the code when it's actually just exported to be used in a test. For example, it could cause issues if someone used `setIsReporting` from some other file. 

In this case, I'm honestly not worried about it either way -- it's a small-ish package and I don't think there's likely to be confusion. Also if we are considering the switch to `uncaughtExceptionMonitor`, a lot of this will get re-worked.

packages/js/test/unit/server/integrations/uncaught_exception.server.test.ts
BethanyBerkowitz: Ah good catch, I forgot the braces! Added.

packages/js/src/server/integrations/uncaught_exception.ts
BethanyBerkowitz: Makes sense to me! In that case let's not change anything with this PR and instead look at the `uncaughtExceptionMonitor` issue first.

packages/js/src/server/integrations/uncaught_exception.ts
subzero10: > In general it's not ideal to export anything that isn't intended to be used by other modules

I agree. But exporting here does not make these functions available to other modules.

> In this case, I'm honestly not worried about it either way -- it's a small-ish package and I don't think there's likely to be confusion.

My concern is that we don't set a precedence and start testing our code this way, i.e. exporting for the sake of tests. It makes me think that there is a better solution out there.
I see 2 possible solutions:
- use rewire
- rewrite this piece of logic so that it can be mocked, for example we could create a `UncaughtExceptionProcessor` class in a separate file wrapped with this logic. Then we can mock the entire file.

> Also if we are considering the switch to uncaughtExceptionMonitor, a lot of this will get re-worked.

True. But as your [investigation](https://github.com/honeybadger-io/honeybadger-js/issues/960#issuecomment-1675320288) has revealed, we may be sticking with this code after all 😅 .

packages/js/src/server/integrations/uncaught_exception.ts
BethanyBerkowitz: Added https://github.com/honeybadger-io/honeybadger-js/issues/1163 for this. I'll also clean up the comment in here.

packages/js/src/server/integrations/uncaught_exception.ts
BethanyBerkowitz: @subzero10 I went ahead and made a separate class for this -- I think that's a reasonable choice anyway since most of this package is structured as classes. I now have an integration test failure that I'm looking into (it appears unrelated at first glance).

subzero10: Thanks for doing that! The updated code seems much cleaner 👌

subzero10 (approved):

Issue #1165, opened 2023-08-17
[js] update structure of unhandled rejection listener to match uncaught exception


BethanyBerkowitz:
In https://github.com/honeybadger-io/honeybadger-js/pull/1161, we moved the unhandled exception listener into a class and added testing to it. We should do the same with the unhandled rejection listener.

PR #1166, opened 2023-08-18
fix(js): move unhandled rejection monitor to a class, fix filter for domainUncaughtExceptionClear

BethanyBerkowitz:
## Status
READY

## Description
Closes https://github.com/honeybadger-io/honeybadger-js/issues/1165

Makes the format of the unhandled rejection listener match that of the uncaught exception listener and adds unit tests. 

I also noticed that the filter for domainUncaughtExceptionClear needed unit tests and was missing the `return` keyword.

## Todos
- [x] Tests

## Steps to Test or Reproduce
```bash
npm run test:server
```


packages/js/test/unit/server/integrations/unhandled_rejection_monitor.server.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        { component: 'unhandledRejection' },
+        { afterNotify: expect.any(Function) }
+      )
+      client.afterNotify(() => {
+        expect(fatallyLogAndExitSpy).toHaveBeenCalledWith(reason)
+        done()
+      }) 
+    })
+  
+    // Seems like it was meant to work this way but was never implemented?
```
BethanyBerkowitz, line 60: This is easy to add to make it match how `uncaughtException` works (eg return right away if `__isReporting` is `true`... but I'm curious if there's a reason it doesn't include that? Also was trying to make a `chore` PR rather than a `fix` PR.

packages/js/src/server/integrations/uncaught_exception_monitor.ts
```diff
@@ -31,7 +31,7 @@ export default class UncaughtExceptionMonitor {
     // Node sets up these listeners when we use domains
     // Since they're not set up by a user, they shouldn't affect whether we exit or not
     const domainListeners = allListeners.filter(listener => {
-      listener.name === 'domainUncaughtExceptionClear' 
+      return listener.name === 'domainUncaughtExceptionClear'
```
subzero10, line 34: Nice catch!

packages/js/src/server/integrations/unhandled_rejection_monitor.ts
```diff
@@ -0,0 +1,42 @@
...
+   * If there are no other unhandledRejection listeners,
+   * we want to report the exception to Honeybadger and
+   * mimic the default behavior of NodeJs,
+   * which is to exit the process with code 1
+   */
+  hasOtherUnhandledRejectionListeners() {
+    return process.listeners('unhandledRejection').length > 1
+  }
+
+  handleUnhandledRejection(reason: unknown, _promise: Promise<unknown>) {
```
subzero10, line 24: Is `_promise` not being used here?

packages/js/src/server/integrations/unhandled_rejection_plugin.ts
```diff
@@ -0,0 +1,17 @@
...
+import UnhandledRejectionMonitor from './unhandled_rejection_monitor'
+
+export default function (): Types.Plugin {
+  return {
+    load: (client: typeof Client) => {
+      if (!client.config.enableUnhandledRejection) {
+        return
+      }
+      const unhandledRejectionMonitor = new UnhandledRejectionMonitor(client)
+      process.on('unhandledRejection', function honeybadgerUnhandledRejectionListener(reason, _promise) {
```
subzero10, line 12: FYI: We usually prefix a parameter with `_` so that eslint won't pick it up and say that _parameter is not used_ (https://github.com/honeybadger-io/honeybadger-js/blob/master/.eslintrc#L33).

subzero10 (approved):
❤️ I love seeing all these tests!

packages/js/test/unit/server/integrations/unhandled_rejection_monitor.server.test.ts
subzero10: 🤔 I don't remember. So, now it was **changed** to return is `__isReporting` is `true`?

packages/js/src/server/integrations/unhandled_rejection_plugin.ts
BethanyBerkowitz: I tried to make as little change as possible to the existing code with this PR. In the existing code, the param was called `_promise` and was unused, so I just left it that way. 

I guess technically this one is now "used" in that it's passed through... but it's passed through to a function that doesn't use it, so it seemed more consistent to keep the naming the same throughout.

packages/js/src/server/integrations/unhandled_rejection_monitor.ts
BethanyBerkowitz: See https://github.com/honeybadger-io/honeybadger-js/pull/1166/files#r1300209911

packages/js/test/unit/server/integrations/unhandled_rejection_monitor.server.test.ts
BethanyBerkowitz: Here's an issue to review this! https://github.com/honeybadger-io/honeybadger-js/issues/1167

packages/js/src/server/integrations/unhandled_rejection_plugin.ts
subzero10: OK, thanks for the details. We can remove this on the next iteration, to avoid the confusion in the future.

Issue #1163, opened 2023-08-14
Reload uncaughtException plugin if enableUncaught changes, same with unhandledRejection


BethanyBerkowitz:
See https://github.com/honeybadger-io/honeybadger-js/pull/1161#discussion_r1291336826. 

Currently, the listeners for `uncaughtException` and `unhandledRejection` are only added if the corresponding setting is `true`. Then the listener decides not to report if the user later changes the setting to `false`. This is reasonable, but inconsistent because if the user sets up Honeybadger with the setting off, then later flips it to `true`, the plugin will not be reloaded and the listener will never be attached. 

We prefer to reload the plugin if `enableUncaught` changes. (ie remove the listener if the value is set to false and add it if it's set to true). Same deal with `enableUnhandledRejection`. 

- [x] make uncaughtException plugin able to be loaded more than once (js package) https://github.com/honeybadger-io/honeybadger-js/pull/1170
- [x] make unhandledRejection plugin able to be loaded more than once (js package) https://github.com/honeybadger-io/honeybadger-js/pull/1172 
- [x] reload the plugins when the client is re-configured based on a setting (core package) https://github.com/honeybadger-io/honeybadger-js/pull/1174
- [x] update core pkg version that the js package uses, turn on setting to reload (js package). Can also remove duplicate checks for enableUncaught and enableUnhandledRejection at this point https://github.com/honeybadger-io/honeybadger-js/pull/1184
- [x] update the docs: https://github.com/honeybadger-io/docs/pull/334

subzero10:
> reload the plugins when the corresponding config option changes (core package)

Have you thought about this? What if we added a property on the Plugin type, such as `shouldReloadOnConfigChange` or something like it?

BethanyBerkowitz:
> > reload the plugins when the corresponding config option changes (core package)
> 
> Have you thought about this? What if we added a property on the Plugin type, such as `shouldReloadOnConfigChange` or something like it?

Yes, that's what I was thinking! Just a boolean (default false) for whether to reload or not. 

Another option would be to add an array of config keys that should cause a reload. We'd load everything the first time (ie when `__pluginsExecuted` is false. Then reload only if one of the designated keys was passed in to the `configure()` function. This would be more complex and I doubt that users are calling `configure()` all that often... so it seems like maybe over-optimizing.

subzero10:
> Just a boolean (default false) for whether to reload or not.

[...]

PR #1170, opened 2023-08-24
chore(js): allow uncaughtException plugin to be loaded more than once

BethanyBerkowitz:
## Status
**READY**

## Description
Part of https://github.com/honeybadger-io/honeybadger-js/issues/1163

This PR makes it safe to load the uncaughtException plugin more than once. 

- If loaded with enableUncaught true, it will add a listener if one has not already been added
- If loaded with enableUncaught false, it will remove the listener if it exists

However, the client itself still only loads the plugin once, so this PR should not cause any user-facing change in behavior. 

## Todos
- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce
Unit tests should mostly cover it. I also tested it by throwing an uncaught error in one of the example applications. I chose `sails`, but it could have been `express` or another node project. The trick is that you have to throw an error in a node application that is not covered by the honeybadger middleware, but also doesn't cause an import to fail entirely. So for example: 

```
Honeybadger.configure({
  enableUncaught: true,
})

setTimeout(() => {
  throw new Error('unhandled err at top-level')
}, 50)
```

Also tested with `enableUncaught: false`, which did not send a report (as expected). 

I also found that you can't test this in AWS Lambda because the wrapper just wraps the whole thing and calls notify(). So if you throw an uncaught error within your lambda, it’s just going to get reported up to the wrapper.

### DIFF
--- a/packages/js/test/unit/server/integrations/uncaught_exception_monitor.server.test.ts
+++ b/packages/js/test/unit/server/integrations/uncaught_exception_monitor.server.test.ts
@@ -4,6 +4,10 @@ import Singleton from '../../../../src/server'
 import UncaughtExceptionMonitor from '../../../../src/server/integrations/uncaught_exception_monitor'
 import * as aws from '../../../../src/server/aws_lambda'
 
+function getListenerCount() {
+  return process.listeners('uncaughtException').length
+}
+
 describe('UncaughtExceptionMonitor', () => {
   // Using any rather than the real type so we can test and spy on private methods
   // eslint-disable-next-line  @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


packages/js/test/unit/server/integrations/uncaught_exception_monitor.server.test.ts
```diff
@@ -42,20 +47,53 @@ describe('UncaughtExceptionMonitor', () => {
 
       // Using any rather than the real type so we can test and spy on private methods
       // eslint-disable-next-line  @typescript-eslint/no-explicit-any
-      const newMonitor = new UncaughtExceptionMonitor(client) as any
+      const newMonitor = new UncaughtExceptionMonitor() as any
       expect(removeLambdaSpy).toHaveBeenCalledTimes(1)
       expect(newMonitor.__isReporting).toBe(false)
       expect(newMonitor.__handlerAlreadyCalled).toBe(false)
+      expect(newMonitor.__listener).toStrictEqual(expect.any(Function))
+      expect(newMonitor.__listener.name).toBe('honeybadgerUncaughtExceptionListener')
```
subzero10, line 55: +1 for named functions!

subzero10 (approved):
This looks good! With this PR, we add support for loading the plugin multiple plugins, but we have yet to reload the plugin when necessary (i.e. when `configure` is called more than once), right?

PR #1172, opened 2023-08-28
chore(js): allow unhandledRejection plugin to be loaded more than once

BethanyBerkowitz:
## Status
**READY**

## Description
Part of https://github.com/honeybadger-io/honeybadger-js/issues/1163

This PR makes it safe to load the unhandledRejection plugin more than once.

If loaded with enableUnhandledRejection true, it will add a listener if one has not already been added
If loaded with enableUnhandledRejection false, it will remove the listener if it exists
However, the client itself still only loads the plugin once, so this PR should not cause any user-facing change in behavior.

## Related PRs
https://github.com/honeybadger-io/honeybadger-js/pull/1170 does the same with unhandledException

## Todos
- [x] Tests
- [x] Documentation

## Steps to Test or Reproduce
Unit tests mostly cover it, however I also tested manually using the `sails` example app. Any node-based example app would be fine, you just need to trigger a promise rejection where the middleware will not pick it up. eg

```
Honeybadger.configure({
  enableUncaught: false,
  enableUnhandledRejection: true,
})


function reject() {
  Promise.reject(new Error('error within a promise'))
}

setTimeout(reject, 500)
```
I tested with `enableUnhandledRejection: true` and `false`.

### DIFF
--- a/packages/js/src/server/integrations/unhandled_rejection_monitor.ts
+++ b/packages/js/src/server/integrations/unhandled_rejection_monitor.ts
@@ -5,10 +5,51 @@ import { Types } from '@honeybadger-io/core'
 export default class UnhandledRejectionMonitor {
   protected __isReporting: boolean
   protected __client: typeof Client
+  protected __listener: (reason: unknown, _promise: Promise<unknown>) => void
 
-  constructor(client: typeof Client) {
+  constructor() {
     this.__isReporting = false
+    this.__listener = this.makeListener()
+  }
+
+  setClient(client: typeof Client) {
     this.__client = client
+  }
+
+  makeListener() {
+    const honeybadgerUnhandledRejectionListener = (reason: unknown, _promise: Promise<unknown>) => {
+      if (!this.__client || !this.__client.config.enableUnhandledRejection) {
+        if (!this.hasOtherUnhandledRejectionListeners() && !this.__isReporting) {
+          fatallyLogAndExit(reason as Error)
+        }
+        return
+      }
+  
+      this.__isReporting = true;
+      this.__client.notify(reason as Types.Noticeable, { component: 'unhandledRejection' }, {
+        afterNotify: () => {
+          this.__isReporting = false;
+          if (!this.hasOtherUnhandledRejectionListeners()) {
+            fatallyLogAndExit(reason as Error)
+          }
+        }
+      })
+    }
+    return honeybadgerUnhandledRejectionListener
+  }
+
+  maybeAddListener() {
+    const listeners = process.listeners('unhandledRejection')
+    if (!listeners.includes(this.__listener)) {
+      process.on('unhandledRejection', this.__listener)
+    }
+  }
+
+  maybeRemoveListener() {
+    const listeners = process.listeners('unhandledRejection')
+    if (listeners.includes(this.__listener)) {
+      process.removeListener('unhandledRejection', this.__listener)
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/js/src/server/integrations/unhandled_rejection_plugin.ts
```diff
@@ -2,16 +2,17 @@ import { Types } from '@honeybadger-io/core'
...
   return {
     load: (client: typeof Client) => {
-      if (!client.config.enableUnhandledRejection) {
-        return
-      }
-      const unhandledRejectionMonitor = new UnhandledRejectionMonitor(client)
-      process.on('unhandledRejection', function honeybadgerUnhandledRejectionListener(reason, _promise) {
-        unhandledRejectionMonitor.handleUnhandledRejection(reason, _promise)
-      })
+      unhandledRejectionMonitor.setClient(client)
```
joshuap, line 10: I can't remember... is there any concern of [multiple clients overriding this](https://docs.honeybadger.io/lib/javascript/guides/working-with-multiple-clients/)? I believe we only load these plugins for the singleton, so maybe it's not a current concern? I'm just wondering if there could ever be a case where someone might want to manually add a plugin.

joshuap (approved):
This looks good to me. If it's not urgent, it would be great if @subzero10 could also review when he's back from vacation.

packages/js/src/server/integrations/unhandled_rejection_plugin.ts
joshuap: Yep, that's all right. Only the singleton client gets our plugins, since many of them are global. I think this is just something we'll need to think about if we ever decide to expose the plugin system and/or plugins to end users (Sentry does that, iirc). I think it's OK to assume there will ever only be one instance of each plugin currently.

packages/js/src/server/integrations/unhandled_rejection_monitor.ts
```diff
@@ -5,38 +5,62 @@ import { Types } from '@honeybadger-io/core'
...
+    const honeybadgerUnhandledRejectionListener = (reason: unknown, _promise: Promise<unknown>) => {
+      if (!this.__client || !this.__client.config.enableUnhandledRejection) {
+        if (!this.hasOtherUnhandledRejectionListeners() && !this.__isReporting) {
+          fatallyLogAndExit(reason as Error)
+        }
+        return
+      }
+  
+      this.__isReporting = true;
+      this.__client.notify(reason as Types.Noticeable, { component: 'unhandledRejection' }, {
```
subzero10, line 29: Unrelated to the PR itself here, but shouldn't we send more metadata to HB since we have the actual `_promise` available?

packages/js/src/server/integrations/unhandled_rejection_plugin.ts
subzero10: > I think as a user, if I used a method called factory, I'd expect to get a duplicate with the same functionality... since users didn't add these plugins in the first place, I don't think they'd expect to have to add them after using the factory.

Do we all agree that this should be the expected behavior? If so, we should make sure that plugins can be loaded in more than one client simultaneously, or even come up with singletons of them if needed.

When we have this sorted out, we can update the `factory` method to load plugins by default for all client instances.

What do you think?

subzero10 (approved):

packages/js/src/server/integrations/unhandled_rejection_monitor.ts
BethanyBerkowitz: Good question! Usually the "reason" is an error, in which case I think this should work fine. But not always! Here's an issue for us to look further into it: https://github.com/honeybadger-io/honeybadger-js/issues/1180

packages/js/src/server/integrations/unhandled_rejection_plugin.ts
BethanyBerkowitz: I created a new issue for us to be able to continue this discussion! In the meanwhile, this PR assumes the plugin is loaded in one client only, which I think is fine for now. 

https://github.com/honeybadger-io/honeybadger-js/issues/1181

subzero10: Thanks for creating the issue!

PR #1187, opened 2023-09-08
fix(nextjs): build copy files script as a commonjs module with rollup

### LINKED ISSUE #1186: `npx honeybadger-copy-config-files` is not working
Script is not built as a commonjs module.

Front logo [Front conversations](https://app.frontapp.com/open/top_scn6p)

subzero10:
## Status
**READY**

## Description
Fixes #1186.
I can't pinpoint how this broke, but I know that it was working before 😓 .
The issue is that the executable script is built with Typescript with a module target to esnext. Nodejs uses commonjs by default and we usually transform to commonjs when bundling with rollup. This PR adds a rollup config file to bundle the `honeybadger-copy-config-files` script in commonjs.

### DIFF
--- a/packages/nextjs/rollup.config.js
+++ b/packages/nextjs/rollup.config.js
@@ -37,8 +37,6 @@ export default {
     commonjs(),
     copy({
       targets: [
-        { src: 'build/copy-config-files.js*', dest: 'dist' },
-        { src: 'build/copy-config-files-exec.js*', dest: 'dist' },
         { src: 'build/*.d.ts', dest: 'dist' },
       ]
     })
--- a/packages/nextjs/rollup.scripts.config.js
+++ b/packages/nextjs/rollup.scripts.config.js
@@ -1,0 +1,28 @@
+import commonjs from '@rollup/plugin-commonjs'
+import executable from 'rollup-plugin-executable'
+import path from 'path'
+
+export default {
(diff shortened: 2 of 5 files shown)


packages/nextjs/rollup.config.js
```diff
@@ -37,8 +37,6 @@ export default {
     commonjs(),
     copy({
       targets: [
-        { src: 'build/copy-config-files.js*', dest: 'dist' },
```
subzero10, line 40: The issue is here. The Typescript built version is transpiled to esnext, not commonjs.

packages/nextjs/rollup.scripts.config.js
```diff
@@ -0,0 +1,28 @@
...
+      },
+    },
+  ],
+  external: [
+    'fs',
+    'path',
+  ],
+  plugins: [
+    commonjs(),
+    executable(),
```
subzero10, line 26: This plugin ensures that the generated file is executable.

packages/nextjs/src/copy-config-files-exec.ts
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env node
```
subzero10, line 1: Had to remove because rollup was complaining. It's added after rollup is done (see comment above).

joshuap (approved):

PR #1171, opened 2023-08-25
ci(e2e): e2e with playwright

subzero10:
## Status
**Ready**

## Description
I decided to give playwright a go and see how it performs versus our current integration tests.

- [x] Run tests on Browserstack from Playwright. Running on minimum and latest versions. Supported browsers: **chrome** and **edge**.
- [x] Run tests on local browser executables (shipped with playwright): Firefox, Chrome, Edge, Safari, Chromium and mobile browsers

### Karma vs Playwright
**Karma/Jasmine**

_Pros:_
- runs on Browserstack, supports all browsers and browser versions available on Browserstack

_Cons:_
- feels outdated
- tests are becoming more and more flaky
- lots of usage of setTimeout (especially since the switch to promises from the callback-based approach)
- difficult to debug

**Playwright**

_Pros:_
- modern api
- fast (even faster when not connected to remote browsers on browserstack)
- support for flaky tests
- local browser testing (installed through npm)
- browserstack support (though limited to chrome and edge)

_Cons:_
- recommended approach is to test only on latest browser versions, but we also want to test on older versions

## Todos
- [x] Migrate all tests to playwright
- [x] Browserstack integration
- [x] Karma/Jasmine vs Playwright comparison
- [x] README.md - explain the minimal architecture (`server.js`)

PR #1190, opened 2023-09-15
ci(e2e): fix setup

subzero10:
## Status
**READY**

## Description
A last minute change in #1171 broke the setup of the integration tests and they wouldn't exit the process when done.

- [x] If tests fail, upload playwright report as an artefact
- [x] fix `global-teardown.ts`
- [x] refactor `global-setup.ts`

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -76,10 +76,10 @@ jobs:
       uses: actions/setup-node@v3
       with:
         node-version: '14.x'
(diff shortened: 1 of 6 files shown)


subzero10:
> This lgtm but I'm not sure why you now have a test failure after the most recent commit?

I think this was a browserstack fluke (all tests for a single browser failed) - I retried and they passed.

BethanyBerkowitz (approved):

Issue #1013, opened 2023-02-02
Extract shared plugin code and share between rollup and webpack plugins


BethanyBerkowitz:
The rollup plugin was designed with code that can be shared with the webpack plugin. Extract the shared code and re-work both plugins so they share the core functionality (hbUtils.js and options.js in the rollup package). The only difference between the plugins should be how they hook into the build process and how they extract the necessary data from the bundle. 

Make sure that we do not lose any functionality from the webpack plugin in this process. Check carefully -- I thin we should be able to avoid breaking changes and a major version bump.

BethanyBerkowitz:
ignorePaths is currently rollup only, ignoreErrors currently webpack only -- need to decide what to do about this

BethanyBerkowitz:
Currently planning to move webpack plugin from adding options validation errors to the list of build errors in the compilation to just throwing an error if passed bad options. I think this this change is a positive one -- the user is only writing their code to set up the options once, so making an error with the options really obvious is positive. An error like forgetting to pass the apiKey is not going to be something intermittent, it's a total failure of setup, so I think it's correct that it would cause the build to fail completely. 

However, I'm not sure if this change could mean a major version update 😬

subzero10:
> ignorePaths is currently rollup only, ignoreErrors currently webpack only -- need to decide what to do about this

What about both?

> Currently planning to move webpack plugin from adding options validation errors to the list of build errors in the compilation to just throwing an error if passed bad options. I think this this change is a positive one -- the user is only writing their code to set up the options once, so making an error with the options really obvious is positive. An error like forgetting to pass the apiKey is not going to be something intermittent, it's a total failure of setup, so I think it's correct that it would cause the build to fail completely.
> 
> However, I'm not sure if this change could mean a major version update 😬

I'm leaning towards a minor version update, and we could play it extra safe if we implement `ignoreErrors` to simply log warning messages even if validation fails. What do you think?

BethanyBerkowitz:
> > ignorePaths is currently rollup only, ignoreErrors currently webpack only -- need to decide what to do about this
> 
> What about supporting both?
> 
https://github.com/honeybadger-io/honeybadger-js/issues/1210
https://github.com/honeybadger-io/honeybadger-js/issues/1209

[...]

PR #1193, opened 2023-09-20
feat(plugin-core): new plugin-core package

BethanyBerkowitz:
## Status
**READY**

## Description
Part of https://github.com/honeybadger-io/honeybadger-js/issues/1013. 

This PR creates a `plugin-core` package, which extracts the aspects of the `rollup-plugin` and `webpack` packages that can be shared between the two. This will mean that we have consistency between these two packages (and any further build plugins) in the following areas: 

1. Config options (defining required options and cleaning them)
2. Uploading sourcemaps to HB API
3. Uploading deploy notifications

This package includes aspects of both existing packages, mostly from https://github.com/honeybadger-io/honeybadger-js/blob/master/packages/rollup-plugin/src/hbUtils.ts and https://github.com/honeybadger-io/honeybadger-js/blob/master/packages/webpack/src/resolvePromiseWithWorkers.js. 

There will be follow-up PRs to have `webpack` and `rollup-plugin` depend on `plugin-core`.

## Todos
- [X] Tests
- [x] Documentation


packages/plugin-core/src/types.ts
subzero10, line 1: I would go for type declaration instead of interfaces:
```
export type HbPluginOptions = {
  // ....
}

File: packages/plugin-core/tsconfig.json [hunk 1]
```diff
@@ -0,0 +1,14 @@
...
+  "extends": "../../tsconfig.base.json",
+  "include": [
+    "src/**/*.ts"
+  ], 
+  "compilerOptions": {
+    "rootDir": "src", 
+    "allowSyntheticDefaultImports": true, 
+    "target": "ES2015",
+		"declaration": true,
+		"declarationDir": "dist/dts",
```
subzero10, line 11: I'm curious to know why you had to set this here 🤔

BethanyBerkowitz: I used `rollup-plugin-dts` to roll all the `.d.ts` files into one file here: https://github.com/honeybadger-io/honeybadger-js/blob/4c084a21d8b646817b39b1025af21bea9211def1/packages/plugin-core/rollup.config.mjs#L20-L21 and then used `rollup-plugin-delete` to get rid of the rest of the `.d.ts` files in the `dts` folder. This is not strictly necessary, since having multiple `.d.ts` files works fine, but makes the bundle nice and tidy with only `index.js`, `index.d.ts`, and `index.js.map`. 

As I learn more about rollup, I'm starting to appreciate making a smaller, tidier bundle... I don't know if we want to choose some kind of best practice standard for bundling but doing it this way was definitely a good learning experience.

packages/plugin-core/README.md
```diff
@@ -0,0 +1,17 @@
+# Honeybadger JavaScript Build Plugin
+
+Shared library for uploading sourcemaps and sending deploy notifications to Honeybadger. 
+
+## Usage
+
+You shouldn't need to use this package directly. This package serves as the core implementation of shared functionality meant to be used within Honeybadger's plugins for JS build/bundling systems such as [webpack](https://github.com/honeybadger-io/honeybadger-js/tree/master/packages/webpack) and [rollup](https://github.com/honeybadger-io/honeybadger-js/tree/master/packages/rollup-plugin).
```
BethanyBerkowitz, line 7: @subzero10 I modeled this README on the `core` package, which is available to the public but not encouraged for people to use. I think we actually may find circumstances later in which this may be useful for people to use directly... basically any time we recommend someone hit the sourcemap upload or deploy notification API in JS-land, they could probably use this instead. But I'd say let's use it for our own purposes first, and then decide if we want our users to use it or not later.

packages/plugin-core/tsconfig.json
subzero10: OK sounds good, I think I will give it a go at some point!

subzero10 (approved):

PR #1207, opened 2023-10-11
chore(webpack): set up basic example project to test plugin against

BethanyBerkowitz:
## Status
**READY**

## Description
In working on https://github.com/honeybadger-io/honeybadger-js/issues/1013, I found I needed a basic project to test sourcemap upload and deploy notifications. I decided to set up the project first so I could see the behavior on the existing package before making changes. 

## Steps to Test or Reproduce

```bash
> cd packages/webpack/example
> npm i
> npm run build
```
You should see 
```
Uploaded main.js.map to Honeybadger API
Successfully sent deploy notification to Honeybadger API.
```

### DIFF
--- a/packages/webpack/example/package.json
+++ b/packages/webpack/example/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "start": "open ./dist/index.html",
+    "start": "open ./index.html",
     "test": "echo \"Error: no test specified\" && exit 1",
     "build": "webpack"
   },
--- a/packages/webpack/example/package.json
+++ b/packages/webpack/example/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "example",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "start": "open ./dist/index.html",
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "webpack"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "@honeybadger-io/webpack": "file:..",
+    "webpack": "^5.88.2",
+    "webpack-cli": "^5.1.4"
+  },
+  "dependencies": {
+    "@honeybadger-io/js": "^6.5.2"
+  }
+}
--- a/packages/webpack/example/src/index.js
+++ b/packages/webpack/example/src/index.js
@@ -1,0 +1,16 @@
+const Honeybadger = require('@honeybadger-io/js');
+
+Honeybadger.configure({
+  apiKey: (prompt('Enter the API key for your Honeybadger project:')),
+})
+
+function notifyButton() {
+  const button = document.createElement('button');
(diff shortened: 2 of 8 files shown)


packages/webpack/example/webpack.config.js
```diff
@@ -0,0 +1,18 @@
+const path = require('path');
+const HoneybadgerSourceMapPlugin = require('@honeybadger-io/webpack')
+
+module.exports = {
+  entry: './src/index.js',
+  mode: 'production',
+  output: {
+    filename: 'main.js',
```
subzero10, line 8: 🤔 I'd assume this would be named `index.js`, but no big deal.

packages/webpack/example/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "example",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "start": "open ./dist/index.html",
```
subzero10, line 7: Is this file generated? Or should this line be removed from here?

packages/webpack/example/src/index.js
```diff
@@ -0,0 +1,16 @@
+const Honeybadger = require('@honeybadger-io/js');
+
+Honeybadger.configure({
+  apiKey: (prompt('Enter the API key for your Honeybadger project:')),
```
subzero10, line 4: Since `HONEYBADGER_API_KEY` is required to be passed as an env variable for this example app, you could read it from there using the [EnvironmentPlugin](https://webpack.js.org/plugins/environment-plugin/).

subzero10 (approved):
Nothing blocking, but I left some comments.

packages/webpack/example/webpack.config.js
BethanyBerkowitz: I followed the webpack intro project instructions 🤷

packages/webpack/example/package.json
BethanyBerkowitz: The webpack example project puts an `index.html` directly in the `dist` folder, but I'll go ahead and move it up to the root since we don't want to check in `dist`.

packages/webpack/example/src/index.js
BethanyBerkowitz: I do think that would work for our testing purposes, but I don't think it's the example we necessarily want to set for our users. From those docs it looks like this will put the API key into the bundled code as a string, which would expose it publicly. 

I didn't intend this little example to be about the main HB functionality... really just about the sourcemap and deploy upload, but I still wouldn't want to publish something that if copy/pasted could cause user API keys to be public.

packages/webpack/example/webpack.config.js
BethanyBerkowitz: But I agree with you 😆

packages/webpack/example/src/index.js
subzero10: 👍 Gotcha.

Btw, API keys on the client side are exposed publicly anyway in some way or another. Best we can do is hide them under minified/uglified code, but they are exposed nevertheless.

subzero10 (approved):

PR #1211, opened 2023-10-13
feat(webpack): rely on plugin-core

BethanyBerkowitz:
## Status
**READY**

## Description
Closes https://github.com/honeybadger-io/honeybadger-js/issues/1013

## Related PRs
https://github.com/honeybadger-io/honeybadger-js/pull/1193
https://github.com/honeybadger-io/honeybadger-js/pull/1195

## Todos
- [x] Tests
- [x] Documentation
- [x] Manual testing of multiple sourcemap file upload
- [x] Modify the example project to have sourcemaps actually get applied to an error -- this would make us more confident that the upload worked correctly
- [x] Manual testing of behavior when sourcemap upload fails 

## Steps to Test or Reproduce
In `/webpack`
```
npm run build
```

In `/webpack/example`
```
npm run build
```
You should see happy logging. 
Screen Shot 2023-10-13 at 4 08 10 PM

You should see the deployment and sourcemap file in the Honeybadger UI.


BethanyBerkowitz:
Sourcemaps applied!
Screen Shot 2023-10-16 at 12 56 53 PM

packages/webpack/test/HoneybadgerSourceMapPlugin.test.js
```diff
@@ -1,533 +1,257 @@
...
       expect(compilation.warnings.length).to.eq(0)
     })
 
     it('should add upload errors to compilation errors', async function () {
-      const compilation = {
-        errors: [],
-        warnings: []
-      }
-      this.plugin.ignoreErrors = false
+      plugin.options.ignoreErrors = false
```
subzero10, line 181: I think it would be good to assert that there was no attempt to send a deploy notification here (or it could be a test on its own if you prefer - "does not send deploy notification when compilation fails".

subzero10 (dismissed):
This is looking great!
It's impressive to see how much the code is reduced in this package now and it's only specific to the webpack plugin.

subzero10 (approved):

PR #1248, opened 2023-11-16
fix(plugin-core): support minified url creation with pathname in assetUrl

maxkostow:
## Status

**READY**

## Description
`honeybadger-io/next` suggest to set the assetUrl to `https://my-app.vercel.app/_next` ([link](https://docs.honeybadger.io/lib/javascript/integration/nextjs/#configuration))

However, the plugin-core uses the `URL` class in a way that does not support this. Instead of overwriting the assetsUrl pathname, this PR joins them.

## Steps to Test or Reproduce
1. Follow the instructions at https://docs.honeybadger.io/lib/javascript/integration/nextjs
2. Observe that sourcemaps are not correctly uploaded including `_next` in the pathname

### DIFF
--- a/packages/plugin-core/src/sourcemaps.ts
+++ b/packages/plugin-core/src/sourcemaps.ts
@@ -3,6 +3,7 @@ import originalFetch from 'node-fetch'
 import FormData from 'form-data'
 import { promises as fs } from 'fs'
 import fetchRetry from 'fetch-retry'
+import path from 'path'
 // @ts-expect-error
 const fetch = fetchRetry(originalFetch)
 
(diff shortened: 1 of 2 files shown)


packages/plugin-core/src/sourcemaps.ts
```diff
@@ -81,7 +82,9 @@ export async function buildBodyForSourcemapUpload(
 ): Promise<FormData> {
   const form = new FormData()
 
-  const minifiedUrl = new URL(sourcemapData.jsFilename, hbOptions.assetsUrl).href
+  const url = new URL(hbOptions.assetsUrl)
```
subzero10, line 85: This is strange, the second parameter of URL() constructor is the baseUrl, so it shouldn't be replacing it.

I will test this on a nextjs project as well.

subzero10: I confirm that the second parameter of the URL constructor will strip out any pathnames in the string.

subzero10 (approved):
Thanks!

subzero10:
This should be fixed with [@honeybadger-io/nextjs@5.8.4](https://github.com/honeybadger-io/honeybadger-js/releases/tag/%40honeybadger-io%2Fnextjs%405.8.4)!

PR #1319, opened 2024-03-24
feat: events api improvements

### LINKED ISSUE #1258: [js-server] Log events to Honeybadger insights with built-in plugin
We should build have a plugin that when enabled automatically sends console logs to Honeybadger insights.
Note: It should serve as an alternative to setting up a log drain from the infrastructure side.

## TODO
- Server events plugin
- Tests
- Documentation

subzero10:
## Status
**READY**

## Description
Closes #1258.
- Adds `event` method, which replaces logEvent to match Client Library Spec. The already existing `logEvent` method is marked as deprecated.
- `Honeybadger.event` method will send `event_type` and `ts` by default.
- Console instrumentation for both client and server (nodejs).

## Todos
- [x] Tests
- [x] Console instrumentation for server side (nodejs)
- [x] Create `event` method
- [x] Mark `logEvent` method as deprecated
- [x] Send `event_type` and `ts` by default
- [x] [Documentation](https://github.com/honeybadger-io/docs/issues/438)

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -7,6 +7,7 @@ import {
   runBeforeNotifyHandlers,
   shallowClone,
   logger,
+  logDeprecatedMethod,
   generateStackTrace,
   filter,
   filterUrl,
@@ -300,6 +301,14 @@ export abstract class Client {
     })
 
     return this
+  }
+
+  /**
+   * @deprecated Use {@link event} instead.
+   */
+  logEvent(data: Record<string, unknown>): void {
+    logDeprecatedMethod(this.logger, 'Honeybadger.logEvent', 'Honeybadger.event')
+    this.event('log', data)
   }
 
   event(type: string, data: Record<string, unknown>): void {
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -302,8 +302,12 @@ export abstract class Client {
     return this
   }
 
-  logEvent(data: Record<string, unknown>): void {
-    this.__eventsLogger.logEvent(data)
+  event(type: string, data: Record<string, unknown>): void {
+    this.__eventsLogger.log({
+      event_type: type,
+      ts: new Date().toISOString(),
+      ...data
+    })
   }
 
   __getBreadcrumbs() {
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -311,7 +311,13 @@ export abstract class Client {
     this.event('log', data)
   }
 
-  event(type: string, data: Record<string, unknown>): void {
+  event(data: Record<string, unknown>): void;
+  event(type: string, data: Record<string, unknown>): void;
+  event(type: string | Record<string, unknown>, data?: Record<string, unknown>): void {
+    if (typeof type === 'object') {
+      data = type
+      type = type['event_type'] as string ?? undefined
+    }
     this.__eventsLogger.log({
       event_type: type,
       ts: new Date().toISOString(),
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1,4 +1,10 @@
+import events from './plugins/events'
+
 export { Client } from './client'
 export * from './store'
 export * as Types from './types'
 export * as Util from './util'
+
+export const Plugins = {
+  events
+}
--- a/packages/core/src/plugins/events.ts
+++ b/packages/core/src/plugins/events.ts
@@ -1,7 +1,7 @@
 /* eslint-disable prefer-rest-params */
 import { Client } from '../client'
 import { globalThisOrWindow, instrumentConsole } from '../util';
-import { EventPayload, Plugin } from '../types'
+import { Plugin } from '../types'
 
 export default function (_window = globalThisOrWindow()): Plugin {
   return {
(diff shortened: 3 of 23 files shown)


stympy:
We have shared the `logEvent` method with a few customers, so assumed it has been publicly documented. 🙂

subzero10:
> We have shared the `logEvent` method with a few customers, so assumed it has been publicly documented. 🙂

Oh OK, I'll bring the method back then and mark it as deprecated.

packages/core/src/client.ts
```diff
@@ -302,8 +303,20 @@ export abstract class Client {
     return this
   }
 
+  /**
+   * @deprecated Use {@link event} instead.
+   */
   logEvent(data: Record<string, unknown>): void {
-    this.__eventsLogger.logEvent(data)
+    logDeprecatedMethod(this.logger, 'Honeybadger.logEvent', 'Honeybadger.event')
```
subzero10, line 310: `logEvent` will be shown as deprecated on IDEs and will also write a warn message at every 100 method calls.

packages/core/src/index.ts
```diff
@@ -1,4 +1,10 @@
+import events from './plugins/events'
+
 export { Client } from './client'
 export * from './store'
 export * as Types from './types'
 export * as Util from './util'
+
+export const Plugins = {
+  events
```
subzero10, line 9: The `events` plugin is the first plugin to be shared between server and client.

packages/core/src/plugins/events.ts
```diff
@@ -22,10 +20,10 @@ export default function (_window = globalThisOrWindow()): Types.Plugin {
...
         }
 
-        // todo: send browser info
-        client.logEvent({
-          level,
-          args
+        client.event('log', {
+          severity: level,
+          message: args[0],
+          args: args.slice(1)
```
subzero10, line 26: I could go with `args: args` here, but I thought that usually console logs have a string message as the first argument along with metadata/context as the following arguments (optionally).
@joshuap @stympy Let me know if you have any other thoughts here.

packages/core/src/util.ts
```diff
@@ -564,3 +564,39 @@ function getSourceCodeSnippet(fileData: string, lineNumber: number, sourceRadius
 export function isBrowserConfig(config: BrowserConfig | Config): config is BrowserConfig {
   return (config as BrowserConfig).async !== undefined
 }
+
+/** globalThis has fairly good support. But just in case, lets check its defined.
+ * @see {https://caniuse.com/?search=globalThis}
+ */
+export function globalThisOrWindow () {
```
subzero10, line 571: This method is not new; it was moved from `browser/util.ts`.

packages/js/examples/express/index.js
```diff
@@ -50,6 +52,16 @@ app.get('/checkin/:id', (req, res) => {
     })
 })
 
+app.get('/event', (req, res) => {
```
subzero10, line 55: Screenshot 2024-03-25 at 3 31 28 PM

packages/core/src/plugins/events.ts
stympy: How about going with this if args[0] is a string but go with args: args if args[0] is an object? I agree that most of the time `log` is called with a string as arg 0, but there are cases when `log` is called with just an object.

subzero10: Good point, done!

stympy (dismissed):

packages/core/src/client.ts
```diff
@@ -302,8 +303,20 @@ export abstract class Client {
...
+    this.event('log', data)
+  }
+
+  event(type: string, data: Record<string, unknown>): void {
```
joshuap, line 314: Will this handle both method signatures from [the spec](https://www.notion.so/honeybadger/Client-Library-Spec-aa891332f7874196aa0695b6d38dca66?pvs=4#08adddbe8f7940ad902e1aed13019df1)?

```ruby
# signature:
Honeybadger.event(event_type, payload={})
Honeybadger.event(payload={})
```

joshuap (approved):
Looks good here!

packages/core/src/client.ts
subzero10: No, but we could do it. Typescript supports function overloading:
```
  event(data: Record<string, unknown>): void;
  event(type: string, data: Record<string, unknown>): void;
  event(type: string | Record<string, unknown>, data?: Record<string, unknown>): void {
    if (typeof type === 'object') {
      data = type
      type = type['event_type'] as string ?? undefined
    }
    this.__eventsLogger.log({
      event_type: type,
      ts: new Date().toISOString(),
      ...data
    })
  }
```
What would be the `event_type` if the function is called with an object only? Should we look for an `event_type` in the object? What if the object does not have an `event_type?` The [ruby gem](https://github.com/honeybadger-io/honeybadger-ruby/blob/580bc6055d146c818bf74e5f1e3a3cee991c4e00/lib/honeybadger/agent.rb#L371-L384) does not check if `event_type` exists.

stympy: An event_type field is not required, so if there isn't one specified either as the string first argument or in the object, don't worry about it. :)

subzero10: OK, that makes sense. I'll go ahead and commit the snippet above then.

PR #1363, opened 2024-06-16
fix(js): bin files as javascript

subzero10:
## Status
**READY**

## Description
I noticed the CI [failing](https://github.com/honeybadger-io/honeybadger-js/actions/runs/8491109872/job/23262781296?pr=1325) recently.

## Problem
Lerna is not working well with our "bin" scripts and this is a known [issue](https://github.com/lerna/lerna/issues/1444). 
Locally, Lerna will install packages, symlink the ones that are inside the monorepo and finally run build scripts (such as Typescript compilation and Rollup bundling).
While symlinking (which happens before build), it will also symlink bin executables. Until now, those bin executables were not available at this phase, as they would have been generated later by the Typescript compilation.

## Solution
A workaround is to make sure that these "bin" scripts are always available and not rely on the "build" script. To do this, we point to raw JavaScript files which require the yet-to-be-built JavaScript files.

Note: I will have to test this manually as soon as we publish a release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "test": "lerna run test --stream",
     "install": "lerna bootstrap",
     "lint": "npx eslint .",
-    "release": "lerna publish --conventional-commits --changelog-preset conventional-changelog-angular --yes --create-release github --loglevel silly"
+    "release": "lerna publish --conventional-commits --changelog-preset conventional-changelog-angular --yes --create-release github --loglevel silly",
+    "clean": "./scripts/clean-repo.sh"
   }
 }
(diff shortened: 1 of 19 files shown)


stympy (dismissed):

joshuap (approved):

subzero10:
I've tested this manually and it seems to work:
```bash
$ npx honeybadger-copy-config-files
Done copying config files.
```

```
$ npx honeybadger-checkins-sync

# the error is expected here
Error: apiKey is required
    at CheckInsManager.<anonymous> (/Users/pangratioscosma/Work/honeybadger/example-nextjs-dashboard/node_modules/@honeybadger-io/js/dist/server/check-ins-sync.js:686:35)
    at step (/Users/pangratioscosma/Work/honeybadger/example-nextjs-dashboard/node_modules/@honeybadger-
....
....
```

Issue #1383, opened 2024-08-09
Includes remotely-hosted code, so cannot be used within a Chrome extension


raindrift:
## What are the steps to reproduce this issue?
1. Create a chrome extension using manifest v3
2. Set up honeybadger in accordance with https://docs.honeybadger.io/lib/javascript/integration/chrome-extension/
3. Submit extension to the Chrome web store

## What happens?
Extension is rejected after review. I received the following notice by email:

> Violation reference ID: [Blue Argon](https://developer.chrome.com/docs/webstore/troubleshooting/#additional-requirements-for-manifest-v3)
> 
> Technical Requirements - Additional Requirements for Manifest V3:
> 
> Violation:
> Including remotely hosted code in a Manifest V3 item.
> Violating Content:
> Code snippet: `service_worker.js: t.getUserFeedbackScriptUrl = function(e) { var t = e.split(".").slice(0, 2).join("."); return "https://js.honeybadger.io/v".concat(t, "/honeybadger-feedback-form.js")`
> Code snippet: `extension.js: }; t.getUserFeedbackScriptUrl = function(e) { var t = e.split(".").slice(0, 2).join("."); return "https://js.honeybadger.io/v".concat(t, "/honeybadger-feedback-form.js") };`

## What were you expecting to happen?
Extension would be accepted, on account of not breaking the rules.

## Any logs, error output, etc?
No

## Any other comments?
It took a few extension versions for them to catch this issue. Because of this, I know that Honeybadger works great in our extension! I am hoping we don't have to stop using it. I know this library has worked fine with mv3 for years. The policy isn't new, but perhaps there have been recent updates to Google's code analysis tools?

I am installing the npm version and packaging with webpack, but the same issue exists in the minified js.

It seems like this would be solvable by:

1. Include the feedback form js with the rest of the honeybadger javascript. It's not very much code, so it seems like lazy-loading it (if that's why it is separate) isn't gaining much.
2. This problem will occur again in the form itself, since it submits data by injecting a <script> tag into the page. Instead, send the form data to the backend with `fetch()`

For the moment, I will try to work around this by forking the Honeybadger library and removing the feedback form entirely, since we aren't currently using it.

## What versions are you using?
**Operating System:** osx
**Package Name:** honeybadger-js
**Package Version:** 6.9.3
**Browser Version:** Chrome 127.0.6533.89 / arm64

raindrift:
[...]

PR #1386, opened 2024-08-19
feat: chrome extension build

subzero10:
## Status
**READY**

## Description
Creates a new chrome extension build that excludes the "Collect User Feedback Form" feature that would cause chrome extensions to be rejected.
The excluded code is isolated in a file `feedback-form.ts` and the removal is done using `rollup-plugin-strip-code`.
The additional build file is `honeybadger.ext.min.js` and is also uploaded on CDN.

Closes: #1383

## Todos
- [x] Move code to be excluded for extension build in a separate file
- [x] Add new extension build: `honeybadger.ext.min.js`
- [x] Update the example app to use the new build 
- [x] Add the new extension build on CDN
- [x] [Documentation](https://github.com/honeybadger-io/docs/pull/517) - point to new extension build and add a callout for the removed feature
- [x] Manual Test
- [x] Fix integration tests


packages/js/test/unit/browser.test.ts
subzero10, line 1: No code change here, just moved to its own file.

packages/js/src/browser.ts
```diff
@@ -99,62 +96,8 @@ class Honeybadger extends Client {
...
-        return true
-      }
-    }
-
-    return false
-  }
-
-  private getUserFeedbackSubmitUrl() {
-    return getUserFeedbackScriptUrl(this.getVersion())
+    const form = new BrowserFeedbackForm(this.config, this.logger, this.getVersion());
```
subzero10, line 99: I moved feedback form code to its own file to reduce chances of breaking the code removal by rearranging methods.

subzero10: This refactor broke the integration tests. I am working on fixing them.

stympy (approved):

PR #1407, opened 2024-10-12
fix(chrome-extension): reorder code to maintain comments so that code strip will work

subzero10:
## Status
**READY**

## Description
Fixes: #1383.
Typescript would remove the comment `ROLLUP_STRIP_CODE_CHROME_EXTENSION_END` when transpiling from .ts to .js because it was sitting alone and not _connected_ to any function. I reordered the code and it seems to work now.

### DIFF
--- a/packages/js/package-lock.json
+++ b/packages/js/package-lock.json
@@ -16,7 +16,7 @@
         "honeybadger-checkins-sync": "scripts/check-ins-sync-bin.js"
       },
       "devDependencies": {
-        "@playwright/test": "^1.48.0",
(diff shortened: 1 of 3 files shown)


subzero10:
Integration tests are failing while installing Playwright browsers. I suspect this is a transient failure and it should be OK a bit later.

stympy (approved):

subzero10:
After more investigation, it seems that we have to update Playwright because the version we were using is not available anymore. I will merge this PR and work on integration tests separately.

Issue #1417, opened 2024-11-16
Print console message on source maps upload to remind users that revisions need to match


subzero10:
We receive support tickets that source maps aren't working. Many times this is because the source maps revision and the Honeybadger configuration revision do not match. We can print a message after uploading source maps to remind users that they need to make sure the revision will match the new uploaded source maps.

The message: 

> Note: For the error to be matched with a source map, the revisions must match. You can learn how to configure honeybadger.js to include the revision here: https://docs.honeybadger.io/lib/javascript/guides/using-source-maps/#versioning-your-project

PR #1418, opened 2024-11-16
chore(plugin-core): print source map revisions message

subzero10:
## Status
**READY**

## Description

Prints a message on successful source maps upload to remind users that Honeybadger needs to be configured with the same revision in order for source maps to match.

Closes: #1417

### DIFF
--- a/packages/plugin-core/src/sourcemaps.ts
+++ b/packages/plugin-core/src/sourcemaps.ts
@@ -7,6 +7,8 @@ import path from 'path'
 // @ts-expect-error
 const fetch = fetchRetry(originalFetch)
 
+const DOCS_VERSIONING_URL = 'https://docs.honeybadger.io/lib/javascript/guides/using-source-maps/#versioning-your-project';
+
 import type { HbPluginOptions, SourcemapInfo } from './types'
 
 /**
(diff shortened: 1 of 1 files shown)


packages/plugin-core/src/sourcemaps.ts
```diff
@@ -33,19 +35,21 @@ export async function uploadSourcemaps(sourcemapData: SourcemapInfo[], hbOptions
     const errorsStr = rejected.map(p => p.reason).join('\n')
     throw new Error(`Failed to upload ${rejected.length} sourcemap file(s) to Honeybadger\n${errorsStr}`)
   }
-  
+
+  console.info(`Note: For the error to be matched with a source map, the revisions must match. You can learn how to configure honeybadger.js to include the revision here: ${DOCS_VERSIONING_URL}`)
```
joshuap, line 39: It looks like we're checking an `hbOptions.silent` config option for the success message logging. I think we should also respect that here?

joshuap (changes requested):

packages/plugin-core/src/sourcemaps.ts
subzero10: I considered that too, but the `silent` option is like a debug/verbose flag. My understanding was that we would print this message at all times, not only in debug mode. I don't have a strong opinion though, we can check for silent here as well. Let me know if you want me to go for it.

subzero10: Oh, you are right, I don't know how I confused that bit 😓 .
In that case, I'll implement your suggestion and check for the flag.

joshuap (approved):
Looks good, thanks!!

Issue #1421, opened 2024-11-30
Update CI to test against newer versions of Node.js


subzero10:
CI is running against Nodejs 14, 16 and 18. Node 18 and 20 are currently in maintenance and 22 is LTS. We should at least update the version matrix to test against 18, 20 and 22. It'd be great if we added 23 as well, which is the current version.

PR #1422, opened 2024-11-30
ci: remove node 14 from matrix

subzero10:
## Status
**READY**

## Description
Removes v14 - first step of upgrading version matrix in CI (see #1421).
Part of #1421.


stympy (approved):
Should we add versions 20 and 22?

subzero10:
> Should we add versions 20 and 22?

Yes, I'll do that separately. We have #1421 to track this work.

PR #1428, opened 2025-02-25
fix(js): improved, unambiguous error message

subzero10:
## Status
**READY**

## Description
Before exiting the process due to uncaught process exceptions or unhandled promise rejections we were logging the same error message, which recently cost hours of debugging for a customer. This should make it easier to identify whether the error originated from (uncaught exception or unhandled rejection).

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -99,7 +99,7 @@ jobs:
         npm run test:integration
 
     - name: Upload Playwright Report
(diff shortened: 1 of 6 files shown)


subzero10:
Looks like integration tests are failing but the issue is unrelated to this PR. We can go ahead and merge and I will fix those in a separate PR.

stympy (approved):

PR #1453, opened 2025-09-01
fix(core): log correct endpoint used

### LINKED ISSUE #1451: Wrong domain used in console log
A customer using the EU stack noticed that the following console message was pointing to the wrong domain:

```
Error report sent ⚡ https://app.honeybadger.io/notice/${uuid}
```

subzero10:
## Status
**READY**

## Description
Note: The log message used to point to APP Honeybadger, but since `endpoint` is configurable, we shouldn't hardcode the log to the APP domain. Therefore, I changed to simply log the url that was used to submit the notice.

Fixes #1451

### DIFF
--- a/packages/js/src/server.ts
+++ b/packages/js/src/server.ts
@@ -10,7 +10,6 @@ import { lambdaHandler } from './server/aws_lambda'
 import { ServerTransport } from './server/transport'
 import { StackedStore } from './server/stacked_store'
 import { CheckInsConfig } from './server/check-ins-manager/types'
-import { CheckInsClient } from './server/check-ins-manager/client';
 
 const { endpoint } = Util
 const DEFAULT_PLUGINS = [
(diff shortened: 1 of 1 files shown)


packages/core/src/client.ts
```diff
@@ -487,7 +489,7 @@ export abstract class Client {
         runAfterNotifyHandlers(merge(notice, {
           id: uuid
         }), this.__afterNotifyHandlers)
-        this.logger.info(`Error report sent ⚡ https://app.honeybadger.io/notice/${uuid}`)
+        this.logger.info(`Error report sent to ⚡ ${endpointUrl}`)
```
stympy, line 492: I don't think this right... we don't want to be displaying the same URL that is used to submit notices to the API. Instead we want to display the corresponding _app_ URL with the `notice/${uuid}` path so that the user can click that link and be redirected to the error detail page in the UI.

packages/js/src/server/check-ins-manager/client.ts
```diff
@@ -22,7 +21,7 @@ export class CheckInsClient {
     const response = await this.transport.send({
       method: 'GET',
       headers: this.getHeaders(),
-      endpoint: `${this.BASE_URL}/v2/project_keys/${projectApiKey}`,
+      endpoint: Util.endpoint(this.config.appEndpoint, `v2/project_keys/${projectApiKey}`),
```
subzero10, line 24: `Util.endpoint` is a helper method that provides safer URL generation.
If `this.config.appEndpoint` is _https://app.honeybadger.io/_ or  _https://app.honeybadger.io_, the generated URL will be the same.

Issue #1208, opened 2023-10-13
Checkins V2


subzero10:
Implement checkin synchronization through code, as recently done for [PHP/Laravel](https://github.com/honeybadger-io/honeybadger-php/issues/172).

To look at the final result, the best place is to go to the [documentation](https://github.com/honeybadger-io/docs/pull/351).

subzero10:
Hey @shanecav84, `CheckInsClient` is not exported, but I just submitted a #1456 to make it public so that you can instantiate it in your code.
Since it's not officially documented (at least for now), I added some example code in the PR on how to use it.

PR #1456, opened 2025-09-16
feat(js): make check-ins client public

subzero10:
## Status
**WIP**

## Description
Exporting check-ins client so that it can be utilized programmatically.

https://github.com/honeybadger-io/honeybadger-js/issues/1208#issuecomment-3286523730

### DIFF
--- a/packages/js/scripts/copy-typedefs.js
+++ b/packages/js/scripts/copy-typedefs.js
@@ -13,7 +13,7 @@ function copyFileSync( source, target ) {
     }
   }
 
-  fs.writeFileSync(targetFile, fs.readFileSync(source, 'utf8').replace(/'\.\//g, '\'./types/'));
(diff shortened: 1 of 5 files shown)


packages/js/scripts/copy-typedefs.js
```diff
@@ -13,7 +13,7 @@ function copyFileSync( source, target ) {
     }
   }
 
-  fs.writeFileSync(targetFile, fs.readFileSync(source, 'utf8').replace(/'\.\//g, '\'./types/'));
+  fs.writeFileSync(targetFile, fs.readFileSync(source, 'utf8'));
```
subzero10, line 16: I don't know why this was here. I had to remove it in order to have correct type definitions.

subzero10:
Update: I've tested this manually and I'm facing some issues with rollup's tree shaking: it's not really exporting the CheckInsClient class because it's not used.

PR #1457, opened 2025-09-17
fix(js): honeybadger.checkInsApi

subzero10:
## Status
**READY**

## Description
I'm taking a different approach by adding `CheckInsClient` as a public property in the Honeybadger singleton. The previous approach, which would simply export the class, didn't work because of our current setup (default export, isomorphic package, etc.).

## Usage

```typescript
// note: this needs to run in a server side environment (nodejs)
import Honeybadger from '@honeybadger-io/js'

Honeybadger.configure({
  debug: false,
  reportData: false,
  endpoint: 'https://api.honeybadger.io',
  projectRoot: 'webpack:///./',
  apiKey: 'project api key',
  appEndpoint: 'https://app.honeybadger.io', // required for check-ins client
  personalAuthToken: 'personal auth token', // required for check-ins client
  environment: 'production',
  hostname: 'badger01',
  revision: 'git SHA/project version',
  component: 'example_comonent',
  action: 'example_action',
  breadcrumbsEnabled: true
})

// this will return an array of `CheckIn`, it will throw if the response from the api is not HTTP 200.
await Honeybadger.checkInsClient.listForProject('project id')

// this will return a single `CheckIn`, it will throw if the response from the api is not HTTP 200.
await Honeybadger.checkInsClient.get('project id', 'check-in id')

// this will return the `CheckIn` that was just created, with the value of `id` also set.
// it will throw if the response from the api is not HTTP 201.
await Honeybadger.checkInsClient.create('project id', new CheckIn({
  name: 'a simple check-in',
  slug: 'simple-check-in',
  scheduleType: 'simple',
  reportPeriod: '1 day',
  gracePeriod: '5 minutes',
}))

// this will return the `CheckIn` that was just created, with the value of `id` also set.
// it will throw if the response from the api is not HTTP 201.
await Honeybadger.checkInsClient.create('project id', new CheckIn({
  name: 'a cron check-in',
  slug: 'cron-check-in',
  scheduleType: 'cron',
  cronSchedule: '* * * * 5',
  cronTimezone: 'UTC',
  gracePeriod: '5 minutes',
}))

// update an existing check-in, identified by its slug.
// it will throw if the response from the api is not HTTP 204.
await Honeybadger.checkInsClient.update('project id', new CheckIn({
  name: 'a simple check-in',
  slug: 'simple-check-in',
  scheduleType: 'simple',
  reportPeriod: '1 day',
  gracePeriod: '15 minutes',
}))
    
// remove an existing check-in, identified by its id. 
// use `checkInsClient.listForProject` to get check-ins and with their ids.
// it will throw if the response from the api is not HTTP 204.
await Honeybadger.checkInsClient.remove('project id', new CheckIn({
  id: 'check-in id',
  name: 'a simple check-in',
  slug: 'simple-check-in',
  scheduleType: 'simple',
  reportPeriod: '1 day',
  gracePeriod: '15 minutes',
}))
```

Also note that you can use Honeybadger to trigger CheckIns:
```typescript
// check-in with an id
await Honeybadger.checkIn('check-in id')

// check-in with a slug
await Honeybadger.checkIn('check-in-slug')
```

This PR builds on #1456.

### DIFF
--- a/packages/gatsby-plugin/CHANGELOG.md
+++ b/packages/gatsby-plugin/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [6.0.40](https://github.com/honeybadger-io/honeybadger-js/compare/@honeybadger-io/gatsby-plugin-honeybadger@6.0.39...@honeybadger-io/gatsby-plugin-honeybadger@6.0.40) (2025-09-16)
+
+**Note:** Version bump only for package @honeybadger-io/gatsby-plugin-honeybadger
+
+
+
+
+
 ## [6.0.39](https://github.com/honeybadger-io/honeybadger-js/compare/@honeybadger-io/gatsby-plugin-honeybadger@6.0.38...@honeybadger-io/gatsby-plugin-honeybadger@6.0.39) (2025-09-16)
 
 **Note:** Version bump only for package @honeybadger-io/gatsby-plugin-honeybadger
--- a/packages/gatsby-plugin/package-lock.json
+++ b/packages/gatsby-plugin/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "@honeybadger-io/gatsby-plugin-honeybadger",
-  "version": "6.0.39",
+  "version": "6.0.40",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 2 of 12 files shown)


subzero10:
Just tested this manually ([v6.12.2](https://github.com/honeybadger-io/honeybadger-js/releases/tag/%40honeybadger-io%2Fjs%406.12.2)) and it works.

PR #1484, opened 2026-01-21
fix(js): preserve original error in browser onunhandledrejection handler

### LINKED ISSUE #1180: Can we provide more data when reporting an unhandled rejection?
See https://github.com/honeybadger-io/honeybadger-js/pull/1172#discussion_r1316741778

When reporting an unhandled rejection in server-side JS using the plugin, we only provide the "reason". 

The "reason" is usually an error, in which case `client.notify(reason)` should work great. However, the "reason" could also be of some other type. In this case, can we provide more information using the promise passed through in the [unhandledRejection event](https://nodejs.org/api/process.html#event-unhandledrejection)?

jdhuntington:
## Summary

- Preserves the original Error instance via `originalError` property when handling unhandled promise rejections in the browser
- Allows `beforeNotify` handlers to access custom properties on Error subclasses
- Adds type definition for `originalError` on Notice interface

## Problem

The browser's `onunhandledrejection` handler creates a new plain object with only `name`, `message`, and `stack` properties, discarding the original Error instance. This makes it impossible to access custom error properties in `beforeNotify` callbacks.

For example, with a custom error class:

```typescript
class UnauthenticatedError extends Error {
  readonly skipReporting = true;
}
```

The following filter never works for unhandled rejections because the original error is lost:

```typescript
honeybadger.beforeNotify((notice) => {
  if (notice.skipReporting) return false;  // Always undefined
});
```

## Solution

Add an `originalError` property to the error object passed to `notify()`:

```typescript
honeybadger.beforeNotify((notice) => {
  if (notice.originalError?.skipReporting) {
    return false;
  }
});
```

## Test plan

- [x] Added tests verifying `originalError` is preserved
- [x] Added tests verifying custom properties are accessible via `originalError`
- [x] All existing tests pass

Closes #1180

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -88,6 +88,7 @@ export interface Notice {
   __breadcrumbs: BreadcrumbRecord[],
   afterNotify?: AfterNotifyHandler,
   cause?: Error|Record<string, unknown>,
+  originalError?: Error,
   [key: string]: unknown
 }
 
--- a/packages/core/src/util.ts
+++ b/packages/core/src/util.ts
@@ -324,7 +324,7 @@ export function makeNotice(thing: Noticeable): Partial<Notice> {
     notice = {}
   } else if (isErrorObject(thing)) {
     const e = thing as Error
-    notice = merge(thing as Record<string, unknown>, { name: e.name, message: e.message, stack: e.stack, cause: e.cause })
+    notice = merge(thing as Record<string, unknown>, { name: e.name, message: e.message, stack: e.stack, cause: e.cause, originalError: e })
   } else if (typeof thing === 'object') {
     notice = shallowClone(thing)
   } else {
--- a/packages/core/test/util.test.ts
+++ b/packages/core/test/util.test.ts
@@ -3,6 +3,7 @@ import {
   merge,
   mergeNotice,
   objectIsEmpty,
+  makeNotice,
   runBeforeNotifyHandlers,
   runAfterNotifyHandlers,
   shallowClone,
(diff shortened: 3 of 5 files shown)


subzero10:
Hey @jdhuntington, thanks for taking the time to submit this PR and for digging into this issue.

This is a solid attempt, but I’m not sure it will fully work as intended. When `client.notify` is called, a new `notice` object is created internally, and that `notice` is what eventually gets passed to the `beforeNotify` handler. Because of that, we need to make sure the `originalError` property is also set at notice creation time, specifically in `makeNotice` (`packages/core/src/client.ts`). It’s been a while since I last worked in this area, but that’s my understanding of the flow.

While reviewing this, I also noticed that we only construct a new error object in the browser `onunhandledrejection` integration. On the server side, both `server/integrations/unhandled_rejection_monitor.ts` and `server/integrations/unhandled_exception_monitor.ts` simply call `client.notify(reason)` directly, passing through the original object.

Given that, I wonder if we could align the browser behavior with the server and avoid creating a new error object there as well. Something along these lines in `browser/integrations/onunhandledrejection.ts`:

```typescript
if (!reason.stack) {
  const fileName = 'unknown'
  const lineNumber = 0
  const stackFallback = `${reason.message}\n    at ? (${fileName}:${lineNumber})`
  reason.stack = stackFallback
}

reason.message = `UnhandledPromiseRejectionWarning: ${reason}`
// ...
client.notify(reason)
```

Then, inside `makeNotice`, we can set `originalError` whenever the input is an `Error` instance.

This way, `originalError` becomes available consistently whenever the original value is an `Error`, not just for the browser `onunhandledrejection` path. It also keeps the behavior between browser and server integrations aligned.

Curious what you think.

subzero10:
Hey @jdhuntington, thanks for the follow-up and for being open to changing the approach.

Yes, passing reason directly to `client.notify(reason)` is the direction we want. Your snippet is on the right track. A few small notes:
1. Your browser change: Mutating `reason` (stack fallback + message) and then calling `client.notify(reason)` is what we’re after. Using `UnhandledPromiseRejectionWarning: ${reason.message}` in the message is good so the notice still shows a clear warning while keeping the original message in the string.
2. `originalError` on the notice: So that `beforeNotify` always receives a notice with the `originalError`, we should set `originalError` when the notice is created, not at the call site. Concretely: in `core`, in `makeNotice` (in `packages/core/src/util.ts`), when the input is an `Error`, set `notice.originalError = thing` (or add it to the object you merge). That way:
    - The notice passed to `beforeNotify` always has `notice.originalError` when the input was an `Error`.
    - The same behavior applies for both browser and server (e.g. server unhandled rejection already passes reason through; with this change, those notices will also get `originalError`).

So the PR could be:
- Browser (`onunhandledrejection.ts`): your change — ensure stack/message on reason, then `client.notify(reason)` (no new object, no `originalError` in the integration).
- Core (`util.ts`): in the `isErrorObject(thing)` branch in `makeNotice`, set `originalError` on the notice to that `Error`.
- Types (`types.ts`): add `originalError?: Error` to the `Notice` interface so it’s typed and documented.

If you’re good with that split (browser: pass-through + your snippet; core: set originalError in makeNotice), go ahead and revise the PR that way. The existing tests that assert on `originalError` should be updated to assert on the notice that gets built (e.g. the `notice` passed to `beforeNotify` or the result of `makeNotice`), since `originalError` will now be set in core rather than in the browser integration.

Happy to review the next iteration.

subzero10 (commented):
This looks good!
I'd ask one more test before approving and merging.
Let's add a test in `core/test/client.test.ts` to test your use case: _custom error properties are accessible in beforeNotify_

subzero10:
Closing in favor of #1495 (this branch was from a fork so I couldn't push directly to it).

PR #1510, opened 2026-07-15
feat(js): legible ui.click breadcrumb selectors via breadcrumbsSelectorAttributes

### LINKED ISSUE #1509: Support a data attribute for legible click-breadcrumb selectors
## Problem

When the browser SDK records a `ui.click` breadcrumb, it builds a CSS selector for the clicked element and its ancestors. Each segment is `tag + #id + .allClasses + [alt/name/title/type] + :nth-child(n)`, and the walk goes all the way up to `<body>`.

In Tailwind (and similar utility-CSS) apps every element carries dozens of utility classes, so the selector becomes an unreadable wall of text. A customer reported a click breadcrumb whose selector began:

```
body > div#root > main > div > div.flex.min-h-screen.flex-col.font-sans.antialiased.sm:px-4.sm:py-12.bg-gray-100.dark:bg-black.text-foreground > div.relative.w-full.pb-4.flex.h-full.flex-auto.flex-col.justify-center.sm:mx-auto.sm:max-w-lg.sm:rounded-lg.sm:shadow-md.bg-background > ... (continues for many more segments)
```

There is currently no way to make this legible — no config option, and no attribute the SDK honors. The only attributes ever emitted are the hardcoded list `['alt', 'name', 'title', 'type']`.

## Proposal

Add a new browser config option:

```js
Honeybadger.configure({
  breadcrumbsSelectorAttributes: ['data-hb-name'] // default; can point at data-testid, etc.
})
```

When an element has one of the configured attributes with a non-empty value, that value **replaces the entire segment** (no tag/classes/nth-child), and reaching a named element **anchors** the selector — the upward walk stops there and class-soup ancestors above it are dropped.

**Example** — click `<h3 class="text-left font-semibold">` inside `<div data-hb-name="deal-card">`:

- `message`  → `h3.text-left.font-semibold`
- `selector` → `deal-card > div > h3.text-left.font-semibold`

Tagging the element itself collapses both `message` and `selector` to just the clean name.

Opt-in and backward-compatible: apps that don't use `data-hb-name` and make no config change see identical output.

## Design

Full design doc: `docs/superpowers/specs/2026-07-15-breadcrumb-selector-attributes-design.md` (branch `feat/breadcrumb-selector-attributes`).

Touches `packages/js/src/browser/util.ts` (`stringNameOfElement`, `stringSelectorOfElement`), `packages/js/src/browser/integrations/breadcrumbs.ts` (call sites), `packages/core/src/types.ts` (`BrowserConfig`), and `packages/js/src/browser.ts` (default).

Front logo [Front conversations](https://app.frontapp.com/open/top_2gfso1)

stympy:
## Summary

Resolves #1509. Lets browser SDK users opt into legible `ui.click` breadcrumb selectors by tagging elements with a configurable data attribute (default `data-hb-name`), instead of long Tailwind/utility-CSS class chains.

When a clicked element — or one of its ancestors — carries one of the configured attributes with a non-empty value, that value replaces the element's entire selector segment, and the named ancestor anchors the selector (the ancestor walk stops there).

**Before** (utility-CSS app):
```
body > div#root > div > div.flex.flex-col.gap-4 > div.rounded-lg... > h3.text-left.font-semibold
```
**After** (with `<div data-hb-name="deal-card">…</div>`):
```
deal-card > h3.text-left.font-semibold
```

## Changes

- **`@honeybadger-io/core`** — new `breadcrumbsSelectorAttributes: string[]` field on `BrowserConfig`.
- **`@honeybadger-io/js`**
  - `DEFAULT_SELECTOR_ATTRIBUTES = ['data-hb-name']` exported from `browser/util.ts`; wired as the browser client default.
  - `cleanNameOfElement` helper resolves the clean name from the first matching attribute, with normalization (`trim` → collapse internal whitespace → `truncate(…, 100)`).
  - `stringNameOfElement` returns the bare clean name when an element is tagged; `stringSelectorOfElement` anchors the chain at the nearest named ancestor.
  - Config threaded into the click listener in `browser/integrations/breadcrumbs.ts`.
- **Docs** — `breadcrumbsSelectorAttributes` documented in `packages/js/README.md`.

## Behavior notes

- **Backward compatible:** with no configured attribute present in the DOM, output is identical to today (existing selector regression tests unchanged).
- **Malformed config never throws:** non-array `attributes`, non-string/empty entries, and elements without `getAttribute` fall back to normal tag/class output.
- `<html>` is never a named anchor.
- Set `breadcrumbsSelectorAttributes: []` to disable the behavior entirely.
- The `selector` string is display-only in the Honeybadger UI (not re-queried against the DOM), so output need not be valid CSS.

## Testing

- New unit tests in `test/unit/browser/util.browser.test.ts` (clean-name resolution, normalization, truncation, precedence, `<html>` exception, malformed config, anchoring, custom attribute lists) and a new `test/unit/browser/integrations/breadcrumbs.browser.test.ts` (default attributes end-to-end + reconfigure-after-load).
- Full js unit suite passes: 102 browser + 128 server.
- Typecheck (`tsc --build`) and `eslint` clean.
- Reviewed by Codex CLI — no concrete bugs found.


docs/superpowers/plans/2026-07-15-breadcrumb-selector-attributes.md
subzero10, line 1: I still haven't decided yet whether we should be pushing plans (and specs) into the repo.
The reason is because you:
- either have to maintain them as the code evolves, or 
- you need to ask the AI Agent to ignore them entirely or 
- if it does read these files you need to choose what to do in case these `.md` files do not match the code anymore (is it a regression? a missed requirement from the original implementation? or has the code evolved?). 

If you choose to maintain them, I think it becomes an _expensive_ task as you add more `.md` files.
If you ask the agent to ignore them after implementing the feature/task, these files could still serve as a reference to the original implementation. However, I've been using other mediums to store technical documents for reference, such as Notion. They are still accessible by AI Agents, but usually I do that only on demand.

What are your thoughts on this?

packages/js/src/browser/util.ts
```diff
@@ -2,19 +2,40 @@ import { Util } from '@honeybadger-io/core';
 
 const { globalThisOrWindow } = Util;
 
+export const DEFAULT_SELECTOR_ATTRIBUTES = ['data-hb-name']
```
subzero10: I think this should be inside the `defaults.ts` file in the `core` package. That's where we store all default configuration options (one reason is because it becomes easy to keep it in sync with the [docs](https://docs.honeybadger.io/lib/javascript/reference/configuration/#configuration-options)).

Probably another Agent SKILL: all configuration options must define their default values in `core/src/defaults.ts`.

packages/js/README.md
```diff
@@ -11,6 +11,34 @@ Universal JavaScript library for integrating apps with the :zap: [Honeybadger Er
 
 For comprehensive documentation and support, [check out our documentation site](http://docs.honeybadger.io/lib/javascript/index.html).
 
+## Configuration
+
+The full list of configuration options lives on the [documentation
+site](http://docs.honeybadger.io/lib/javascript/index.html). A few browser
+options are documented here.
+
+#### `breadcrumbsSelectorAttributes`
```
subzero10: Do we want to have this here? I think we settled on keeping all documentation in docs.honeybadger.io.

subzero10 (changes requested):
Looks good to me, only a couple of minor comments.

docs/superpowers/plans/2026-07-15-breadcrumb-selector-attributes.md
stympy: I don't like leaving the superpowers docs in the repo for the reasons you outlined. They are a snapshot in time, and IMO aren't worth maintaining after the implementation is complete. Sometimes (depending on the complexity of the feature(s)) in other repos I will have Claude build an as-built doc to replace the superpowers docs, but I'm not convinced of the long-term usefulness of that, either. :)

The spec file _can_ be useful for helping explain to the reviewer the motivation for the change and/or decisions made about the implementation, so that's why I tend to include it in the PR. Perhaps istwould be useful to replace the spec with an ADR to track/explain decisions, then lean on the PR description to explain the motivation. I don't see any value in keeping/reviewing the implementation plan, but I forget to have the agent remove it before opening the PR. Maybe that should become a skill. :)

All that said, my plan is to remove them before merging the branch.

packages/js/README.md
stympy: Agreed. Another skill candidate. :)

stympy:
Docs for this option are in honeybadger-io/docs#823 — the two should ship together.

Also addressed @subzero10's review here:
- Removed the superpowers spec and plan from the repo.
- Moved the default to `core/src/defaults.ts` as `BREADCRUMBS_SELECTOR_ATTRIBUTES`. Note it's exported alongside `CONFIG` rather than added to it — `CONFIG` is spread into every client including Node, but this option is browser-only (it lives on `BrowserConfig`, not `Config`). Happy to move it inside `CONFIG` if you'd rather have it there for docs-sync purposes.
- Dropped the README section in favor of the docs site.

subzero10 (approved):

docs/superpowers/plans/2026-07-15-breadcrumb-selector-attributes.md
subzero10: Noted on the skill. I can take a stab at creating these skills.

packages/js/README.md
subzero10: https://github.com/honeybadger-io/honeybadger-js/issues/1511

docs/superpowers/plans/2026-07-15-breadcrumb-selector-attributes.md
subzero10: https://github.com/honeybadger-io/honeybadger-js/issues/1511

PR #1461, opened 2025-10-22
ci: switch to NPM trusted publishing

subzero10:
## Status
**READY**

## Description
- [x] Setup trusted publishing for all packages
- [x] Replace PATs with GitHub App
- [x] Configure GitHub App to allow pushing on master without a pull request / review
- [x] Update Readme
- [x] Bonus: Update GitHub Actions versions to latest versions

PR #1514, opened 2026-07-17
ci: upgrade to lerna 9 + npm workspaces so NPM trusted publishing works

stympy:
## Why

Every **Publish New Release** run since the switch to NPM Trusted Publishing (#1461) has failed at the npm publish step with:

```
E404 Not Found - PUT https://registry.npmjs.org/@honeybadger-io%2fcore
```

Root cause: `lerna publish` in lerna 6 goes through its bundled `libnpmpublish@7`, which predates trusted publishing. The OIDC token exchange (GitHub Actions ID token → short-lived npm token) is performed by the npm CLI ≥ 11.5.1, but lerna never shells out to the npm CLI — so the registry PUT went out **unauthenticated**, and npm reports that as a 404 on scoped packages. The workflow's "install npm@latest" step couldn't help for the same reason. Recent releases (6.9.0, 6.10.0) only made it to npm because someone published outside CI.

Lerna added native OIDC trusted publishing support in **v9** ([recipe](https://lerna.js.org/docs/recipes/oidc-trusted-publishing), [lerna#4219](https://github.com/lerna/lerna/issues/4219)); the installed v9 publish code contains the same token-exchange flow as the npm CLI. With lerna 9, the existing workflow setup (`id-token: write`, trusted publisher configured on npmjs.com per package) should work as-is.

## What

- **lerna 6.6.2 → 9.0.7**; drop the removed `useNx` option from `lerna.json`. All flags used by the `release` script (`--conventional-commits`, `--changelog-preset`, `--create-release github`) still exist in v9.
- **Migrate to npm workspaces** — `lerna bootstrap` was removed in v9. Root `package.json` gains `"workspaces": ["packages/*"]`, the root `"install": "lerna bootstrap"` script is gone, the 10 per-package lockfiles are deleted in favor of the single root lockfile, and all 12 packages link as workspace symlinks (verified: all `@honeybadger-io/*` deps resolve as `link` entries, `lerna list` sees all packages).
- **CI (`node.js.yml`)**: unit job Node 18.x → 22.x. lerna 9 requires Node `^20.19 || ^22.12 || >=24`. 22.x specifically because the js `shutdown_plugin` test crashes jest workers on Node 20 (`process.listeners` blows up when the async SIGTERM listener outlives the test — likely the old "CI failures with 20.x" from the previous comment; passes on 18/22/25). Cache paths collapsed to the root lockfile.
- **AGENTS.md** updated for the workspaces world.

## Test fixes (fallout of workspace re-resolution)

Hoisting + fresh in-range dependency versions surfaced several latent issues:

| Package | Fix |
|---|---|
| core | Cast sinon fakes passed to typed helpers — `@types/sinon` (react's devDep) is now hoisted where core's `tsc` sees it |
| react-native | `jest.restoreAllMocks()` in `afterEach` — jest ≥ 29.6 `mockReset` no longer reinstates a spy's original implementation, so a leaked `ErrorUtils.setGlobalHandler` spy became a no-op mock |
| react | Declare `react-dom@19` devDep (previously relied on auto-installed peer) and pin `react`/`react-dom` to package-local copies in jest config — root hoists react 18 for other packages, and mixing it with react 19 elements broke rendering |
| rollup-plugin | `import type` for type-only rollup imports; add `sourcemapFileName`/`preliminaryFileName` to `OutputChunk` fixtures (rollup 3.30 types); `.mocharc.cjs` disables Node ≥ 22.6 native TS type-stripping so mocha keeps using the ts-node/register CJS path these tests depend on |
| vue | Declare `babel-plugin-dynamic-import-node` — `.babelrc` uses it, but it only existed transitively via older `@babel/plugin-transform-modules-*` |
| nextjs | `jest` from PATH instead of hardcoded `./node_modules/jest/bin/jest.js` (gone under hoisting) |
| webpack | Raise mocha timeout to 10s — integration tests run real webpack builds and can exceed the 2s default under parallel `lerna run test` |

## Verification

- `npm run build` (lerna run build): ✅ 11/11 projects, Node 22
- `npm test` (lerna run test): ✅ 11/11 projects, 396 tests across jest + mocha suites, Node 22
- `npm run lint`: ✅ 0 errors (9 pre-existing warnings)
- rollup-plugin additionally verified on Node 20 / 22 / 25
- lerna 9 `publish` code inspected: contains the OIDC exchange (`/-/npm/v1/oidc/token/exchange/...`) for GitHub Actions

## Notes for the release after merge

- The real end-to-end proof is the next `Publish New Release` run on master.
- Double-check on npmjs.com that each package's trusted publisher points at `lerna-publish.yml` (the fallback workflow docs suggest it may have been switched at some point).
- This commit touches most packages, so lerna will patch-bump them on the next release — expected.

https://claude.ai/code/session_01NXVjnv3HhFRoMkn1kgRbZd


subzero10 (dismissed):

subzero10:
This is one reason why we didn't upgrade to Lerna v7 in the past. NPM workspaces are not so great, compared to PNPM or Yarn. I created issue #1518 to migrate to PNPM as a follow-up.

subzero10:
@stympy I will merge and monitor the publish action.

stympy:
Thanks! Should I put Claude on that for you? 🙂

subzero10:
> Thanks! Should I put Claude on that for you? 🙂

No need this time, I can ask Claude as soon as I merge this and confirm publishing works.

subzero10 (approved):

PR #1611, opened 2026-08-19
ci: merge the three publish workflows into one dispatchable workflow

subzero10:
## Problem

Releases run through `lerna-publish.yml` on push to `master`. When that fails partway — most often at the NPM publish step — the fallback is `npm-publish.yml` (or `cdn-publish.yml` for a CDN-only failure).

But **NPM Trusted Publishing binds to a workflow *filename*, and it's configured per package.** This repo publishes 12 public packages, so using a fallback today means:

1. Flip the Trusted Publisher on all 12 packages to `npm-publish.yml`
2. Run the workflow
3. Flip all 12 back

Any release that happens while the config is flipped fails.

## Fix

Fold all three publish paths into `lerna-publish.yml` behind a `workflow_dispatch` `mode` input. Because every path now runs from the same file, the OIDC claim NPM sees is identical, so **the fallback modes need no NPM-side changes at all.**

| Mode | What it does |
| --- | --- |
| `full` (default) | Runs CI, then `pnpm run release`. This is what a push to `master` runs — unchanged behaviour. |
| `npm` | Skips CI, runs `lerna publish from-package` against the versions already in `master`. The `@honeybadger-io/js` `postpublish` hook runs too, so the CDN is updated as well. |
| `cdn` | Skips CI, re-uploads the `packages/js` browser bundle only. |

`npm-publish.yml` and `cdn-publish.yml` are deleted.

## Notes for reviewers

- **The filename must stay `lerna-publish.yml`** — that's what the existing Trusted Publisher config on all 12 packages points at. Renaming or re-splitting this file later requires updating every package on NPM first. This is now called out in `AGENTS.md`.
- `inputs.mode || 'full'` is what preserves the push behaviour: the `inputs` context is empty on a `push` event.
- The `publish` job needs `if: ${{ !failure() && !cancelled() }}`. Without a status function in that expression, GitHub would skip `publish` whenever `ci` is skipped — i.e. on every `npm`/`cdn` run.
- The GitHub App token is now gated to `full` mode, with checkout falling back to `github.token`. The deleted fallback workflows used the default token, and keeping that means an App-token problem can't also block the recovery paths.
- The CDN path moves from Node 20 to 22, matching CI. `release-cdn.sh` only shells out to `jq`/`aws`/`curl`, so the Node version only affects the build that runs during `pnpm install`.
- The unused `workflow_call` trigger is dropped (no in-repo caller).

## Verification

- `actionlint` is clean on the merged workflow.
- After merge, both fallback modes are safe no-ops when the last release succeeded:
  - `gh workflow run lerna-publish.yml -f mode=npm` — should report no packages to publish and exit 0. **This is the key check**: it proves the Trusted Publishing handshake still works from this file with no NPM-side change.
  - `gh workflow run lerna-publish.yml -f mode=cdn` — re-uploads the current bundle; idempotent.
  - In both, confirm `ci` is *skipped* (not failed) and `publish` still runs.
- The `full` path is exercised by the next real merge to `master`.

## Also included

A second commit symlinks this repo's `.agents/skills` into `.claude/skills`. Claude Code discovers skills from `.claude/skills`, so the three skills already committed here (including `review-before-commit`) were never actually being loaded.


.github/workflows/lerna-publish.yml
```diff
@@ -1,20 +1,26 @@
...
+# publisher on every package and flipping it back afterwards.
+#
+#   full (default) - run CI, then version + publish to NPM + CDN + create the GitHub release
+#   npm            - re-publish the versions already committed to master (lerna from-package)
+#   cdn            - re-upload the packages/js browser bundle to the CDN only
 on:
   push:
     branches: [ master ]
   workflow_dispatch:
-  workflow_call:
```
subzero10, line 6: This is not used anymore.

subzero10:
Went through the review feedback. One change applied, the rest declined with reasons.

**Applied** — duplicated env vars (Copilot): the AWS/Bunny block was repeated in all three publish steps. Hoisted to job-level `env` in 462c7cd; they feed the same `postpublish` hook in every mode. `GH_TOKEN`/`GITHUB_TOKEN` stay step-level since only `full` uses them.

**Declined:**

- *README table renders as an extra empty column (Copilot)* — false positive. No row starts with `||`; the source is a standard `| Mode | What it does |` table.
- *Blank line inside blockquote / MD028 (CodeRabbit)* — removing it would merge the `[!NOTE]` and `[!WARNING]` blocks into one blockquote and break the second alert's rendering on GitHub. The blank line is required here, and it matches the existing style elsewhere in this README. markdownlint isn't in CI.
- *`concurrency: ci-release` lets a recovery run cancel an active release (Copilot)* — `cancel-in-progress` defaults to `false`, so runs queue rather than cancel. Queueing is the desired behaviour: a fallback run should wait for an in-flight release, not race it.
- *`ref: master` is non-deterministic (Copilot)* — pre-existing, and required. `lerna publish` in `full` mode pushes the version commit and tags, which needs an attached branch; a detached `github.sha` checkout would break it. The `ci-release` concurrency group is what prevents overlap.
- *Restrict the App token with `permission-*` inputs (CodeRabbit)* — legitimate hardening, but out of scope here and on a path that can't be exercised without a real release. Better as its own PR. Note this PR already narrows exposure by gating the App token to `full` mode; `npm`/`cdn` now run on `github.token`.

@subzero10 on `workflow_call` — confirmed unused, removed in this PR.

stympy (dismissed):
I agree with the suggestion to have the sensitive environment variables set closer to where they are needed to avoid unnecessary exposure, even though it means some duplication.

subzero10:
Reverted the hoist in 17a4ba6 — @coderabbitai is right and this one's on me. Job-level `env` exposes the AWS/Bunny credentials to *every* step, and `pnpm install --frozen-lockfile` runs the root `prepare` script (a full `lerna run build`), so every dependency lifecycle script in the tree would have seen `AWS_SECRET_ACCESS_KEY`. Scoping beats DRY here; the credentials are back on just the three publish steps.

Worth noting the two bots pulled in opposite directions on this: Copilot flagged the duplication, CodeRabbit flagged the exposure. The duplication is the correct trade.

---

On the other half of the merge-risk note — *"may use an unsupported npm version for Trusted Publishing"* — I dug into this and **it doesn't apply to lerna 9**, so I'm not adding an npm upgrade step:

- lerna 9.0.7 publishes through its bundled `libnpmpublish@11.1.2`, not by shelling out to the npm CLI, so the runner's global npm version isn't what performs the publish.
- It carries its own OIDC implementation at `dist/libs/core/src/lib/oidc.d.ts` ("Adapted from https://github.com/npm/cli/blob/latest/lib/utils/oidc.js"), which reads `ACTIONS_ID_TOKEN_REQUEST_URL` and does the token exchange itself. There's no npm-version gate on that path.

This is also why the `npm install -g npm@latest` step ("needed for trusted publishing", added in ef52b16) was dropped in the pnpm migration (0b2f0fc) and didn't need replacing.

One thing that did surface while checking, unrelated to this PR but relevant to why the npm step keeps failing — lerna's OIDC helper is documented as *"intended to never throw... OIDC is always an optional feature, and the function should not throw if OIDC is not configured by the registry."* So a failed OIDC exchange doesn't surface as an OIDC error; lerna falls through to whatever token is in config and the publish then fails as a generic auth/404. Worth a separate issue rather than widening this PR.

stympy (approved):

PR #1630, opened 2026-08-20
fix(ci): stop dependabot grouping major updates into one pull request

subzero10:
## Status
**READY**

## Description

The root entry I added in #1612 grouped every version update with `patterns: ["*"]` and no
`update-types` restriction. Because `directory: /` covers **all published packages** through the
single `pnpm-lock.yaml`, that produced #1624:

> chore(deps): bump the workspace group with 62 updates

**137 dependency changes across all 11 packages, 116 of them major version jumps.** It could never
merge:

- `typescript ^4.5.3 → ^7.0.2` in 8 packages — verified to break `vue-tsc`
  (`ERR_PACKAGE_PATH_NOT_EXPORTED`; TS 7 no longer exports `typescript/lib/tsc`)
- `next ^13.2.3 → ^16.3.1` — verified incompatible with the webpack config
  `@honeybadger-io/nextjs` injects, since Next 16 defaults to Turbopack
- `fastify ~4 → ~5` plus `fastify-plugin ^4 → ^6`, `jest 26/27/29 → 30`, `eslint 8 → 10`,
  `rollup 2/3 → 4`, `sinon 14 → 22`, `chai 4 → 6`, `@babel/* 7 → 8`

Most seriously, it bumped **runtime `dependencies`** of `@honeybadger-io/plugin-core` and
`@honeybadger-io/rollup-plugin`: `node-fetch ^2.6.9 → ^3.3.2`, `fetch-retry ^5 → ^6`,
`picomatch ^2 → ^4`. **node-fetch 3 is ESM-only**, so that would break CommonJS consumers of
published packages — and merging would have cut a release of them.

## The fix

Restrict every version-update group to `update-types: ["minor", "patch"]`.

**Majors are not skipped.** They fall outside the group and arrive as individual pull requests, so
each can be reviewed, merged, or closed on its own. This is
[the documented pattern](https://docs.github.com/en/code-security/tutorials/secure-your-dependencies/optimizing-pr-creation-version-updates#example-3-individual-pull-requests-for-major-updates-and-grouped-for-minorpatch-updates).

| Entry | Before | After |
| --- | --- | --- |
| root `/` | grouped **everything** | `dependency-type: development` + `update-types: [minor, patch]`, plus an explicit `open-pull-requests-limit: 5` |
| 8 example entries | grouped everything | version-update groups restricted to `[minor, patch]` |
| example security groups | broad | unchanged — collapsing per-app security noise is why they exist |

The root entry groups **development** dependencies only, because anything in `dependencies` there
ships to users. Its security group is dropped so root security updates arrive individually — each
of the four we triaged (undici, esbuild, fastify, next) needed its own decision, and grouping them
would have produced one unmergeable PR.

## Same flaw, three other PRs

This is exactly why these were unmergeable — one broken major poisoned an otherwise fine batch:

- **#1618** nextjs group — carried Next 16
- **#1619** react-native group — carried expo 57 / react-native 0.87
- **#1627** vue group — carried TypeScript 7

## Also

Corrected a header comment that hedged on whether a grouped PR spans multiple directories. It
does — observed as titles like *"bump the nextjs-examples group across 2 directories with 14
updates"*.

## Verification

- `.github/dependabot.yml` parses as YAML: `version: 2`, 9 entries, every version-update group
  carrying `update-types: [minor, patch]`.
- Confirmed against GitHub's docs that restricting a group to minor/patch opens majors as
  individual PRs rather than dropping them, and that removing the root security group does not
  suppress root security updates — they simply arrive ungrouped.
- No code change, so nothing to build or test; `npx lerna changed` reports no changed packages.

After merge, confirm on the next Dependabot run that no `bump the workspace group with N updates`
PR reappears, and that majors show up as separate PRs.

## Related

**#1624 should be closed, not merged.**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,10 +13,17 @@ version: 2
 #   - The root entry groups development dependencies only. Anything in `dependencies` there
 #     ships to users, so those updates each get their own pull request.
 #   - `directories` (plural) supports globbing; `directory` (singular) does not.
-#   - A single grouped pull request spans every directory the glob matches: observed as
-#     titles like "bump the nextjs-examples group across 2 directories with 14 updates".
-#   - `open-pull-requests-limit` does not apply to security updates, so grouping is the
-#     only effective cap on security-update noise.
+#   - Observed, not guaranteed by this config: grouped runs on the example entries have
+#     produced one pull request spanning several matched directories, e.g. "bump the
+#     nextjs-examples group across 2 directories with 14 updates". The documented way to
+#     consolidate a single dependency across directories is `group-by: dependency-name`,
+#     which no group here sets, so don't depend on that shape holding.
+#   - `open-pull-requests-limit` applies only to version updates, never to security updates.
+#     Grouping reduces security-update pull request volume but is not a hard cap; Dependabot
+#     applies its own internal limit. The example entries group theirs to keep that noise
+#     down. The root entry deliberately does not: those updates can reach code that ships to
+#     users, and each of the four we triaged needed a separate decision, so they are left
+#     individual even though that means more pull requests.
 #   - There is no interaction between this file and Dependabot security alerts: it cannot
 #     stop security PRs, only shape how they are batched.
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,14 +4,17 @@ version: 2
 # dozens of single-dependency pull requests.
 #
 # Notes for future maintainers:
-#   - Each entry needs TWO groups. A group defaults to `applies-to: version-updates`;
-#     security updates are only grouped when a group explicitly declares
-#     `applies-to: security-updates`.
+#   - A group defaults to `applies-to: version-updates`; security updates are only grouped
+#     when a group explicitly declares `applies-to: security-updates`.
+#   - Version-update groups are restricted to `update-types: [minor, patch]` on purpose.
+#     Grouping majors produces pull requests that cannot be merged as a unit: one broken
+#     major blocks every other update in the group. Majors arrive individually so each can
+#     be reviewed, merged, or closed on its own.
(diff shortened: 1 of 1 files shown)


subzero10:
Both comments addressed in `docs(ci): clarify the dependabot grouping notes` — comments only, no
configuration values changed.

**@copilot — correct, the note contradicted the config.** It claimed grouping is "the only
effective cap on security-update noise" while this PR removes the root entry's security group. I
took option (b), documenting the intent, since leaving root security updates ungrouped is
deliberate: each of the four we triaged (undici, esbuild, fastify 4→5, next 13→15) needed its own
decision, and grouping them would have produced one unmergeable PR. The note now explains that
`open-pull-requests-limit` covers version updates only, that the example entries group their
security updates to keep that noise down, and why the root entry does not.

**@coderabbitai — right, and I was overstating it.** I'd written that a grouped PR "spans every
directory the glob matches" as though this config guaranteed it. It doesn't: cross-directory
consolidation of a single dependency is documented to require `group-by: dependency-name`, which no
group here sets. The observation itself is real — #1618 came through as "across 2 directories with
14 updates" and #1617 as "across 3 directories" — but that's an observed result, not a promise of
this configuration, and the note now says so explicitly and points at `group-by: dependency-name`
as the documented mechanism.

While rewriting, two further inaccuracies of mine got fixed:

- "grouping is the only way to cap security-update volume" — grouping *reduces* volume, it isn't a
  cap; Dependabot applies its own internal limit.
- "every root security update touches published code" — too broad. All four root security PRs were
  in fact `deps-dev`. Reworded to "can reach code that ships to users", with individual review
  framed as the deliberate policy rather than a claim about every update.

Verified after the change: the file still parses as `version: 2` with 9 entries, the root entry
still has exactly one group (`workspace-dev-minor-patch`, limit 5), all 8 example entries still
carry a security group, and no version-update group is missing `update-types`.

